PR #48, opened 2019-02-06
Emphasize Offline/Intermittent Connection Support

### LINKED ISSUE #28: [Docs] Main page: emphasize support for intermittent connections
vs. "when a stable internet connection isn’t guaranteed" .. specifically, Ground works offline, but  requires occasional access to the internet. This allows seamless sync of data w/o any special user knowledge or actions.

Something like "Ground is an open source, map-first data collection platform that enables users to collect, map, and share data. Ground is built from the ground up (pun intended) to provide seamless sync of data over internet connections that are only occasionally available".

scolsen:
I tweaked the site copy to emphasize mobile support. I also moved the **Offline First** feature bullet to a more prominent position.

As a bonus, I formatted the file contents to fit within 80 columns. 👍 

Fixes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,3 +21,90 @@ This repo contains all Ground cloud-based / hosted components:
 Firestore Cloud Functions currently only supports Node.js v6.11.5, 
 while the web dashboard is built using a newer version of Node.js. Contributors should use the appropriate version of Node for the component they are working on. Follow the instructions provided in the README of each subdirectory to set up the proper
 development environment for each component. 
+
+## Contributing
+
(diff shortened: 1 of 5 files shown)


scolsen:
Note this doesn't emphasize mobile 📵support--it emphasizes offline/occasional connection support. 😅

scolsen:
I had to perform a rebase on this as I mucked up the PR with some commits that had to be discarded (related to asset paths). I've managed to undo all that and get us back to a state in which this PR only contains the original change--just some updates to the homepage copy!

@gino-m

gino-m:
LGTM. Thanks for the changes!

Issue #19, opened 2018-11-19
[Form editor] Big scary warnings when deleting forms, fields, options


gino-m:
"WARNING: Deleting a  ** will cause any associated data to be lost. Are you sure?"

In the future we should mark these as "deleted" and purge them only when there is no data associated.

gyarill:
Can you assign to me?

PR #50, opened 2019-03-07
Add warning when user wants to delete form or element of form. Issue #19.

gyarill:
Clicking confirm proceeds with the action. canceling closes the dialog


gino-m:
First PR for Ground Web! W00t! 🎉

PR #54, opened 2019-04-12
Routes module

gyarill:
Create file with all the route paths and make sure that match selector extracts all available information from patch 

#9

### DIFF
--- a/web/src/components/gnd-main/index.js
+++ b/web/src/components/gnd-main/index.js
@@ -25,12 +25,13 @@ import { connectProject, connectFeature } from "../../datastore.js";
 import { featurePath } from '../../route.js'
 
 
-function GndObservations({ match }) {
+function GndFeatureDetails({ match }) {
   return (
     <div>
-      TODO(maralka): exatact observations for feature
-      { match.params.projectId } in project { match.params.featureId } and
-      display in side panel.
+      { /* TODO(maralka): Display feature data loaded in the store inside of
+        side panel. */ }
+      Placeholder for { match.params.featureId } of project
+      { match.params.projectId }
     </div>
   );
 }
@@ -40,10 +41,12 @@ class GndMain extends React.Component {
     return (
       <React.Fragment>
         <GndMap />
-        <GndHeader />
-        <GndLegend />
-        <GndFeatureTypeEditor />
-        <Route path={featurePath} component={connectFeature(GndObservations)} />
+        <GndHeader projectId={this.props.match.params.projectId} />
+        <GndLegend projectId={this.props.match.params.projectId} />
+        <GndFeatureTypeEditor projectId={this.props.match.params.projectId} />
+        <Route
+          path={featurePath}
+          component={connectFeature(GndFeatureDetails)} />
       </React.Fragment>
     );
   }
--- a/web/src/components/gnd-main/index.js
+++ b/web/src/components/gnd-main/index.js
@@ -16,10 +16,24 @@
  */
 
 import React from "react";
+import { Route } from 'react-router' // react-router v4
 import GndMap from "../gnd-map";
 import GndHeader from "../gnd-header";
 import GndLegend from "../gnd-legend";
 import GndFeatureTypeEditor from "../gnd-feature-type-editor";
+import { connectProject, connectFeature } from "../../datastore.js";
+import { featurePath } from '../../route.js'
+
+
+function GndObservations({ match }) {
+  return (
+    <div>
+      TODO(maralka): exatact observations for feature
+      { match.params.projectId } in project { match.params.featureId } and
+      display in side panel.
+    </div>
+  );
+}
 
 class GndMain extends React.Component {
   render() {
@@ -29,9 +43,10 @@ class GndMain extends React.Component {
         <GndHeader />
         <GndLegend />
         <GndFeatureTypeEditor />
+        <Route path={featurePath} component={connectFeature(GndObservations)} />
       </React.Fragment>
     );
   }
 }
 
-export default GndMain;
+export default connectProject(GndMain);
--- a/web/src/datastore.js
+++ b/web/src/datastore.js
@@ -24,13 +24,10 @@ import firebaseConfig from "./.firebase-config.js";
 // empty, or undefined if key is not present (i.e., still loading,).
 const getFirestoreData = (store, key) =>
 	key in store.firestore.data ? store.firestore.data[key] || {} : undefined;
-const getActiveProjectId = store => store.path.projectId;
 
 // TOOD: Handle loading state.
 const getActiveProject = store =>
-	getActiveProjectId(store) === ":new"
-		? {}
-		: getFirestoreData(store, "activeProject");
+	getFirestoreData(store, "activeProject") || {};
 
 const getMapFeatures = store => getFirestoreData(store, "mapFeatures");
 
@@ -99,9 +96,9 @@ const connectFeature =
 	firestoreConnect((store, props) => [
 		{
 			collection: `projects/${store.match.params.projectId}/records`,
-      where: [
-         ['featureId', '==', store.match.params.featureId]
-      ],
+			where: [
+				['featureId', '==', store.match.params.featureId]
+			],
 			storeAs: "featureRecords"
 		}
 	]);
@@ -114,7 +111,6 @@ const exportKml = props => (projectId, featureTypeId) => {
 
 export {
 	getActiveProject,
-	getActiveProjectId,
 	exportKml,
 	updateProject,
 	updateProjectTitle,
@@ -122,7 +118,7 @@ export {
 	getProfile,
 	getMapFeatures,
 	getLocalizedText,
-  connectFeature,
+	connectFeature,
 	connectProject,
 	generateId
 };
--- a/web/src/datastore.js
+++ b/web/src/datastore.js
@@ -15,8 +15,6 @@
  * limitations under the License.
  */
 
-import { compose } from "redux";
-import { connect } from "react-redux";
 import { firestoreConnect } from "react-redux-firebase";
 import firebaseConfig from "./.firebase-config.js";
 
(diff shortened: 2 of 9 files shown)


web/src/route.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+import { matchPath } from "react-router";
+import { getLocation } from "connected-react-router";
+
+export const featureDisplayPath = "/p/:projectId/f/:featureId"
+export const projectPath = "/p/:projectId"
+
+// Match selector will choose first match so put them in order.
+export const paths = [
```
gino-m: Don't think we want to export the whole list, only the base paths needed by <Route> right?

web/src/route.js
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { matchPath } from "react-router";
+import { getLocation } from "connected-react-router";
+
+export const featureDisplayPath = "/p/:projectId/f/:featureId"
```
gino-m: Could you kindly reverse the order of these two and then:
```javascript
export const featureDisplayPath = `${projectBasePath}/f/:featureId`;
```

web/src/route.js
```diff
@@ -0,0 +1,53 @@
...
+export const projectPath = "/p/:projectId"
```
gino-m: Consider renaming `projectBasePath` since it will be used to form the basis of multiple other paths.

web/src/route.js
```diff
@@ -0,0 +1,53 @@
...
+
+  return state => {
+    const { pathname } = getLocation(state) || {}
+    if (pathname === lastPathname) {
+      return lastMatch
+    }
+    lastPathname = pathname
+    for (var i in paths) {
+      const match = matchPath(pathname, paths[i])
+      if (!match || !lastMatch || match.url !== lastMatch.url) {
```
gino-m: Why do we bail out if the first path doesn't match or if there was no previous match?

web/src/route.js
```diff
@@ -0,0 +1,53 @@
...
+  projectPath,
+];
+
+// It only makes sense to recalculate the `matchPath` whenever the pathname
+// of the location changes. That's why `createMatchSelector` memoizes
+// the latest result based on the location's pathname.
+// Note: function based on createMatchSelector from "connected-react-router".
+export const createMatchSelector = () => {
+  let lastPathname = null
+  let lastMatch = null
```
gino-m: These are local, so how do `lastMatch` and `lastPathname` get preserved across invocations?

web/src/route.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
```
gino-m: Nit: s/2018/2019/

web/src/route.js
```diff
@@ -0,0 +1,53 @@
...
+  featureDisplayPath,
+  projectPath,
+];
+
+// It only makes sense to recalculate the `matchPath` whenever the pathname
+// of the location changes. That's why `createMatchSelector` memoizes
+// the latest result based on the location's pathname.
+// Note: function based on createMatchSelector from "connected-react-router".
+export const createMatchSelector = () => {
+  let lastPathname = null
```
gino-m: Although ";" is optional in ES6, I'd recommend using them throughout. It's not always obvious when they're needed, so better to be consistent and to avoid potential surpises.

web/src/route.js
gino-m: Ohhh, I see. This is a function that creates a function. 

To avoid surprises, can we add these to module scope and just export matchSelector() instead of createMatchSelector()?

web/src/route.js
gino-m: Also, do we need to iterate over both of these paths for us to match, or would the /p/:projectId/f/:featureId also match /p/:projectId?

gino-m: We spoke - the fn returned by createMatchSelector doesn't match featureId if it's not present.

web/src/route.js
gyarill: so we decided to jut create array of selectors and reuse the library code in the same way it was.

gino-m:
Gentle ping 🐱

gino-m:
Hi @gyarill! Are you actively working on this? If not should I prepare a separate PR for it?

gyarill:
So I think I updated it in the way our colleagues advised us during the meeting. The data is pulled on creation of the elements. wdyt?

web/src/components/gnd-main/index.js
```diff
@@ -16,10 +16,24 @@
...
 import GndLegend from "../gnd-legend";
 import GndFeatureTypeEditor from "../gnd-feature-type-editor";
+import { connectProject, connectFeature } from "../../datastore.js";
+import { featurePath } from '../../route.js'
+
+
+function GndObservations({ match }) {
+  return (
+    <div>
+      TODO(maralka): exatact observations for feature
```
gino-m: Nit: Please wrap in comment tags `<!--` `-->`

web/src/components/gnd-main/index.js
```diff
@@ -16,10 +16,24 @@
...
+import { Route } from 'react-router' // react-router v4
 import GndMap from "../gnd-map";
 import GndHeader from "../gnd-header";
 import GndLegend from "../gnd-legend";
 import GndFeatureTypeEditor from "../gnd-feature-type-editor";
+import { connectProject, connectFeature } from "../../datastore.js";
+import { featurePath } from '../../route.js'
+
+
+function GndObservations({ match }) {
```
gino-m: Naming nit: Maybe this should be `GndFeatureDetails`? More realistically this will live inside another component that represents the side panel, but out of scope of this PR.

web/src/reducers.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
```
gino-m: I recently learned that we don't retroactively update the copyright year on existing files, so this should stay 2018.

web/src/route.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
```
gino-m: Nit: 2019

web/src/datastore.js
```diff
@@ -82,22 +80,31 @@ const generateId = props => () => props.firestore.collection("ids").doc().id;
...
+const connectFeature =
+	firestoreConnect((store, props) => [
+		{
+			collection: `projects/${store.match.params.projectId}/records`,
+      where: [
+         ['featureId', '==', store.match.params.featureId]
+      ],
+			storeAs: "featureRecords"
+		}
+	]);
```
gino-m: Nit: Please use two spaces for indentation:
https://google.github.io/styleguide/jsguide.html#formatting-block-indentation

Consider using Beautify for VSCode or equivalent in Atom, Sublime, or your favorite editor. Eventually we'll set up formatting requirements w/lint checks.

web/src/datastore.js
```diff
@@ -115,6 +122,7 @@ export {
 	getProfile,
 	getMapFeatures,
 	getLocalizedText,
-	connectGndDatastore,
+  connectFeature,
```
gino-m: Nit: Indentation.

web/src/components/gnd-main/index.js
gino-m: Nit: Extract

web/src/components/gnd-main/index.js
gyarill: Yes, I used it as a placeholder and just to show that it is possible to get projectId and featureId from the match

web/src/datastore.js
gyarill: so i used prettier and it resulted in more diffs then I expected

gyarill: in my atom it looks like tabs are being used... I just copy pasted what we in the lines that you added... 

Maybe I will try VSCode for next change..

web/src/reducers.js
gino-m: This needs to stay as 2018 - here it looks like it's being updated 2019.

web/src/route.js
gino-m: New files should have Copyright 2019.

web/src/datastore.js
gino-m: Looking at the diff in GitHub the Indentation is still off. Please check your editor settings and if necessary format manually. Thanks!

web/src/datastore.js
gino-m: It still looks pretty funky. Could you update your settings and try again, manually if necessary?

web/src/components/gnd-main/index.js
gyarill: i had to do {/**/}

web/src/datastore.js
```diff
@@ -99,9 +96,9 @@ const connectFeature =
 	firestoreConnect((store, props) => [
 		{
 			collection: `projects/${store.match.params.projectId}/records`,
-      where: [
-         ['featureId', '==', store.match.params.featureId]
-      ],
+			where: [
+				['featureId', '==', store.match.params.featureId]
+			],
```
gino-m: These are tabs. Please use spaces instead.

web/src/datastore.js
gyarill: is it still funky? I did it by hand..

gino-m:
Thanks! Looks about right, just one more nit and can merge! W00t!

web/src/datastore.js
gyarill: Gino, if I open any of your files and replace indent with two spaces than I see a diff. I am really confused with what you are asking me for here because I literally copy pasted what is in lines that you consider proper...

gino-m: If you highlight them in the diff above you'll see they're actually tabs; I believe your editor may be replacing the two spaces with tabs. Is that possible?

gino-m: Ah, I see what you mean, I may have misunderstood, since when I highlight other text in GitHub it also gives the same effect. Please disregard, I'll figure out what's goin gon in a follow-up!

gyarill:
yeeeey victory

PR #82, opened 2019-06-20
Refactor routing and add selected feature path

gyarill:
Create file with all the route paths. Add subroute for feature and connect loading of feature data to a placeholder component that sits under it. This fixes lint problems with files affected.

### DIFF
--- a/web/src/components/gnd-main/index.js
+++ b/web/src/components/gnd-main/index.js
@@ -16,22 +16,47 @@
  */
 
 import React from 'react';
+import {Route} from 'react-router'; // react-router v4
 import GndMap from '../gnd-map';
 import GndHeader from '../gnd-header';
 import GndLegend from '../gnd-legend';
 import GndFeatureTypeEditor from '../gnd-feature-type-editor';
+import {connectProject, connectFeature} from '../../datastore.js';
+import {featurePath} from '../../route.js';
 
-class GndMain extends React.Component {
+function GndFeatureDetails({match}) {
+  return (
+    <div>
+      {/* TODO(maralka): Display feature data loaded in the store inside of
+        side panel. */}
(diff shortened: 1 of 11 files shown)


web/src/components/gnd-main/index.js
```diff
@@ -16,22 +16,47 @@
...
   render() {
     return (
       <React.Fragment>
         <GndMap />
-        <GndHeader />
-        <GndLegend />
-        <GndFeatureTypeEditor />
+        <GndHeader projectId={this.props.match.params.projectId} />
+        <GndLegend projectId={this.props.match.params.projectId} />
+        <GndFeatureTypeEditor projectId={this.props.match.params.projectId} />
```
gino-m: Can we pass projectId in as a prop rather than digging into the match params directly?

In either case we should `const projectId = this.props.match.params.projectId`.

web/src/components/gnd-project-editor/index.js
```diff
@@ -54,10 +53,10 @@ const styles = (theme: Object) => ({
 // TODO: Define Project object type.
 type Props = {
   classes: Object,
+  close: Function,
```
gino-m: Is `type` a feature of Flow or React? How does it relate to `PropTypes`?

gyarill: To be honest I've never seen this type of declaring prop types before. 
I used to just use "Component.propTypes ="

but now when you say it googling leads to flow pages indeed https://flow.org/en/docs/react/components/

Do you want me to change it?

gino-m: It sounds like the linter then was checking for Flow types, not React PropTypes? I think using Flow for props type checking is better because it will now also now be able to do static type checking in code that references these props, so this looks right. Thanks!

PR #86, opened 2019-07-08
Issue 79: Renamed labels to Layer

deepakbansal:
Renamed labels "Legend" and "feature type" to "Layer" and removed one "Feature type" heading entirely.

### DIFF
--- a/web/src/components/gnd-legend/index.js
+++ b/web/src/components/gnd-legend/index.js
@@ -174,7 +174,7 @@ class GndLegend extends React.Component<Props> {
       <Card className={classes.card}>
         <CardContent className={classes.cardContent}>
           <Typography className={classes.heading} color="textSecondary">
-            Legend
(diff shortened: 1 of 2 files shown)


web/src/components/gnd-legend/index.js
```diff
@@ -174,7 +174,7 @@ class GndLegend extends React.Component<Props> {
       <Card className={classes.card}>
         <CardContent className={classes.cardContent}>
           <Typography className={classes.heading} color="textSecondary">
-            Legend
+            Layer
```
gino-m: Please use "Layers" instead (plural), since the panel currently shows all layers available, not just one.

deepakbansal: Thanks! Updated the code.

gino-m:
Resolves #121.

Issue #13, opened 2018-11-18
[Form editor] Drag to reorder form fields


gino-m:
Mock:
![image](https://user-images.githubusercontent.com/228050/48669624-a4961400-ead6-11e8-84d6-8879e6bea233.png)

Redlines:
![image](https://user-images.githubusercontent.com/228050/48669627-a9f35e80-ead6-11e8-9ab8-a93350c8a363.png)

gino-m:
Note: Only drag handle (top part of form field card) should be draggable.

gino-m:
@DaoyuT Hi! Any updates here? ⭐️

DaoyuT:
https://github.com/google/ground-platform/pull/90 created

PR #90, opened 2019-07-15
#13: [Form editor] Drag to reorder form fields	

DaoyuT:
Need help:
- [x] ~~drag reorder works and saves the data, but row shows blank when dragging, how to debug?~~ css below
- [x] `npm install -s` https -> http, keep?

TODO:
- [x] make handle prettier 
- [x] css: change row to be on top of table

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -15163,6 +15163,16 @@
         }
       }
     },
+    "react-sortable-hoc": {
+      "version": "1.9.1",
+      "resolved": "https://registry.npmjs.org/react-sortable-hoc/-/react-sortable-hoc-1.9.1.tgz",
+      "integrity": "sha512-2VeofjRav8+eZeE5Nm/+b8mrA94rQ+gBsqhXi8pRBSjOWNqslU3ZEm+0XhSlfoXJY2lkgHipfYAUuJbDtCixRg==",
+      "requires": {
+        "@babel/runtime": "^7.2.0",
+        "invariant": "^2.2.4",
+        "prop-types": "^15.5.7"
+      }
+    },
     "react-swipeable-views": {
       "version": "0.12.18",
       "resolved": "https://registry.npmjs.org/react-swipeable-views/-/react-swipeable-views-0.12.18.tgz",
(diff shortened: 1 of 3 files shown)


web/package-lock.json
```diff
@@ -1035,7 +1035,7 @@
         },
         "whatwg-fetch": {
           "version": "2.0.4",
-          "resolved": "https://registry.npmjs.org/whatwg-fetch/-/whatwg-fetch-2.0.4.tgz",
+          "resolved": "http://registry.npmjs.org/whatwg-fetch/-/whatwg-fetch-2.0.4.tgz",
```
gino-m: Does https://github.com/npm/npm/issues/20106#issuecomment-383770315 help?

Looks like we may need to upgrade to a newer version of nvm to permanently fix the problem. Please lmk if it works?

web/src/components/gnd-feature-type-editor/gnd-form-editor.js
```diff
@@ -48,6 +53,23 @@ const styles = (theme) => ({
   },
 });
 
+const DragHandle = sortableHandle(() => <span>::</span>);
```
gino-m: For this I think we can use `drag_indicator` from https://www.npmjs.com/package/@material-ui/icons

web/src/components/gnd-feature-type-editor/gnd-form-editor.js
```diff
@@ -48,6 +53,23 @@ const styles = (theme) => ({
...
+    <DragHandle />
+    <GndFormElementEditor
+      key={element.id}
+      element={element}
+      onChange={onChange}
+    />
+  </div>
+));
+
+const SortableContainer = sortableContainer(({children}) => {
```
gino-m: According to [the docs](https://github.com/clauderic/react-sortable-hoc#sortablecontainer-hoc), we can also set `lockAxis` to `y` to ensure dragging happens only vertically.

web/src/components/gnd-feature-type-editor/gnd-form-editor.js
```diff
@@ -48,6 +53,23 @@ const styles = (theme) => ({
   },
 });
 
+const DragHandle = sortableHandle(() => <span>::</span>);
+
+const SortableItem = sortableElement(({element, onChange}) => (
```
gino-m: You can try setting the z-index on a helper as in https://github.com/clauderic/react-sortable-hoc/issues/87#issuecomment-405911683 to make sure the item stays above the modal while dragging.

web/package-lock.json
DaoyuT: It worked, thanks!

Issue #92, opened 2019-07-20
Sheets integration


gino-m:
This was working but probably doesn't work anymore.

- [ ] Explore possibility of Ground creating and managing spreadsheet rather than user.

gino-m:
Assigning to @sergeydolgov1 for prioritization (MVP or RC, relative priority).

gino-m:
Moving to RC backlog for now.

gino-m:
+1 from Jacob @ SERVIR

Issue #68, opened 2019-06-14
[Import/Export] Implement CSV Export


gino-m:

gino-m:
This will expose a URL, e.g.:
https://us-central1-gnddemo1.cloudfunctions.net/exportCsv?project=frogs&featureType=FaKQLlkBOA29I5tmNRkF

The output will look like
```
Field name 1,field name 2
Record 1 field 1,Record 1 field 2
Record 2 field 1,Record 2 field 2
```

PR #93, opened 2019-07-20
Upgrade and refactor Cloud Functions

gino-m:
In preparation for #68 and #92. This should also fix the 3 vulnerability warnings reported by GitHub.

@DaoyuT do you mind taking a look? You can click on each commit below this description to see changes incrementally rather than in aggregate. Tyyy!


functions/auth.js
```diff
@@ -23,7 +23,7 @@ const {JWT} = require('google-auth-library');
 
 const API_SCOPES = ['https://www.googleapis.com/auth/spreadsheets'];
 
-class GndAuth {
+class Auth {
```
DaoyuT: If we are removing 'gnd' prefix here, are we removing 'gnd' prefix for other variables like GndFocusableRow & GndFormEditor ect. ?

gino-m:
Those are actually UI components not variables; I think the convention for
application specific components is to prefix them, so we can leave those
for now. We can revisit this once we get to productizing things. Thanks!

functions/auth.js
gino-m: See comment below - basically this is a normal class, while the Gnd* classes in web are React components, which might be treated as a special case. I think it's probably unnecessary for those as well, I think it might be a pattern I carried over from the Polymer 2 impl.

PR #97, opened 2019-07-31
Fix bug#94 and change reorder behavior to match UI

DaoyuT:
For bug#94, somehow using component class instead of functional component prevents other elements in SortableContainer from re-rendering. Might be a weird behavior in react-sortable-hoc.

Sneaking another bug fix (2nd commit) found during fixing bug#94:  change reorder logic from 'swap idxA idxB' to 'remove idxA then insert it back at idxB' so that it matches what renders in UI.

### DIFF
--- a/web/src/components/gnd-feature-type-editor/gnd-sortable.js
+++ b/web/src/components/gnd-feature-type-editor/gnd-sortable.js
@@ -1,0 +1,71 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


web/src/components/gnd-feature-type-editor/gnd-sortable.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
```
gino-m: Copyright 2019

Thanks for refactoring this

gino-m:
Please fix the copyright before merging.

web/src/components/gnd-feature-type-editor/gnd-sortable.js
DaoyuT: Done, need LGTM again =)

PR #101, opened 2019-08-22
Fix Cloud Functions deps

gino-m:
- [x] Resolves Firebase deprecation warnings (old versions of admin and shell)
- [x] Resolves missing Sheets API dep
- [x] Resolves all known vulnerabilities
- [x] Removes unused `require`


gino-m:
Resolved conflicts and resubmitted as #102

PR #103, opened 2019-08-23
Open drawer on marker click

### LINKED ISSUE #69: [Map] Update URL path on feature click
Split off from and blocking #9

gyarill:
- [x] Update URL on marker click
- [x] Prevent drawer from crashing app when not logged in

Resolves #69.

### DIFF
--- a/web/src/components/gnd-feature-details/index.js
+++ b/web/src/components/gnd-feature-details/index.js
@@ -25,6 +25,7 @@ import CardHeader from '@material-ui/core/CardHeader';
 import Typography from '@material-ui/core/Typography';
 import {getFeatureRecords} from '../../datastore.js';
 import {withStyles} from '@material-ui/core/styles';
+import {connectFeature} from '../../datastore.js';
 
 const styles = (theme) => ({
   paper: {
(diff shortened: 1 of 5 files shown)


web/src/components/gnd-feature-details/index.js
```diff
@@ -81,9 +82,11 @@ class GndFeatureDetails extends React.Component<Props> {
         </Typography>;
       });
       return <Card key={index} classes={{root: classes.card}}>
-        <CardHeader title={record.created.user.displayName} classes={{title: classes.title, root: classes.cardHeaderRoot}}/>
+        <CardHeader
+          title={record.created && record.created.user ? record.created.user.displayName : ""}
+          classes={{title: classes.title, root: classes.cardHeaderRoot}}/>
         <CardContent classes={{root: classes.cardContentRoot}}>
-          <p className={classes.date}>{record.created.clientTimestamp}</p>
+          <p className={classes.date}>{record.created ? record.created.clientTimestamp : ""}</p>
```
gino-m: Please assign this and display name to local var for readability. Could also use short circuit operators instead of ternary e.g.:
const created = record.created;
const creatorName = created && created.user && created.user.displayName;

web/src/components/gnd-feature-details/index.js
```diff
@@ -92,7 +95,7 @@ class GndFeatureDetails extends React.Component<Props> {
       <Drawer
         anchor="right"
         open={true}
-        pullRight={true}
+        pullright={"true"}
```
gino-m: Why does this need to be a string? Please add to comment.

Issue #128, opened 2019-09-10
[Layers dialog] Make "Layers" heading larger/more visible


gino-m:
Hi @DaoyuT Mind taking this small polish/nit FR?

gino-m:
Note that this refers to the title "Layers" in the floating panel on the right, not the customize layer dialog.

gino-m:
Obsolete; will be reimplemented in new version.

Issue #127, opened 2019-09-10
[Default style] Add padding around default style selector


gino-m:
Some additional polish for the demo

PR #129, opened 2019-09-10
Add padding around default style selector and increase size of Layer header

DaoyuT:
Issue #127 &
Issue #128


gino-m:
I made a tweak that I had to force push, but it seems GitHub thinks I wanted to close the issue. I'll create a new PR with all the commits for restamp.

PR #130, opened 2019-09-11
[Edit layer dialog] Additional spacing tweaks and commits from #129

gino-m:
image

### DIFF
--- a/web/src/components/gnd-feature-type-editor/index.js
+++ b/web/src/components/gnd-feature-type-editor/index.js
@@ -77,14 +77,12 @@ const styles = (theme) => ({
     zIndex: 3,
   },
   defaultStyle: {
-    paddingTop: 8,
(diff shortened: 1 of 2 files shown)


web/src/components/gnd-feature-type-editor/index.js
```diff
@@ -73,14 +73,16 @@ const styles = (theme) => ({
   },
   colorPickerTab: {
     position: 'absolute',
-    left: 145,
+    left: 165,
```
DaoyuT: Move this leftward since you have moved the default style, I tried and 142 looks like a good fit for you new change.

gino-m: Sgtm.

gino-m:
@DaoyuT please approve if you think it LG.

Issue #118, opened 2019-09-10
[Feature panel] Can not scroll down/up


DaoyuT:
When I open feature panel, I cannot scroll down/up.

PR #131, opened 2019-09-12
remove extra padding from cards

### LINKED ISSUE #126: [Feature panel] Reduce padding around record/observation cards

dhvogel:
closes #126

### DIFF
--- a/web/src/components/gnd-feature-details/index.js
+++ b/web/src/components/gnd-feature-details/index.js
@@ -26,12 +26,15 @@ import Typography from '@material-ui/core/Typography';
 import {getActiveProject, getFeatureRecords} from '../../datastore.js';
 import {withStyles} from '@material-ui/core/styles';
 import {connectFeature} from '../../datastore.js';
+import GridList from '@material-ui/core/GridList';
 
 const styles = (theme) => ({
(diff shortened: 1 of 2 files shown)


dhvogel:
also closes #118

gino-m:
@dhvogel Seems there are some conflicts now.. would you mind merging with `master` and pushing again?

dhvogel:
@gino-m There is one change here that I would like your opinion on: I changed the default font to "Google Sans". It looks like the font used in the mocks (and the one that google uses). Let me know if you think this is a good change or not. If not we can take it out.

dhvogel:
Otherwise this is merge-able

gino-m:
> @gino-m There is one change here that I would like your opinion on: I changed the default font to "Google Sans". It looks like the font used in the mocks (and the one that google uses). Let me know if you think this is a good change or not. If not we can take it out.

Google Sans is generally used for larger heading and titles in Google-branded products, so it wouldn't really be appropriate to use it here. Please feel free to change it in a follow up PR if you don't want to revert it in this one. Thanks for the fixes!

dhvogel:
reverted the font change, needs another lgtm

PR #132, opened 2019-09-12
Fix feature details store

gino-m:
There was a bug in the redux-firestore that caused it to accumulate results in `firestore.data` in the store. Upgrading the version fixed this. Also adding a few minor tweaks here.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -12003,6 +12003,11 @@
         "minimatch": "3.0.4"
       }
     },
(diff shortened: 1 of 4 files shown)


dhvogel:
so this will stop the side panel from accumulating records when multiple map points are clicked?

gino-m:
> so this will stop the side panel from accumulating records when multiple map points are clicked?

Yes, that's correct.

PR #137, opened 2019-09-15
when feature is selected its marker size should be bigger

gyarill:


gino-m:
I tried it out and it works. Very cool, thank you!

Would it be possible to set the `anchor` attribute so that the icon points to the location when small/large?
https://developers.google.com/maps/documentation/javascript/markers

gino-m:
> Would it be possible to set the `anchor` attribute so that the icon points to the location when small/large?
> https://developers.google.com/maps/documentation/javascript/markers

I just added it to this PR so I can merge it now. Thanks again!

Issue #116, opened 2019-09-10
[Feature panel] the X button is not working


gyarill:
Nothing happens when I click it

gino-m:
Obsolete; will be reimplemented in new version.

PR #139, opened 2019-09-16
Close drawer on click. Closes issue #116

gyarill:


gino-m:
I hacked something up for G4G, but this is much better (the "right" way), thanks for this!

PR #141, opened 2019-10-24
Add generated ng template code with README

DaoyuT:


web-ng/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 8.3.14.
+
+## Development server
+
+Run `ng serve` for a dev server. Navigate to `http://localhost:4200/`. The app will automatically reload if you change any of the source files.
+
+## Code scaffolding
+
+Run `ng generate component component-name` to generate a new component. You can also use `ng generate directive|pipe|service|class|guard|interface|enum|module`.
```
gino-m: Maybe mention there are generator plugins for popular text editors?

web-ng/angular.json
```diff
@@ -0,0 +1,125 @@
...
+          "configurations": {
+            "production": {
+              "devServerTarget": "web-ng:serve:production"
+            }
+          }
+        }
+      }
+    }},
+  "defaultProject": "web-ng"
+}
```
gino-m: Nit: Final LF

web-ng/browserslist
```diff
@@ -0,0 +1,12 @@
...
+# https://github.com/browserslist/browserslist#queries
+
+# You can see what browsers were selected by your queries by running:
+#   npx browserslist
+
+> 0.5%
+last 2 versions
+Firefox ESR
+not dead
+not IE 9-11 # For IE 9-11 support, remove 'not'.
```
gino-m: Nit: Final LF

web-ng/e2e/protractor.conf.js
```diff
@@ -0,0 +1,32 @@
...
+    defaultTimeoutInterval: 30000,
+    print: function() {}
+  },
+  onPrepare() {
+    require('ts-node').register({
+      project: require('path').join(__dirname, './tsconfig.json')
+    });
+    jasmine.getEnv().addReporter(new SpecReporter({ spec: { displayStacktrace: true } }));
+  }
+};
```
gino-m: LF

web-ng/e2e/src/app.po.ts
```diff
@@ -0,0 +1,11 @@
+import { browser, by, element } from 'protractor';
+
+export class AppPage {
```
gino-m: Please add a class comment so readers know what this models.

web-ng/karma.conf.js
```diff
@@ -0,0 +1,32 @@
+// Karma configuration file, see link for more information
```
gino-m: Please add license header in all JS files. Example:
https://github.com/google/ground-platform/blob/master/web/src/datastore.js

(Use 2019 for year)

web-ng/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "web-ng",
```
gino-m: "ground"?

web-ng/src/app/app-routing.module.ts
```diff
@@ -0,0 +1,11 @@
+import { NgModule } from '@angular/core';
+import { Routes, RouterModule } from '@angular/router';
+
+
```
gino-m: Extra line

web-ng/src/app/app.component.html
```diff
@@ -0,0 +1,538 @@
+<!-- * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * -->
```
gino-m: Please add copyright header in HTML files as well.

web-ng/src/app/app.component.html
```diff
@@ -0,0 +1,538 @@
...
+<!-- * * * * * * * * * * * The content below * * * * * * * * * * * -->
+<!-- * * * * * * * * * * is only a placeholder * * * * * * * * * * -->
+<!-- * * * * * * * * * * and can be replaced. * * * * * * * * * * * -->
+<!-- * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * -->
+<!-- * * * * * * * * * Delete the template below * * * * * * * * * * -->
+<!-- * * * * * * * to get started with your project! * * * * * * * * -->
+<!-- * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * -->
```
gino-m: Could you kindly replace this example boilerplate with a simple component representing the "main" page of the app? I assume we don't need all the styles and template here?

web-ng/tslint.json
```diff
@@ -0,0 +1,91 @@
+{
```
gino-m: Are these the same as the style defined by https://github.com/google/gts ?

web-ng/tslint.json
```diff
@@ -0,0 +1,91 @@
...
+    "no-outputs-metadata-property": true,
+    "template-banana-in-box": true,
+    "template-no-negated-async": true,
+    "use-lifecycle-interface": true,
+    "use-pipe-transform-interface": true
+  },
+  "rulesDirectory": [
+    "codelyzer"
+  ]
+}
```
gino-m: LF

gyarill:
> Overall LGTM, a few nits/suggestions. Thanks for setting up all this boilerplate!
> 
> * [ ]  @gyarill, could you kindly clone this repo and make sure you're able to stand up the app locally and also run tests?

so i followed readme and did

> npm install -g @angular/cli
> nvm install node --reinstall-packages-from=node

but then we i run 

> ng serve

i've got error so i had to run 

> npm install --save-dev @angular-devkit/build-angular

so you should probably add it to readme

with this ng serve run nicely and I got a pretty welcome page under http://localhost:4200/

so looks good to me but please add the missing command to README

Thank you for creating all of this!

web-ng/karma.conf.js
DaoyuT: Added, what about TS files?

web-ng/README.md
DaoyuT: I am not quite sure what you mean here. Can you elaborate more?

gino-m: I was suggesting to mention that they could use editor plugins to do the same, but I think we can leave that as an exercise for the reader.

web-ng/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "lib": [
+			"es2018",
+			"dom"
```
gino-m: Looks like this is using spaces instead of tabs. Is there a linter that would detect or fix this? Or are .json not linted?

web-ng/src/test.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * @license
```
gino-m: Please remove JDSoc @license from ts and css files.

web-ng/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+npm install --save-dev @angular-devkit/build-angular
+```
+
+Upgrade Node.js to the newest version using nvm: (Angular requires Node.js version 10.9.0 or later.)
+```
+nvm install node --reinstall-packages-from=node
+```
+
+# WebNg
```
gino-m: Maybe mention what WebNg is here.

web-ng/src/ground-maps/ground-maps.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'ground-maps',
+  templateUrl: './ground-maps.ng.html',
+  styleUrls: ['./ground-maps.css'],
+})
+export class GroundMaps {
```
gino-m: Nit: I assume instances of this class represent a single map? If so it should be named `GroundMap`.

Also, what's the convention on prefixing app-specific components in Angular? Should we choose a short prefix (gnd-*) and use it for app components?

DaoyuT: I think `GroundMap` make sense, I think I was thinking Google Maps and put `Maps` there. In terms of `ground-*` or `gnd-*`, I've only seen guides saying "don't abbreviate key words like 'module', 'service', etc.", I am okay with both, I can change them to `gnd` and let's stick to this.

gino-m: It seems the style guide recommends a prefix for area of the app, but not for the app name itself:

> Do use a prefix that identifies the feature area or the app itself.
> 
> Why? Prevents element name collisions with components in other apps and with native HTML elements.
> 
> Why? Makes it easier to promote and share the component in other apps.
> 
> Why? Components are easy to identify in the DOM.

Source: https://angular.io/guide/styleguide#component-custom-prefix

So in this we'd probably have a bunch of "Map" components, e.g. `MapContainer`, `MapControls`, etc.

web-ng/tsconfig.json
DaoyuT: No, I think you meant this is using tabs instead of spaces. I'll change it to spaces. The tslint.json generated by gts excluded all the .json files.

gino-m: Sorry, that's what I meant. Thanks for fixing it!

PR #165, opened 2019-11-11
Update docs and npm package versions

gino-m:
Fixed Node.js version, formatted README.me, and performed `npm update` on Angular deps. @DaoyuT could you please take a quick look?

I excluded mention of the node version required by Angular since that will change frequently.


web-ng/README.md
```diff
@@ -3,40 +3,43 @@
...
-Install angular-devkit:
-```
-npm install --save-dev @angular-devkit/build-angular
-```
+Install recent version of Node.js using nvm:
 
-Upgrade Node.js to the newest version using nvm: (Angular requires Node.js version 10.9.0 or later.)
 ```
-nvm install node --reinstall-packages-from=node
+nvm install 13.1.0 --reinstall-packages-from=node
```
DaoyuT: Why do we stick to a specific version instead of the newest one?

web-ng/README.md
```diff
@@ -3,40 +3,43 @@
...
 ```
 
 ## Development server
 
-Run `ng serve` for a dev server. Navigate to `http://localhost:4200/`. The app will automatically reload if you change any of the source files.
+Run `ng serve` for a dev server. Navigate to http://localhost:4200/. The app
+will automatically reload if you change any of the source files.
```
DaoyuT: How did you do this format this? I had the [google/gts](https://github.com/google/gts) configured for this project, and now we all could do a `npx gts fix` to fix format for most of the code, but it did not fix README. Can you share how you have your formatter configured?

web-ng/README.md
gino-m: A few reasons:
- So that our builds will be deterministic (especially CI builds and releases); at a particular commit the supported version is well-defined.
- For debugging developer issues. npm packages sometimes require different versions of node, and we'll get lots of questions from developers when they try to build with potentially newer versions than us.

Wdyt?

web-ng/README.md
gino-m: I believe I did it in VS Code using the default settings in the Prettier plugin. It doesn't look like gts supports Markdown.

PR #170, opened 2019-11-21
Replace Angular Google Maps(AGM) with Google Maps API

DaoyuT:
- [x] use google api instead of AGM
- [x] make it look like original web app
- [ ] ~~make API key injectable~~


web-ng/package.json
```diff
@@ -27,6 +26,7 @@
     "@angular/platform-browser": "8.2.14",
     "@angular/platform-browser-dynamic": "8.2.14",
     "@angular/router": "8.2.14",
+    "@types/googlemaps": "^3.38.0",
```
gino-m: Please use fixed version numbers (remove "^") for deterministic builds.

web-ng/src/index.html
```diff
@@ -25,5 +25,6 @@
 </head>
 <body>
   <app-root></app-root>
+  <script src="http://maps.googleapis.com/maps/api/js?key="></script>
```
gino-m: Is this basically used to import the APIs? Should/can this be placed inside the Map component?

gino-m: We spoke; `google` needs to be a global, so this needs to be defined at top level for now.

PR #173, opened 2019-11-22
Add data store, project service, and reactive header

gino-m:
- [x] Create immutable Project model object (with only title for now)
- [x] Create Project service for streaming latest version of currently active project
- [x] Connect active project with project id in URL using Angular router
- [x] Fetch actual Project title from Firestore
- [x] Read Firestore config from gitignored file in environment
- [x] Define datastore service
- [x] Manage subscriptions either using `Subscription` class or Angular `| async`

### DIFF
--- a/web-ng/src/app/services/project/project.service.ts
+++ b/web-ng/src/app/services/project/project.service.ts
@@ -15,22 +15,45 @@
  */
 
 import { Injectable } from '@angular/core';
-import { Observable, of, BehaviorSubject } from 'rxjs';
+import { Observable, ReplaySubject, Subscription } from 'rxjs';
+import { switchMap, map, shareReplay } from 'rxjs/operators';
 import { Project } from '../../shared/models/project.model';
-
-const EMPTY_PROJECT = new Project();
+import { AngularFirestore } from '@angular/fire/firestore';
 
 @Injectable({
   providedIn: 'root',
 })
 export class ProjectService {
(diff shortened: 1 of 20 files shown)


gino-m:
@DaoyuT, @gyarill, @bermessa, FYI.

gino-m:
I started implementing the above tasks, but realized I need to take a step back and review the Angular documentation to get a better handle on all the relevant concepts. Submitting this in the meantime because it can be useful if others want to build on top of it.

web-ng/src/app/services/project/project.service.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Injectable } from '@angular/core';
+import { Observable, ReplaySubject, Subscription } from 'rxjs';
+import { switchMap, map, shareReplay } from 'rxjs/operators';
+import { Project } from '../../shared/models/project.model';
+import { AngularFirestore } from '@angular/fire/firestore';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ProjectService {
```
gino-m: @DaoyuT , @gyarill FYI, working example loading project object from Firestore. Let's discuss!

gino-m:
@DaoyuT could you review and approve if it looks ok? I'll add tests in a follow up, just wanted to get a basic end-to-end flow working.

@gyarill @bermessa FYI, please take a look as well!

gino-m:
Thank you Daoyu!

PR #181, opened 2019-12-20
[Auth] Implement Sign In with Google using Firebase auth

gino-m:
UI is just a placeholder to allow sign in/out and for showing avatar.

Authorization to be set up in Firebase and client redirected if project not available. This will be done in future PRs.


gino-m:
@gyarill Added some comments, refactored data store methods, and updated tests (see commits above). PTAL?

gino-m:
Thanks for reviewing! 🎉

PR #187, opened 2019-12-29
[Data model] Introduce immutable model objects

### LINKED ISSUE #184: [Model] Make Project immutable
Likely candidate for modeling immutable collections:
https://immutable-js.github.io/immutable-js/

This involves several steps:
- [x] `npm install --save immutable`
- [x] Define `StringDictionary` as `Immutable.Map<string, string>` and rename to `StringMap`.
    https://github.com/google/ground-platform/blob/816483fccaed8aae2b8d48d3f2c99f994d250473/web-ng/src/app/shared/models/string-dictionary.model.ts#L20-L22
- [x] Create new `Project` from  `DocumentSnapshot` instead of casting:
    https://github.com/google/ground-platform/blob/7e478c250a50e7a29d6bd6ab498f3a9d334ca5dc/web-ng/src/app/services/data-store/data-store.service.ts#L37
- [x] Extra credit: create `Layer` object with `name: StringMap` and to `Project` as `layers: Map<string, Layer>`, building map at load time as above.

gino-m:
- [x] Imported immutable-js
- [x] Function to convert object to immutable object
- [x] Use `Immutable.Map` to represent strings and rename to `StringMap`
- [x] Loads immutable `Layer` object.
- [x] Adds uuids to `Project` and `Layer`

Fixes #184.

@shobhitagarwal1612, @gyarill, FYI.


web-ng/src/app/shared/models/project.model.ts
```diff
@@ -14,11 +14,32 @@
...
+   * immutable Project instance.
+   *
+   * @param id the uuid of the project instance.
+   * @param data the source data in a dictionary keyed by string.
+   */
+
+  static fromJson(id: string, data: { [key: string]: any }): Project {
+    return new Project(
+      id,
+      StringMap(data.title),
```
DaoyuT: I am not very familiar with the language, but if we put just `Map<string, string>` instead of `StringMap` here (and everywhere else presumably), will it still work? If it still works, I think having `StringMap` as additional layer wouldn't help much here. If not, it looks fine to me.

gino-m:
That would work too, but it's useful to have a binding for the semantics of
user translatable strings, which are used throughout the model. It would
also be helpful if we want to add logic specific to these types of maps.
Let's keep it now, and remove it turns out to be redundant?

PR #193, opened 2019-12-29
[Edit layer] Add skeleton "Customize layer" modal based on URL fragment

gino-m:
- [x] Show dialog when appropriate URL Fragment is set (`#l=123456`)
- [x] Parse fragment to get layer id
- [x] Pass layer id to dialog
- [x] Disable close on blur
- [x] Inject mocked dependencies into tests 

image

Depends on #187.


gino-m:
@gyarill, @DaoyuT FYI.

gino-m:
Hi @gyarill! Could you PTAL, since this overlaps a bit with your recent PR #194?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -0,0 +1,42 @@
...
+  ) {
+    this.layerId = data.layerId!;
+    // Disable closing on clicks outside of dialog.
+    dialogRef.disableClose = true;
+  }
+
+  ngOnInit() {}
+
+  closeDialog() {
+    this.dialogRef.close('Result');
```
gyarill: just curious.. what does it mean?

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -48,9 +52,30 @@ export class MainPageComponent implements OnInit {
...
   ngOnDestroy() {
     this.subscription.unsubscribe();
   }
+
+  private onFragmentChange(fragment?: string) {
+    if (!fragment) {
+      return;
+    }
+    const params = new URLSearchParams(fragment);
+    if (params.get('l')) {
```
gyarill: where does this 'l' come from?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: It means I forgot to delete the unused example code. Good catch!

web-ng/src/app/components/main-page/main-page.component.ts
gino-m: Added a comment to clarify: "The 'l' param is used to represent the layer id being edited."

We should probably encapsulate these by wrapping the URL state and router in our objects, but I'll save that for future work for now.

PR #194, opened 2019-12-31
[Side panel] Set up and add empty container for upcoming components

gyarill:


web-ng/src/app/components/main-page/main-page.component.html
```diff
@@ -13,8 +13,14 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-
-<div *ngIf="(activeProject$ | async) as project">
-  <div style="height: 48px">{{ project.title[lang] }}</div>
-</div>
-<ground-map></ground-map>
+<mat-sidenav-container class="main-page-container">
+  <mat-sidenav #sidenav mode="over" position="end" [(opened)]="isSidePanelOpen">
```
gino-m: Nit: Can we move `mat-sidenav` after `mat-sidenav-content`? Seems to be more intuitive since `content` is the main page area, while `sidenav` is the secondary view area. It also appears second on screen if visually scanning the page from left to right.

web-ng/src/app/components/side-panel/side-panel.component.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ProjectService } from '../../services/project/project.service';
+
+@Component({
+  selector: 'ground-side-panel',
+  templateUrl: './side-panel.component.html',
+  styleUrls: ['./side-panel.component.css']
+})
+export class SidePanelComponent {
+  activeProject$: Observable<Project>;
+  lang: string;
```
gino-m: Please mark both these fields `readonly` so they can't be modified (similar to `final` in Java).

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -15,10 +15,16 @@
...
+  [key: string]: Layer;
+}
 
 export class Project {
   constructor(
     readonly title: StringDictionary,
-    readonly description: StringDictionary
-  ) {}
+    readonly description: StringDictionary,
+    readonly layers: LayerMap,
```
gino-m: In `master` I have this represented as an `Immutable.List<Layer>`. If it's more convenient to leave it as a map we can do that as well. If so, when merging please replace this with an `Immutable.Map` instead.

gyarill: List sounds good:)

web-ng/src/app/components/main-page/main-page.component.html
gyarill: done

web-ng/src/app/components/side-panel/side-panel.component.ts
gyarill: done

web-ng/angular.json
```diff
@@ -24,6 +24,7 @@
               "src/assets"
             ],
             "styles": [
+              "./node_modules/@angular/material/prebuilt-themes/indigo-pink.css",
```
gino-m: Please remove the indigo-pink.css here and below - we'll use our own custom style in src/styles.css.

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -24,7 +24,7 @@ export class Project {
     readonly title: StringMap,
     readonly description: StringMap,
     readonly layers: List<Layer>
-  ) {}
+  ) { }
```
gino-m: Could you try using Prettier to reformat if you're not already? It should use eslint, which in turn will apply Google TypeScript formatting (gts).

web-ng/src/app/components/side-panel/side-panel.module.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright 2019 Google LLC
```
gino-m: Please update to 2020 for all files added in this PR.

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -42,6 +43,7 @@ export class MainPageComponent implements OnInit {
   ) {
     // TODO: Make dynamic to support i18n.
     this.lang = 'en';
+    this.isSidePanelOpen = true;
```
gino-m: Might be better to name this `sideNavOpened` for consistency with component/attribute name and to avoid confusing name with a function.

gyarill:
Thank you Gino for looking at it! 

So it seems that  feature.service.spec.ts is failing with "FeatureService should be created FAILED". I am not really sure how my cl could affect it

gyarill:
Ah i do not know why this trigger check is failing. Is there a way to check?

gino-m:
> Ah i do not know why this trigger check is failing. Is there a way to check?

It seems some lint checks are failing:
https://console.cloud.google.com/cloud-build/builds/b80fbaf9-0c37-456f-b8f4-ac1d8f54c5f7;step=4?project=ground-android-gcb

You can find that by clicking "Details" next to the failing check, and clicking through the link at the bottom of the page that appears.

gino-m:
P.S. `npm run fix` should reformat the code to spec.

gino-m:
> Thank you Gino for looking at it!
> 
> So it seems that feature.service.spec.ts is failing with "FeatureService should be created FAILED". I am not really sure how my cl could affect it
> 
> But I fixed it.

Thanks for fixing this test! It may have already been broken (!) We'll set up CI to make sure all tests pass.

gyarill:
When I click the link i see "You do not have sufficient permissions to view this page"

"npm run fix" did not help:(

gino-m:
> When I click the link i see "You do not have sufficient permissions to view this page"
> 
> "npm run fix" did not help:(

Checks are passing!

PR #199, opened 2020-01-12
[Auth] Create user profile server-side instead of client-side

gino-m:
Writes user profile data to `/user/{uid}` in Cloud Firestore.

Since admin user is now used to create user profile, user does't need permission to write to it directly, simplifying permissions model.

This also now causes profiles to be created for users who first sign in using Android app.

### DIFF
--- a/web-ng/src/app/services/data-store/data-store.service.ts
+++ b/web-ng/src/app/services/data-store/data-store.service.ts
@@ -55,22 +55,4 @@ export class DataStoreService {
   user$(uid: string): Observable<User | undefined> {
     return this.db.doc<User>(`users/${uid}`).valueChanges();
   }
- 
-  /**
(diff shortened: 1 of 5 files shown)


gino-m:
@DaoyuT gentle ping. Could you PTAL?

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -57,33 +57,4 @@ export class DataStoreService {
...
-        uid,
-        email,
-        displayName,
-        photoURL,
-      },
-      { merge: true }
-    );
-  }
-
-  features$({ id }: Project): Observable<List<Feature>> {
```
gino-m: @DaoyuT Sorry I accidentally deleted this here. Adding it back in #203.

PR #201, opened 2020-01-16
Show markers on map unstyled

DaoyuT:


web-ng/src/app/components/map/map.component.ts
```diff
@@ -23,6 +26,11 @@ import { Component } from '@angular/core';
 })
 export class MapComponent {
   zoom = 3;
-  lat = 40.767716;
-  lng = -73.971714;
+  features$: Observable<Feature[]>;
```
gino-m: Please use ImmutableList<Feature> instead

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -73,4 +74,21 @@ export class DataStoreService {
       { merge: true }
     );
   }
+
+  loadFeatures$({ id }: Project): Observable<Feature[]> {
+    return this.db
+    .collection(`projects/${id}/features`)
+    .get()
```
gino-m: Please use `valueChanges()` instead

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -73,4 +74,21 @@ export class DataStoreService {
...
+    .pipe(
+      map(
+        querySnapshot => {
+          var features: Feature[] = [];
+          querySnapshot.docs.forEach(doc => {
```
gino-m: This can be simplified using array.map()..

PR #209, opened 2020-01-31
[side-panel] create skeleton for displaying observation data. Also: load feature data from firebase

gyarill:
.


web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -71,6 +95,33 @@ export class DataStoreService {
...
+   */
+  observations$(
+    project: Project,
+    feature: Feature
+  ): Observable<List<Observation>> {
+    return this.db
+      .collection(`projects/${project.id}/observations`, ref =>
+        ref.where('featureId', '==', feature.id)
+      )
+      .valueChanges({ idField: 'id' }) // what to put here?
```
gino-m: Please remove comment

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -95,18 +151,100 @@ export class DataStoreService {
...
+   *     // ...
+   *   }
+   */
+  private static toField(id: string, data: DocumentData): Field {
+    return new Field(
+      id,
+      FieldType.TEXT,
+      StringMap(data.label),
+      data.required,
+      data.options ? new MultipleChoice(
```
gino-m: I think you can use `data.options && new MultipleChoice(..)` with the ternary operator.

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -120,4 +258,59 @@ export class DataStoreService {
   private static toFeature(id: string, data: DocumentData): Feature {
     return new Feature(id, data.layerId, data.location);
   }
+
+  /**
+   * Converts the raw object representation deserialized from Firebase into an
+   * immutable Observation instance.
+   *
+   * @param data the source data in a dictionary keyed by string.
+   * {
```
gino-m: Please surround this and other examples with `<pre><code></code></pre>`

web-ng/src/app/services/observation/observation.service.ts
```diff
@@ -0,0 +1,40 @@
+import { Feature } from './../../shared/models/feature.model';
+import { Project } from './../../shared/models/project.model';
```
gino-m: Move imports below comments (the VSCode plugin sometimes misplaces the auto-imports :/)

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -15,14 +15,27 @@
...
 export class Project {
   constructor(
     readonly id: string,
     readonly title: StringMap,
     readonly description: StringMap,
-    readonly layers: List<Layer>
+    readonly layers: Map<string, Layer>
   ) {}
+
+  getForm(layerId: string,  formId: string): Form {
```
gino-m: Nit: Extra space here, missing spaces after "if", use single instead of double quotes.

`npm run fix` should fix these.

web-ng/src/app/components/layer-list/layer-list.component.html
```diff
@@ -0,0 +1,21 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div *ngIf="activeProject$ | async as project">
+  <div *ngFor="let entry of project.layers">
```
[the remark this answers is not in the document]
gyarill: I used to have it like that but then console was complaining that  project.layers.keys() is changing and I should not use it in the ngFor loop. I would need to bind it to a variable inside of the class I think but i think just taking entry[1] is nicer for now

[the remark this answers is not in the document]
gino-m: Sgtm!

web-ng/tsconfig.app.json
```diff
@@ -9,10 +9,12 @@
     "src/polyfills.ts"
   ],
   "include": [
-    "src/**/*.ts"
+    "src/**/*.ts",
+    "src/**/**/*.ts"
```
gino-m: `**/` is already recursive.. are you sure this is needed?

gyarill: so i did it because my ng test was complaining that is does not see color-picker test file after merge but it does not seem to be a problem now

gyarill: could you approve again?:P

PR #210, opened 2020-02-01
Update URL fragment on click marker, enlarge selected marker in respo…

DaoyuT:
…nse to fragment change in URL.


web-ng/src/app/components/map/map.component.ts
```diff
@@ -14,22 +14,63 @@
...
+  styleUrls: ['./map.component.css'],
 })
-export class MapComponent {
+export class MapComponent implements OnInit {
+  subscription: Subscription = new Subscription();
   zoom = 3;
+  focusedFeatureId = '';
   features$: Observable<List<Feature>>;
+  icon = {
+    url: 'http://maps.google.com/mapfiles/kml/paddle/red-circle.png',
```
gino-m: Can we import these icons and serve them from our app instead of from google.com?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -14,22 +14,63 @@
...
 
 @Component({
   selector: 'ground-map',
   templateUrl: './map.component.html',
-  styleUrls: ['./map.css'],
+  styleUrls: ['./map.component.css'],
 })
-export class MapComponent {
+export class MapComponent implements OnInit {
+  subscription: Subscription = new Subscription();
```
gino-m: Please make relevant fields private and/or readonly.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -14,22 +14,63 @@
...
+      this.route.fragment.subscribe(fragment => {
+        const params = new HttpParams({ fromString: fragment });
+        if (params.get('f')) {
+          this.focusedFeatureId = params.get('f')!;
+        }
+      })
+    );
+  }
+
+  featureDetails(featureId: string) {
```
gino-m: Please name to "onFeatureClick" so it's clear it's an event handler.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -14,22 +14,63 @@
...
+    const primaryUrl = this.router
+      .parseUrl(this.router.url)
+      .root.children['primary'].toString();
+    const navigationExtras: NavigationExtras = { fragment: `f=${featureId}` };
```
gino-m: Please add a TODO to refactor URL read/write logic into its own service.

DaoyuT: Added, but what do you mean by that?

gino-m: Right now multiple code points interact directly with the router to fetch project id, feature id, and layer id, and to determine current view state. This "leaks" the details that we're using a particular URL router, URL format, etc. to multiple parts of of the code base. I was suggesting we encapsulate and abstract this behavior, so that component don't need to know _how_ the state is set and retrieved.

PR #213, opened 2020-02-10
Tolerate missing dictionaries in project definition

### LINKED ISSUE #212: [Load project] Loading project with no layers fails
**To Reproduce**
Create a new project or open a project with no layers defined.

Exception:
```
ERROR TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Function.toLayer (data-store.service.ts:163)
    at data-store.service.ts:145
    at Array.map (<anonymous>)
    at Function.toProject (data-store.service.ts:143)
    at MapSubscriber.project (data-store.service.ts:56)
    at MapSubscriber._next (map.js:29)
    at MapSubscriber.next (Subscriber.js:49)
    at angularfire2.js:58
    at ZoneDelegate.invoke (zone-evergreen.js:359)
```

Need to handle null case here - can probably do  `layers || []` to default to empty list if null.

gino-m:
Fixes #212.

### DIFF
--- a/web-ng/src/app/services/data-store/data-store.service.ts
+++ b/web-ng/src/app/services/data-store/data-store.service.ts
@@ -32,6 +32,14 @@ import { Observation } from '../../shared/models/observation/observation.model';
 import { Response } from '../../shared/models/observation/response.model';
 import { StringMap } from './../../shared/models/string-map.model';
 
+/**
+ * Helper to return either the keys of a dictionary, or if missing, returns an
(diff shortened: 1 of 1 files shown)


gino-m:
> Do we need to do the same with toForm and toField function if there is no elements or options?

Good catch, pushing an update.

gino-m:
Should be ok now, PTAL!

PR #231, opened 2020-02-20
Read marker color from its layer in active project.

DaoyuT:


web-ng/src/app/components/map/ground-pin-svg-template.ts
```diff
@@ -0,0 +1,28 @@
+export const svgTemplate = `<?xml version="1.0" encoding="UTF-8"?>
```
gino-m: Design nit: Could we make this a function that returns the SVG with the appropriate color already set? This way the template placeholder doesn't need to be exposed outside this file.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -74,4 +73,35 @@ export class MapComponent implements OnInit {
...
+    focusedFeatureId: string,
+    project: Project
+  ): google.maps.MarkerOptions {
+    // Icon is not yet an input for <map-marker>, this is the only way to change icon for now.
+    // Consider break this down when more inputs are available for <map-marker>.
+    const normalScale = 30;
+    const enlargedScale = 50;
+    const defaultIconColor = 'red';
+    const color =
+      project.layers.get(feature.layerId)!.color || defaultIconColor;
```
gino-m: Doe we want `?` instead of `!` here to avoid failing catastrophically if the data of a single feature is corrupt?

Issue #232, opened 2020-02-21
[Layer dialog] Load/save layer model


gino-m:
Implementation order as discussed
- [ ] Add `layer` field to `LayerDialogComponent`. This will allow @DaoyuT to start implementing UI in parallel with the next steps:
- [ ] Add "Save" button that overwrites layer in current project in Firestore. If `layerId == ':new'` replace it with a UUID (can be generated using Firebase APIs - just be sure to define function in datastore service).
- [ ] When opening dialog, if `layerId != ':new'`, copy relevant layer from active Project definition in component (e.g. in constructor?). This may require reading the currently active project.

PR #233, opened 2020-02-24
[Layer dialog] Load/save layer model

parulraheja98:
Refs: #232

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -77,14 +77,13 @@ export class LayerDialogComponent implements OnDestroy {
     const layer = {
       id: this.layerId,
     };
-    try {
-      // TODO: Inform user layer was saved
-      this.dataStoreService
-        .updateProjectLayer(this.projectId, layer)
-        .then(() => this.onClose());
-    } catch (err) {
-      alert('Layer update failed.');
-    }
+    // TODO: Inform user layer was saved
+    this.dataStoreService
+      .updateProjectLayer(this.projectId, layer)
+      .then(() => this.onClose())
+      .catch(err => {
+        alert('Layer update failed.');
+      });
   }
 
   onClose() {
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -78,6 +78,7 @@ export class LayerDialogComponent implements OnDestroy {
       id: this.layerId,
     };
     try {
+      // TODO: Inform user layer was saved
       this.dataStoreService
         .updateProjectLayer(this.projectId, layer)
         .then(() => this.onClose());
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -82,6 +82,7 @@ export class LayerDialogComponent implements OnDestroy {
       .updateProjectLayer(this.projectId, layer)
       .then(() => this.onClose())
       .catch(err => {
+        // TODO: Show in toast message, i18n.
         alert('Layer update failed.');
       });
   }
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -70,17 +70,14 @@ export class LayerDialogComponent implements OnDestroy {
   }
 
   onSave() {
+    // TODO  wait for project to load before showing dialog
     if (!this.projectId) {
-      throw Error('No Project exists with that id');
+      throw Error('Project not yet loaded');
     }
     const layer = {
       id: this.layerId,
     };
-    this.dataStoreService.updateProjectLayer(
-      this.projectId,
-      this.layerId,
-      layer
-    );
+    this.dataStoreService.updateProjectLayer(this.projectId, layer);
     this.onClose();
   }
 
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -14,24 +14,74 @@
  * limitations under the License.
  */
 
-import { Component, Inject, OnInit } from '@angular/core';
+import { Component, Inject, OnDestroy } from '@angular/core';
+import { Observable } from 'rxjs';
+import { ProjectService } from '../../services/project/project.service';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { Project } from '../../shared/models/project.model';
+import { Layer } from '../../shared/models/layer.model';
+import { Subscription } from 'rxjs';
+import { DataStoreService } from '../../services/data-store/data-store.service';
+import { Router } from '@angular/router';
 
 @Component({
   selector: 'app-layer-dialog',
   templateUrl: './layer-dialog.component.html',
   styleUrls: ['./layer-dialog.component.css'],
 })
-export class LayerDialogComponent {
+export class LayerDialogComponent implements OnDestroy {
   layerId: string;
-
+  layer!: Layer;
+  project!: Project;
+  activeProject$!: Observable<Project>;
+  subscription: Subscription = new Subscription();
   constructor(
     // tslint:disable-next-line:no-any
     @Inject(MAT_DIALOG_DATA) data: any,
-    private dialogRef: MatDialogRef<LayerDialogComponent>
+    private dialogRef: MatDialogRef<LayerDialogComponent>,
+    private projectService: ProjectService,
+    private dataStoreService: DataStoreService,
+    private router: Router
   ) {
-    this.layerId = data.layerId!;
     // Disable closing on clicks outside of dialog.
     dialogRef.disableClose = true;
+    this.layerId = data.layerId;
+    this.activeProject$ = this.projectService.getActiveProject$();
+    this.subscription.add(
+      this.activeProject$.subscribe(project => {
+        this.layer = project.layers.get(this.layerId)!;
+        this.project = project;
+      })
+    );
+  }
+
+  onSave() {
+    let updatedLayers;
+    if (this.layerId === ':new') {
+      updatedLayers = this.project.layers.set(
+        this.dataStoreService.generateId(),
+        {
+          id: this.dataStoreService.generateId(),
+        }
+      );
+    } else {
+      updatedLayers = this.project.layers.set(this.layerId, {
+        id: this.layerId,
+      });
+    }
+    this.dataStoreService.updateProjectLayer(
+      this.project.id,
+      JSON.stringify(updatedLayers)
+    );
+    this.onClose();
+  }
+
+  onClose() {
+    this.dialogRef.close();
+    return this.router.navigate([`p/${this.project.id}`]);
+  }
+
+  ngOnDestroy() {
+    this.subscription.unsubscribe();
   }
 }
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -33,7 +33,7 @@ export class LayerDialogComponent implements OnDestroy {
   layerId: string;
   layer!: Layer;
   project!: Project;
-  activeProject$!: Observable<Project>;
+  activeProject$: Observable<Project>;
   subscription: Subscription = new Subscription();
   constructor(
     // tslint:disable-next-line:no-any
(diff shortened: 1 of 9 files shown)


web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,74 @@
...
   selector: 'app-layer-dialog',
   templateUrl: './layer-dialog.component.html',
   styleUrls: ['./layer-dialog.component.css'],
 })
-export class LayerDialogComponent {
+export class LayerDialogComponent implements OnDestroy {
   layerId: string;
-
+  layer!: Layer;
+  project!: Project;
```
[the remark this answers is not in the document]
parulraheja98: It is just a typescript thing because it doesn't recognizes assigning of variables inside subscribe

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -58,26 +58,21 @@ export class LayerDialogComponent implements OnDestroy {
...
     let updatedLayers;
     if (this.layerId === ':new') {
-      updatedLayers = this.project.layers.set(
-        this.dataStoreService.generateId(),
-        {
-          id: this.dataStoreService.generateId(),
-        }
-      );
+      updatedLayers = {
+        [`layers.${this.dataStoreService.generateId()}`]: {},
```
gino-m: Nit: Could we make this more D.R.Y. by assigning id to a local var, e.g. 
`const layerId = this.layerId === ':new' ? this.dataStoreService.generateId() : this.layerId;`

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -74,18 +74,18 @@ export class DataStoreService {
       .then(() => projectId);
   }
 
-  updateProjectLayer(projectId: string, layers: string) {
-    const updatedLayers = JSON.parse(layers);
+  // TODO: Define return types for methods in this class
+  updateProjectLayer(projectId: string, layers: {}) {
```
gino-m: By accepting an `any` object with path:value pairs, this method exposes Firebase's API for updating data. DatastoreService abstracts away all Firebase-specific details (Separation of Concerns), so this method should really just take the following:
`updateProjectLayer(projectId: string, layerId: string, layer: Layer)`

And build the dot.notation.path itself instead of having the caller do it as above.

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -74,18 +74,18 @@ export class DataStoreService {
...
     return new Promise((resolve, reject) => {
       this.db
         .collection('projects')
         .doc(projectId)
-        .set({ layers: updatedLayers }, { merge: true })
+        .update(layers)
```
gino-m: This method should remove `id` from the Layer before writing it, since unlike in our application model, the database "Layer" object doesn't contain `id` as a member.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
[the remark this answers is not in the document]
gino-m: Not sure I follow.. What do you mean by "recognize assigning of variables" and "inside subscribe"?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,69 @@
...
   ) {
-    this.layerId = data.layerId!;
     // Disable closing on clicks outside of dialog.
     dialogRef.disableClose = true;
+    this.layerId = data.layerId;
+    this.activeProject$ = this.projectService.getActiveProject$();
+    this.subscription.add(
+      this.activeProject$.subscribe(project => {
+        this.layer = project.layers.get(this.layerId)!;
+        this.project = project;
```
gino-m: Let's replace this with `this.projectId`.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,69 @@
...
+    private router: Router
   ) {
-    this.layerId = data.layerId!;
     // Disable closing on clicks outside of dialog.
     dialogRef.disableClose = true;
+    this.layerId = data.layerId;
+    this.activeProject$ = this.projectService.getActiveProject$();
+    this.subscription.add(
+      this.activeProject$.subscribe(project => {
+        this.layer = project.layers.get(this.layerId)!;
```
gino-m: Via chat I was suggesting we set the layerId when creating the model object so that it's never null. So here you would check if this.layerId === "new" and create an empty one with a generated id, otherwise you would load from the existing ones and report and error if layer.get() returns null.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,69 @@
...
+        this.projectId = project.id;
+      })
+    );
+  }
+
+  onSave() {
+    if (!this.projectId) {
+      return;
+    }
+    const layerId =
```
gino-m: In that case this would move into the callback in the constructor, or even better, the subscribe lambda can be moved into a function this.onProjectLoaded(project: Project).

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,69 @@
...
+      this.layerId === ':new'
+        ? this.dataStoreService.generateId()
+        : this.layerId;
+    const layer = {
+      id: layerId,
+    };
+    this.dataStoreService.updateProjectLayer(this.projectId, layerId, layer);
```
gino-m: As discussed we shouldn't need to pass layerId here.

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -73,6 +73,25 @@ export class DataStoreService {
       .set({ title: { en: newTitle } }, { merge: true })
       .then(() => projectId);
   }
+
+  // TODO: Define return types for methods in this class
+  updateProjectLayer(projectId: string, layerId: string, layer: Layer) {
+    return new Promise((resolve, reject) => {
```
gino-m: Immediately before doing these you should copy `Layer` in an `any` and remove the `id` field

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
parulraheja98: oops my bad I will fix it , thanks @gino-m

web-ng/src/app/services/data-store/data-store.service.ts
parulraheja98: I could have used any here but it would have given me typescript error and the build would have failed , so I opted to use { id: string } which is safer as we are expecting id as one of fields and removing it :)

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,83 @@
...
+    }
+    if (!this.layer) {
+      throw Error('No layer exists');
+    }
+    this.projectId = project.id;
+  }
+
+  onSave() {
+    if (!this.projectId) {
+      throw Error('No Project exists with that id');
```
gino-m: This will only happen if we somehow end up here before the project is loaded (not when no project exists with this id). This could happen if the user opens a bookmark with the layer dialog open and clicks "Save" before the project config finished loading. Ideally, we wouldn't open the layer dialog until the project dialog is loaded, which can be achieved by waiting for a "project' to be loaded before calling `showEditLayerDialog()` in `MainPageComponent`.

For now, could you kindly update the error message to 'Project not yet loaded' and add a TODO to wait for project to load before showing dialog?

web-ng/src/app/services/data-store/data-store.service.ts
gino-m: Cool trick, didn't know you could do this. Thanks!

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -73,6 +73,27 @@ export class DataStoreService {
       .set({ title: { en: newTitle } }, { merge: true })
       .then(() => projectId);
   }
+
+  // TODO: Define return types for methods in this class
+  updateProjectLayer(projectId: string, layerId: string, layer: Layer) {
+    const updatedLayer = layer as { id: string };
```
gino-m: Since in TypeScript objects are passed by reference, this will modified to provided Layer instead of making a copy. With the new ES7 spread operator you can do some magic:
```
  const { id: layerId, ...layerDoc } = layer;
```

That will give you two variables, one with the id, and `layerDoc` with everything else.

You can also remove `layerId` from the function params.

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -29,11 +29,14 @@ export class Project {
 
   getForm(layerId: string, formId: string): Form {
     const layer = this.layers.get(layerId);
+    let form;
     if (!layer) {
       throw Error('Layer not found.');
     }
-    const form = layer.forms.get(formId);
-    if (!form) {
+    if (layer.forms) {
```
gino-m: This block can be replaced with:
`form = layer.forms?.get(formId);`

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -29,11 +29,14 @@ export class Project {
...
+      form = layer.forms.get(formId);
+    }
+    if (!layer.forms || !form) {
```
gino-m: We've already checked `layer.forms` above, so let's only check the output `form` here.

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -73,6 +73,27 @@ export class DataStoreService {
...
+    delete updatedLayer['id'];
+    return new Promise((resolve, reject) => {
+      this.db
```
gino-m: `update` already returns a `Promise` - can we remove the `then` `catch` clause and return the Promise directly?

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -73,6 +73,19 @@ export class DataStoreService {
...
+  // TODO: Define return types for methods in this class
+  updateProjectLayer(projectId: string, layer: Layer) {
+    const { id: layerId, ...layerDoc } = layer;
+    return this.db
+      .collection('projects')
+      .doc(projectId)
+      .update({
+        [`layers.${layerId}`]: layerDoc,
+      })
+      .then(() => projectId);
```
gino-m: Why do we need the `then` clause here?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -77,7 +77,11 @@ export class LayerDialogComponent implements OnDestroy {
     const layer = {
       id: this.layerId,
     };
-    this.dataStoreService.updateProjectLayer(this.projectId, layer);
+    try {
```
gino-m: This function won't throw an exception, it'll return a Promise that fails with error.

Also, I assume we'll want to close the dialog and let the user know the layer was saved on completion? (feel free to add a `// TODO: ....` for now for that part).

parulraheja98: thank you for catching this :)

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -14,24 +14,86 @@
...
+    }
+    const layer = {
+      id: this.layerId,
+    };
+    try {
+      // TODO: Inform user layer was saved
+      this.dataStoreService
+        .updateProjectLayer(this.projectId, layer)
+        .then(() => this.onClose());
+    } catch (err) {
```
gino-m: You'll need to catch the error in .catch(). Not sure if JavaScript throw/catch will do here.

parulraheja98: I tried passing bad data value for updated and it is getting caught in the catch block that I created :). Also when I passed bad data value for update it didn't got caught in the .catch() block

gino-m: That's really odd! Can you paste the stack trace here so we can figure out where the exception is being thrown?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: Very funny :)

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: See also Firestore API docs where they show how to catch errors in the returned Promise:
https://firebase.google.com/docs/firestore/manage-data/add-data#update-data

PR #248, opened 2020-03-18
Observation list and layer list UI styling

gyarill:


gyarill:
Screen Shot 2020-03-18 at 5 57 51 PM

web-ng/src/app/components/map/ground-pin.ts
```diff
@@ -14,8 +14,14 @@
...
-export const renderPin = (color: string) => {
-  return svgTemplate.replace('{{ color }}', color);
+export const defaultIconColor = 'red';
+
+export const renderPin = (color: string | undefined) => {
+  return svgTemplate.replace('{{ color }}', color || defaultIconColor);
+};
+
+export const getPinSrc = (color: string | undefined) => {
+  return '/assets/img/pin-red.svg';
```
gino-m: Can you we renderPin() here instead of having a static marker?

gyarill: done

web-ng/src/app/components/map/map.component.ts
```diff
@@ -83,11 +83,9 @@ export class MapComponent implements OnInit {
     // Consider break this down when more inputs are available for <map-marker>.
     const normalScale = 30;
     const enlargedScale = 50;
-    const defaultIconColor = 'red';
-    const color =
-      project.layers.get(feature.layerId)?.color || defaultIconColor;
+    const color = project.layers.get(feature.layerId)?.color;
     const icon = {
-      url: 'data:image/svg+xml;charset=UTF-8;base64,' + btoa(renderPin(color)),
+      url: getPinDangerousSrc(color),
```
gino-m: Why DangerousSrc?

gino-m:
@DaoyuT could you PTAL?

web-ng/src/app/components/map/map.component.ts
DaoyuT: +1

PR #249, opened 2020-03-18
Add inline editor and use it to edit layer title

DaoyuT:


web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -76,13 +76,14 @@ export class DataStoreService {
...
   updateLayer(projectId: string, layer: Layer) {
-    const { id: layerId, forms, ...layerDoc } = layer;
+    const { id: layerId, name, forms, ...layerDoc } = layer;
 
     return this.db
       .collection('projects')
       .doc(projectId)
       .update({
         [`layers.${layerId}`]: {
+          name: name?.toJS() || {},
```
gino-m: Is this necessary? I believe it's already included in `layerDoc`?

web-ng/src/app/components/layer-dialog/layer-dialog.component.html
```diff
@@ -14,7 +14,12 @@
   limitations under the License.
 -->
 
-<h2 mat-dialog-title>Customize layer</h2>
+<inline-editor 
+  mat-dialog-title
+  class="layer-name"
+  [data]="layerName" 
+  placeholder='Unnamed layer' 
+  (focusOut)='saveLayerName($event)'></inline-editor>
```
gino-m: Naming: Perhaps `setLayerName` would be more appropriate, since it doesn't actually write to the db?

web-ng/src/app/services/data-store/data-store.service.ts
DaoyuT: yes, name is a StringMap(Map<string, string>), it needs a `toJS()` for the same reason `forms` needs `toJS()`

gino-m: Oooic.. Thanks for reminding!

DaoyuT:
> @DaoyuT Excellent work! Just a few nits, also some failing code checks/lint.
> 
> @gyarill could you also kindly review?

Sorry, I was trying to squash the lint commit into its previous commit so that not commits has the ❌, but I accidentally added a commit prior to this PR into this PR.

gino-m:
> > @DaoyuT Excellent work! Just a few nits, also some failing code checks/lint.
> > @gyarill could you also kindly review?
> 
> Sorry, I was trying to squash the lint commit into its previous commit so that not commits has the ❌, but I accidentally added a commit prior to this PR into this PR.

No worries about the "X"s in previous commits - unless something leaked there's rarely need to squash, imo. Thanks for trying to keep the PRs clean!

gino-m:
@gyarill could you PTAL?

Issue #223, opened 2020-02-20
[Form editor] “Add a question” button adds a new question block


gino-m:
Default to type "Text"

![image](https://user-images.githubusercontent.com/228050/74902391-f4b3f400-5373-11ea-8e19-056ccdc2a803.png)

PR #250, opened 2020-03-22
[Form editor] “Add a question” button adds a new question block

parulraheja98:
Refs: #223 
image

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.css
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.css
@@ -15,8 +15,7 @@
  */
 
 .mat-dialog-content {
-  width: 800px;
-  height: 600px;
+  width: 600px;
 }
 
 .btn-save {
@@ -36,6 +35,7 @@
   border-radius: 5px;
   font-family: 'Open Sans', Arial, sans-serif;
   cursor: pointer;
+  color: green;
 }
 
 .layer-name {
@@ -53,9 +53,10 @@
 }
 
 .add-question {
-  border-style: outset;
-  padding: 15px;
+  padding: 20px;
+  border: 1px solid lightgray;
   border-radius: 10px;
+  margin-top: 20px;
 }
 
 .plus-icon {
@@ -81,4 +82,19 @@
 .icon-field-type {
   vertical-align: middle;
   margin-right: 20px;
+}
+
+.container-question {
+  border-radius: 10px;
+  padding-top: 20px;
+  padding-left: 10px;
+  padding-right: 10px;
+  border: 0px thin;
+  margin-top: 10px;
+  border: 1px lightgray;
+  border-style: solid;
+}
+
+.mat-dialog-actions {
+  margin-top: 10px;
 }
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.css
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.css
@@ -25,7 +25,7 @@
   height: 36px;
   width: 82px;
   border-radius: 5px;
-  font-family: "Open Sans", Arial, sans-serif;
+  font-family: 'Open Sans', Arial, sans-serif;
   margin-left: 10px;
   cursor: pointer;
 }
@@ -34,7 +34,7 @@
   height: 36px;
   width: 82px;
   border-radius: 5px;
-  font-family: "Open Sans", Arial, sans-serif;
+  font-family: 'Open Sans', Arial, sans-serif;
   cursor: pointer;
 }
 
@@ -46,4 +46,39 @@
   margin-left: 0;
   margin-right: 0;
   font-weight: bold;
+}
+.link-add-question {
+  color: green;
+  cursor: pointer;
+}
+
+.add-question {
+  border-style: outset;
+  padding: 15px;
+  border-radius: 10px;
+}
+
+.plus-icon {
+  display: inline-block;
+  width: 20px;
+  height: 20px;
+  vertical-align: top;
+  background: linear-gradient(green, green), linear-gradient(green, green);
+  background-position: center;
+  background-size: 50% 2px, 2px 50%;
+  background-repeat: no-repeat;
+}
+
+.field-type {
+  margin-left: 30px;
+  font-size: 12px
+}
+
+.main-container-questions {
+  margin-top: 10px;
+}
+
+.icon-field-type {
+  vertical-align: middle;
+  margin-right: 20px;
 }
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.css
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.css
@@ -35,7 +35,7 @@
   border-radius: 5px;
   font-family: 'Open Sans', Arial, sans-serif;
   cursor: pointer;
-  color: green;
+  color: #009113;
 }
 
 .layer-name {
@@ -47,14 +47,15 @@
   margin-right: 0;
   font-weight: bold;
 }
+
 .link-add-question {
-  color: green;
+  color: #009113;
   cursor: pointer;
 }
 
 .add-question {
   padding: 20px;
-  border: 1px solid lightgray;
+  border: 1px solid #e6e7ea;
   border-radius: 10px;
   margin-top: 20px;
 }
@@ -64,7 +65,8 @@
   width: 20px;
   height: 20px;
   vertical-align: top;
-  background: linear-gradient(green, green), linear-gradient(green, green);
+  background: linear-gradient(#009113, #009113),
+    linear-gradient(#009113, #009113);
   background-position: center;
   background-size: 50% 2px, 2px 50%;
   background-repeat: no-repeat;
@@ -72,7 +74,7 @@
 
 .field-type {
   margin-left: 30px;
-  font-size: 12px
+  font-size: 12px;
 }
 
 .main-container-questions {
@@ -91,7 +93,7 @@
   padding-right: 10px;
   border: 0px thin;
   margin-top: 10px;
-  border: 1px lightgray;
+  border: 1px #e6e7ea;
   border-style: solid;
 }
 
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
@@ -23,28 +23,54 @@
 <form [formGroup]="layerForm" (ngSubmit)="onSave($event)">
   <mat-dialog-content>
     <div>Layer id: {{ layerId }}</div>
-    <input type="text" formControlName="question" />
-    <div>TODO: Layer UI goes here</div>
-  </mat-dialog-content>
-  <mat-dialog-actions>
-    <div>
-      <button
-        mat-raised-button
-        color="primary"
-        (click)="onClose()"
-        type="button"
-        class="btn-cancel"
+    <div formArrayName="questions" class="main-container-questions">
+      <div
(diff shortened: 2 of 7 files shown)


parulraheja98:
image

gino-m:
Thanks for the change @parulraheja98 ! @DaoyuT could you PTAL?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -65,6 +66,24 @@ export class LayerDialogComponent implements OnDestroy {
...
+  }
+
+  initQuestion() {
+    return this.formBuilder.group({
+      en: [''],
+    });
+  }
+
+  addQuestion() {
+    const control = this.layerForm.controls['questions'] as FormArray;
```
gino-m: Where does the field `controls` come from? We should be using the existing `Form` model object, based on this data model:
https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation

gino-m: Ah sorry, misunderstood that this is an Angular form, not a Ground form :) Please disregard.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -41,6 +41,7 @@ export class LayerDialogComponent implements OnDestroy {
   activeProject$: Observable<Project>;
   subscription: Subscription = new Subscription();
   layerForm: FormGroup;
+  selected = 'text';
```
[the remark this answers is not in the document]
parulraheja98: The purpose is that by mat-option with value text gets selected by default, I was trying to follow the angular docs on selecting mat-option default value, refs: [mat-select](https://material.angular.io/components/select/examples)

parulraheja98:
Screen Shot 2020-03-24 at 6 48 28 PM
@gino-m, @DaoyuT I just updated the styling

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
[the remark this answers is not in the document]
DaoyuT: Yes, `selectedFieldType` is more clear to me than `selected`.

parulraheja98:
@DaoyuT , Can you please look at my recent changes?

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -47,3 +47,54 @@
   margin-right: 0;
   font-weight: bold;
 }
+.link-add-question {
```
gino-m: Add blank line before

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -47,3 +47,54 @@
...
+  color: green;
```
gino-m: Please use color code from mocks

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -47,3 +47,54 @@
...
+  cursor: pointer;
+}
+
+.add-question {
+  padding: 20px;
+  border: 1px solid lightgray;
```
gino-m: Same here

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -47,3 +47,54 @@
...
+  border-radius: 10px;
+  margin-top: 20px;
+}
+
+.plus-icon {
+  display: inline-block;
+  width: 20px;
+  height: 20px;
+  vertical-align: top;
+  background: linear-gradient(green, green), linear-gradient(green, green);
```
gino-m: And here

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -47,3 +47,54 @@
...
+}
+
+.add-question {
+  padding: 20px;
+  border: 1px solid lightgray;
+  border-radius: 10px;
+  margin-top: 20px;
+}
+
+.plus-icon {
```
gino-m: Is there an Angular Material component we can use to render this styling instead?

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -47,3 +47,54 @@
...
+}
+
+.container-question {
+  border-radius: 10px;
+  padding-top: 20px;
+  padding-left: 10px;
+  padding-right: 10px;
+  border: 0px thin;
+  margin-top: 10px;
+  border: 1px lightgray;
```
gino-m: Use color code from mocks

web-ng/src/app/components/layer-dialog/layer-dialog.component.html
```diff
@@ -23,28 +23,55 @@
...
+          let question of layerForm.controls.questions.controls;
+          let i = index
+        "
+        class="container-question"
       >
-        Save
-      </button>
+        <div [formGroupName]="i" fxLayout>
+          <mat-form-field appearance="fill" fxFlex="2 1 auto">
+            <mat-label> Question</mat-label>
```
gino-m: Remove leading space

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -85,20 +104,10 @@ export class LayerDialogComponent implements OnDestroy {
     this.projectId = project.id;
   }
 
-  // TODO: Make getForm accomodate multiple fields
-  getForm(question: string, fieldId: string, formId: string): Form {
+  getForm(fields: Map<string, Field>, formId: string) {
```
gino-m: nit: I'd naturally think the formId would be the first arg

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -108,19 +117,25 @@ export class LayerDialogComponent implements OnDestroy {
...
+    let fields = Map<string, Field>();
+    this.layerForm.value.questions.map((question: { en: string }) => {
+      const fieldId = this.dataStoreService.generateId();
+      const field: Field = {
+        id: fieldId,
+        type: FieldType['TEXT'],
+        required: false,
+        label: StringMap(question),
+      };
+      fields = fields.set(fieldId, field);
```
gino-m: map != forEach; map is a functional operator whose lambda should return a value for each element.  Here it appears map() is being used as a proxy for a for loop. Please either convert to fully functional form, or replace with. forEach or for loop.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -108,19 +117,25 @@ export class LayerDialogComponent implements OnDestroy {
...
+    });
     const formId = this.dataStoreService.generateId();
-    const fieldId = this.dataStoreService.generateId();
     const layer = new Layer(
       this.layerId,
-      this.layer?.color,
+      this.layer?.color ? this.layer.color : '#ff9131',
```
gino-m: Please move default color into constant.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -108,19 +117,25 @@ export class LayerDialogComponent implements OnDestroy {
...
       this.layer?.name?.set(this.lang, this.layerName),
-      this.layerForm.value.question
+      this.layerForm.value.questions
```
gino-m: Is `questions` undefined when there are no items, or empty array?

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
parulraheja98: I tried finding mat-icon but seems like there is +1 but no isolated +

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -129,15 +129,16 @@ export class LayerDialogComponent implements OnDestroy {
       fields = fields.set(fieldId, field);
     });
     const formId = this.dataStoreService.generateId();
+    const layerColor = '#ff9131';
     const layer = new Layer(
       this.layerId,
-      this.layer?.color ? this.layer.color : '#ff9131',
+      this.layer?.color ? this.layer.color : layerColor,
```
gino-m: Please move into a class-level constant eg https://stackoverflow.com/a/37265420

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: Question s till stands - what happens if questions is empty?

parulraheja98: questions is not empty it is initialized with  { 'en': '' }

gino-m: I'm probably confused about how layerForm works - what is this.layerForm.value.questions? From its now I thought it was an array of components?

gino-m: Also, why does this ternary use fields but check `questions`? Also, can't `this.getForm` return the right thing directly instead of checking it here?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: You should be able to use `this.layer?.color || DEFAULT_LAYER_COLOR` instead of ternary here.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -24,11 +24,13 @@ import { Form } from '../../shared/models/form/form.model';
...
 import { DataStoreService } from '../../services/data-store/data-store.service';
 import { Router } from '@angular/router';
-import { FormBuilder, FormGroup } from '@angular/forms';
-import { FieldType } from '../../shared/models/form/field.model';
+import { FormBuilder, FormGroup, FormArray } from '@angular/forms';
+import { FieldType, Field } from '../../shared/models/form/field.model';
 import { StringMap } from '../../shared/models/string-map.model';
 import { Map } from 'immutable';
 
+const layerColor = '#ff9131';
```
gino-m: This name implies this represents the layer color (as if there were one, fixed color), whereas it actually represents the default one used when there is none set, so DEFAULT_LAYER_COLOR would be a more appropriate name imo. (Caps+underscore since it's a top-level constant.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -68,6 +71,24 @@ export class LayerDialogComponent implements OnDestroy {
...
   }
 
+  getFieldType() {
+    return {
+      icon: 'short_text',
+      label: 'Text',
+    };
+  }
+
+  initQuestion() {
```
gino-m: `init` implies it changes some internal state - here it just returns an empty group. Perhaps "createQuestionGroup" would be a better name?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: We spoke - `questions` is never null, but in theory can be empty. we should handle those cases here instead.

PR #255, opened 2020-03-27
Fixed the Layer named to be saved in the DB

parulraheja98:
Layer name was set using this.layer?.name?, this works when name is defined but on creating new layer, name is not defined so saving of new layer name fails

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -134,7 +134,7 @@ export class LayerDialogComponent implements OnDestroy {
     const layer = new Layer(
       this.layerId,
       this.layer?.color || DEFAULT_LAYER_COLOR,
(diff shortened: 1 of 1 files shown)


parulraheja98:
@DaoyuT , @gino-m

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -134,7 +134,7 @@ export class LayerDialogComponent implements OnDestroy {
     const layer = new Layer(
       this.layerId,
       this.layer?.color || DEFAULT_LAYER_COLOR,
-      this.layer?.name?.set(this.lang, this.layerName),
+      StringMap({ [this.lang]: this.layerName }),
```
gino-m: This will only work for projects with only one language. In the future we'll support multiple languages, so `layerName` itself should be a map. Care to add a TODO for that?

PR #259, opened 2020-04-16
[Add observation] Implement add observation form

gyarill:
i am still working on actual saving to DB but form is wired already
Screen Shot 2020-04-15 at 11 32 29 PM


gino-m:
@gyarill Can we close this draft?

gyarill:
Screen Shot 2020-06-16 at 6 16 56 PM
i updated css a bit

web-ng/src/app/components/observation-form/observation-form.component.spec.ts
```diff
@@ -0,0 +1,190 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
gino-m, line 1: Please add license header here and where missing.

web-ng/src/app/components/observation-form/observation-form.component.html
```diff
@@ -0,0 +1,60 @@
+<!--
+  Copyright 2019 Google LLC
```
gino-m, line 2: 2019->2020 here and in other new files.

web-ng/src/app/shared/models/observation/observation.model.ts
```diff
@@ -28,4 +28,15 @@ export class Observation {
     readonly lastModified: AuditInfo,
     readonly responses: Map<string, Response>
   ) {}
+
+  copy(responses: Map<string, Response>) {
```
gino-m, line 32: I'd expect a method called "copy" to only copy an object to another of the same class, whereas this is actually copying and modifying repsonses. I think the idiomatic way of calling this might be "withResponses()".

gino-m, line 32: Return type missing here.

web-ng/src/app/components/observation-form/observation-form.component.css
```diff
@@ -0,0 +1,50 @@
...
+.observation .field-label {
+    font-size: 12px;
+    line-height: 16px;
+    color: #70757A;
+    margin-bottom: 3px;
+}
+.observation .field-response {
+    font-size: 14px;
+    line-height: 20px;
+    color: #3C4043;
```
gino-m, line 29: Please use `mat-color()` instead.

gyarill: i deleted custom color. it will be default now

gino-m: Thank you!

PR #261, opened 2020-04-18
[Form editor] Allow user to mark/unmark fields "required"

### LINKED ISSUE #158: [Form editor] Allow user to mark/unmark fields "required"
image

parulraheja98:
Closes #158 
image


web-ng/src/app/components/layer-dialog/layer-dialog.component.html
```diff
@@ -96,6 +96,11 @@
               Add option
             </a>
           </div>
+          <div fxLayout fxLayoutAlign="end" class="question-actions">
+            <span class="required-label"> Required </span>
```
gino-m, line 100: Can we use CSS here to add padding instead of " "?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -175,7 +177,7 @@ export class LayerDialogComponent implements OnDestroy {
       let field: Field = {
         id: fieldId,
         type: FieldType['TEXT'],
-        required: false,
+        required: question.required || false,
```
gino-m, line 180: Can this ever actually be undefined or null?

web-ng/src/app/components/layer-dialog/layer-dialog.component.html
parulraheja98: I am adding margin-right here, forgot to remove extra spaces around the "Required"

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
parulraheja98: Yes this cannot be undefined as it has default value false

parulraheja98:
@gino-m, here is the updated UI after merge with master.
image

parulraheja98:
@gino-m , here is the updated UI after update.
image

PR #262, opened 2020-04-18
[Form editor] Show trash button used to delete fields 

### LINKED ISSUE #257: [Form editor] Show trash button used to delete fields
image

A confirmation dialog should show when the button is clicked:
"Are you sure you wish to delete this field? Any associated data will be lost. This cannot be undone."

In the future we can do something smarter than removing the field from the project metadata (e.g., mark "deleted: true", and purge deleted fields if no data exists for them).

parulraheja98:
Closes #257 

The placement of the bin will be changed in a future PR and aligned with required toggle.

image
image

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
@@ -96,6 +96,11 @@
               Add option
             </a>
           </div>
+          <div fxLayout fxLayoutAlign="end" class="question-actions">
+            <mat-icon (click)="deleteQuestion(j)" class="delete-icon">
+              delete_forever
+            </mat-icon>
+          </div>
         </div>
       </div>
       <div class="add-question">
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
@@ -97,9 +97,9 @@
             </a>
           </div>
           <div fxLayout fxLayoutAlign="end" class="question-actions">
-            <mat-icon (click)="deleteQuestion(j)" class="delete-icon">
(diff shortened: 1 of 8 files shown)


web-ng/src/app/components/layer-dialog/layer-dialog.component.html
```diff
@@ -96,6 +96,11 @@
               Add option
             </a>
           </div>
+          <div fxLayout fxLayoutAlign="end" class="question-actions">
+            <mat-icon (click)="deleteQuestion(j)" class="delete-icon">
```
gino-m, line 100: Please use `mat-icon-button` to get appropriate Material hover and click behavior:
https://material.angular.io/components/button/overview

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -138,6 +144,27 @@ export class LayerDialogComponent implements OnDestroy {
...
+        maxWidth: '500px',
+        data: {
+          title: 'Warning',
+          message:
+            'Are you sure you wish to delete this field? Any associated data will be lost. This cannot be undone.',
+        },
+      }
+    );
+
+    dialogRef.afterClosed().subscribe(dialogResult => {
```
gino-m, line 160: How does the subscription get disposed? Don't we need to keep it around in order to delete it when the UI is destroyed?

parulraheja98: No need to unsubscribe as it auto completes itself. Check this out [Refs](https://github.com/angular/components/blob/ae41a0ad69ca26c600f0f56c68dd5a1c102d4f1f/src/lib/dialog/dialog-ref.ts#L75)

PR #263, opened 2020-04-19
Add suffix to url in README.md

Tiyarah:
Add "/p/new" suffix to http://localhost:4200/ in README.md of web-ng folder to see the map UI, otherwise a blank page is shown.

### DIFF
--- a/web-ng/README.md
+++ b/web-ng/README.md
@@ -74,8 +74,10 @@ This last step will fail if you haven't yet following instructions above in
 
 ## Development server
 
-Run `ng serve` for a dev server. Navigate to http://localhost:4200/. The app
-will automatically reload if you change any of the source files.
+Run `ng serve` for a dev server. Navigate to http://localhost:4200/. Add 
+`p/<project_name>` as a suffix to the url to see the map interface in your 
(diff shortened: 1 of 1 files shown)


web-ng/README.md
```diff
@@ -74,8 +74,10 @@ This last step will fail if you haven't yet following instructions above in
 
 ## Development server
 
-Run `ng serve` for a dev server. Navigate to http://localhost:4200/. The app
-will automatically reload if you change any of the source files.
+Run `ng serve` for a dev server. Navigate to http://localhost:4200/. Add 
+`p/<project_name>` as a suffix to the url to see the map interface in your 
+browser (e.g. http://localhost:4200/p/new).
```
gino-m, line 79: Thanks for the edit! The special key for new projects is `:new` (with preceding colon).

Tiyarah: Done! Is it also p/:_<existing_project>_ or p/_<existing_project>_ is correct for existing projects?

gino-m: ":" is used only before "new" to indicate it's a special value. This was totally arbitrary - would be happy to replace it with /new or something else if it's helpful!

PR #264, opened 2020-04-19
Add authentication methods to CONTRIBUTING.md

Tiyarah:
Add authentication methods to CONTRIBUTING.md, since two-step verification is enabled by default for us, it is not trivial to use PAT instead of GitHub password. Another possibility would also be to use SSH.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,11 +99,11 @@ submit them to the base repository using a pull request.
     
     Where `<branch>` is the branch name you used in step 2.
 
-    If asked for github username and password, make sure to use your [Personal
-    Access Token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) instead of your password, since two-factor authentication is
-    enabled by default.
+    Note (Googlers-only): If asked for github username and password, make sure
+    to use your [Personal Access Token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) instead of your password, since two-factor 
+    authentication is enabled by default for accounts associated with Google.
 
-    Alternatively, one can [connect to GitHub via SSH](https://help.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh) instead of HTTPS and use
+    Alternatively, you can [connect to GitHub via SSH](https://help.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh) instead of HTTPS and use
     a public/private key pair for authentication.
 
 1. Create a [pull request](https://help.github.com/articles/about-pull-requests/) to have your changes reviewed and merged into the base 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,10 @@ submit them to the base repository using a pull request.
     
     Where `<branch>` is the branch name you used in step 2.
 
+    If asked for github username and password, make sure to use your [Personal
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -99,6 +99,13 @@ submit them to the base repository using a pull request.
     
     Where `<branch>` is the branch name you used in step 2.
 
+    If asked for github username and password, make sure to use your [Personal
```
scolsen, line 102: I forget--is 2FA a requirement for all contributors to google repos? Or is it just a requirement for googlers?

Either way looks good, just asking since if it only applies to googlers we may want to make this a note or something instead of weaving it into the main text.

CONTRIBUTING.md
```diff
@@ -99,6 +99,13 @@ submit them to the base repository using a pull request.
...
+    Access Token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) instead of your password, since two-factor authentication is
+    enabled by default.
+
+    Alternatively, one can [connect to GitHub via SSH](https://help.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh) instead of HTTPS and use
```
scolsen, line 106: nit: s/one/you

CONTRIBUTING.md
Tiyarah: I believe you are right! In my security setting it says 2FA is required because your account is affiliated with an org or corporation that requires it.

CONTRIBUTING.md
Tiyarah: Done!

CONTRIBUTING.md
```diff
@@ -99,11 +99,11 @@ submit them to the base repository using a pull request.
     
     Where `<branch>` is the branch name you used in step 2.
 
-    If asked for github username and password, make sure to use your [Personal
-    Access Token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) instead of your password, since two-factor authentication is
-    enabled by default.
+    Note (Googlers-only): If asked for github username and password, make sure
```
scolsen, line 102: Sorry to be a nag about this, but instead of (Googlers-only) I'd do something like:

```
**Note**: If you have two factor authentication enabled, make sure to use your [Personal Access Token](link) instead of your password. 
```

That way we also account for the case where someone has 2fa enabled but happens to belong to a different org or just has it set up for personal use.

CONTRIBUTING.md
```diff
@@ -99,6 +99,13 @@ submit them to the base repository using a pull request.
     
     Where `<branch>` is the branch name you used in step 2.
 
+    Note (Googlers-only): If asked for github username and password, make sure
```
gino-m, line 102: Nit: Please capitalize as "GitHub".

web-ng/README.md
```diff
@@ -74,8 +74,10 @@ This last step will fail if you haven't yet following instructions above in
 
 ## Development server
 
-Run `ng serve` for a dev server. Navigate to http://localhost:4200/. The app
-will automatically reload if you change any of the source files.
+Run `ng serve` for a dev server. Navigate to http://localhost:4200/. Add 
+`p/<project_name>` as a suffix to the url to see the map interface in your 
+browser (e.g. http://localhost:4200/p/new).
```
gino-m, line 79: Should be `/p/:new` (with colon). May want to specify that that's the URL to create a new blank project.

CONTRIBUTING.md
Tiyarah: Done!

CONTRIBUTING.md
Tiyarah: Done!

PR #271, opened 2020-04-22
[Form editor] Drag to reorder form fields

### LINKED ISSUE #226: [Form editor] Drag to reorder form fields
element `index` is used to represent field order:
https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation#document-structure

image

image

parulraheja98:
Closes #226 
image


web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -228,12 +228,13 @@ export class DataStoreService {
 
   private static formToJS(form: Form): {} {
     const { fields, ...formDoc } = form;
+    let index = 0;
```
gino-m, line 231: Could you please extract the call to `reduce` into its own named method to help make it clearer that this is where indices are reassigned here based on order in the form?

General question - do we ever sort the fields by index when displaying the fields in the form editor?

parulraheja98:
@gino-m , here is the updated UI after drag icon change
image

web-ng/src/app/services/data-store/data-store.service.ts
parulraheja98: Good question, right now the forms are not being rendered in the UI and only new forms are being created. I am going to make my next PR for rendering existing forms and ability to create new form using tabular switch between forms. I will consider sorting in that PR.

gino-m: Ok, will let this through for now. Thanks for creating a bug to track.

gino-m:
@parulraheja98 could you kindly resolve conflict so we can merge? Sorry I didn't merge it yet, I was waiting for checks to pass and forgot to come back to it!

parulraheja98:
Resolved conflicts, @gino-m

web-ng/src/app/components/layer-dialog/layer-dialog.component.css
```diff
@@ -152,3 +151,11 @@
 .trash-icon {
   color: #707074;
 }
+
+.drag-icon {
+  display: block;
+  margin: 0px auto;
+  cursor: move;
+  padding: 5px;
+  color: #dadde1;
```
gino-m, line 160: Could we turn this into scss and use `mat-color($foreground, lighter)` instad?

I was trying to follow https://material.angular.io/guide/theming-your-components but was unsuccessful.. Any idea how to set that up?

gino-m:
💯, great work @parulraheja98 !

PR #273, opened 2020-04-23
Fix new fields to be appended in the existing fields rather than creating a new form

### LINKED ISSUE #272: New fields should be appended to the existing fields in the form of layer dialog
New fields right now are being added as a new form whereas they should be appended to the existing form if it exists or a new form with fields will be created.

parulraheja98:
Closes #272

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -244,17 +244,21 @@ export class LayerDialogComponent implements OnDestroy {
       );
     });
     const formId = this.dataStoreService.generateId();
+    let forms = this.layer?.forms;
+    forms = forms
+      ? forms.set(formId, new Form(formId, fields))
+      : this.layerForm.value.questions &&
+        this.layerForm.value.questions.length > 0
+      ? Map({
+          [formId]: new Form(formId, fields),
+        })
+      : Map<string, Form>();
     const layer = new Layer(
       this.layerId,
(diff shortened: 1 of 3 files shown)


web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -244,17 +244,21 @@ export class LayerDialogComponent implements OnDestroy {
       );
     });
     const formId = this.dataStoreService.generateId();
+    let forms = this.layer?.forms;
```
gino-m, line 247: It's not obvious what this is doing - could you refactor this to make it more self-explanatory? (e.g. using carefully named local variables and/or functions)? Thanks!

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -218,9 +218,9 @@ export class DataStoreService {
     return new Form(
       id,
       Map<string, Field>(
-        keys(data.elements).map((id: string) => [
+        keys(data.fields).map((id: string) => [
```
gino-m, line 221: Same here - perhaps extracting to a private helper will help clarity.

gino-m:
Hi @parulraheja98 Thanks for the change! General question - is there a way to simplify this implementation a bit? 

Also, Firebase is the persistence layer, which the UI shouldn't know anything about. Instead, it should operate only on model objects in the domain and UI layer. Could we revisit this design a bit to refine it further?

gino-m:
Hi @parulraheja98 can I close this one for now in favor of your recent refactors?

PR #276, opened 2020-05-08
[CI] Run unit tests in Cloud Build

gino-m:
@shobhitagarwal1612 FYI.


shobhitagarwal1612:
@gino-m Tests are failing due to missing `*-config` files

gino-m:
Looks like this won't work until I provide fake keys for testing or another workaround:
```
Step #3 - "Run tests for web-ng": 21 import { googleMapsConfig } from './.google-maps-config';
Step #3 - "Run tests for web-ng":                                     ~~~~~~~~~~~~~~~~~~~~~~~
Step #3 - "Run tests for web-ng": 
Step #3 - "Run tests for web-ng": src/environments/environment.ts:22:32 - error TS2307: Cannot find module './.firebase-config'.
Step #3 - "Run tests for web-ng": 
Step #3 - "Run tests for web-ng": 22 import { firebaseConfig } from './.firebase-config';
Step #3 - "Run tests for web-ng":                                   ~~~~~~~~~~~~~~~~~~~~
```

@shobhitagarwal1612 any suggestions on how to best approach this?

shobhitagarwal1612:
We can store the config in GCS and copy them in the workspace in GCB before everything else. But then we need to make extra sure who has access to the GCP.

gino-m:
> We can store the config files in GCS and copy them in the workspace in GCB before everything else. But then we need to be extra sure who has access to the GCP.
> 
> Or, we can store the keys as SECRETS in the GCB config and create the config files on the fly.

The keys shouldn't actually be needed for running unit tests; is there a way we can inject empty keys at runtime while running tests via the GCB config.

gino-m:
> We can store the config files in GCS and copy them in the workspace in GCB before everything else. But then we need to be extra sure who has access to the GCP.
> 
> Or, we can store the keys as SECRETS in the GCB config and create the config files on the fly.

We don't need real keys for the unit tests, so we should be able to just create the config files on the fly. Is there an example of how to do that?

shobhitagarwal1612:
> We don't need real keys for the unit tests, so we should be able to just create the config files on the fly. Is there an example of how to do that?

How about adding a dummy file to the project like `dummy-google-maps-config` an `dummy-firebase-config` and renaming/moving them to correct location using another step in GCB?

```
  # Install node packages for web
  - name: 'gcr.io/cloud-builders/npm'
    id: 'Prepare for testing'
    waitFor: ['-']
    entrypoint: 'bash'
    args:
     - '-c'
     - |
       mv dummy-google-maps-config web-ng/src/environments/google-maps-config
```

Would that work? I'm also wondering if it's possible to create another environment file which should be used when --CI is passed with build command. In that case, we need not add another step in GCB to copy the configs as they can be used in the ts file directly. Wdyt?

gino-m:
I have almost everything working using Angular environments, but it seems the node image is too old (v10), and Angular requires 12 or newer. Tried upgrading, but now it seems the browser isn't  set up for running tests in that Docker config. I may need to run through this codelab:
https://codelabs.developers.google.com/codelabs/cloud-cardboard-viewer

@shobhitagarwal1612 does that sound like the right path, or is there an easier way?

gino-m:
I built an image using:
https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/ng

And updated the cloudbuild.yaml to point to it, but it doesn't have a Chrome image either (`: No binary for Chrome browser on your platform.`)..

shobhitagarwal1612:
This seems related https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/349

gino-m:
That issue has already been resolved; the problem I'm coming up against is
we need an image with Nodejs 12 or higher and a Chrome. We can create our
own, but I was hoping to find one preconfigured.

gino-m:
I tried configuring PhantomJS but ran into several issues. Also read that in some cases tests in PhantomJS may behave differently than when run in a browsers. After some more research, discovered [Puppeteer](https://github.com/puppeteer/puppeteer), which will install and run headless Chrome for you. Unfortunately this failed with `/workspace/web-ng/node_modules/puppeteer/.local-chromium/linux-737027/chrome-linux/chrome: error while loading shared libraries: libX11-xcb.so.1: cannot open shared object file: No such file or directory` in https://console.cloud.google.com/cloud-build/builds/3b883b64-e418-49e0-9e4b-aade05d04e16?project=264658303667

gino-m:
Now failing with:
```
Step #3 - "Run tests for web-ng": 13 05 2020 23:00:28.399:ERROR [launcher]: Cannot start ChromeHeadless
Step #3 - "Run tests for web-ng": 	[0513/230028.355719:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
Step #3 - "Run tests for web-ng": 
```

Will need to set user and fine-grained permissions.

Note: Created a custom docker image, defined in `cloud-build/`. The image includes `node-pre-gyp`; that dep contains native code, and building from source each time was taking around 5 min.

gino-m:
Running with --no-sandbox solved the previous issue. Still having trouble pre-installing node-pre-gyp libraries, but tests are passing:

image

@gyarill Could you PTAL? I can try to follow up on speeding up the test time in a separate PR.

gino-m:
@shobhitagarwal1612 could you PTAL?

gino-m:
> Thanks @gino-m, this is amazing!
> 
> The last build took ~ 90 mins. Do you know which step is behind this? I can't access GCB right now due to some account constraints.

Hm.. the last few runs seemed to be pretty fast: 

image

Maybe the build status never got updated on the PR? GitHub was having issues with this not long ago.

shobhitagarwal1612:
Let's merge it!

PR #277, opened 2020-05-08
change MultipleChoice.options field from List to Map. 

gyarill:
This way we will be able to easier iterate via them in html template files (Map values keeps to be recalculated when iterating)


gino-m:
Awesome, thanks!

PR #281, opened 2020-05-08
[feature panel] load observation data

gyarill:
when url fragment contains o=<observation.id> then load observation data


web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -136,6 +136,19 @@ export class DataStoreService {
...
   }
 
+  loadObservation$(project: Project, observationId: string) {
+    return this.db
+      .collection(`projects/${project.id}/observations`)
+      .doc(observationId)
+      .get()
+      .pipe(
+        map(doc => {
+          const form = project.getForm(doc.data()!.formId);
```
gino-m, line 146: The observation data returned from Firestore will have the ids necessary to rebuild the model graph; perhaps we can move this step into the converter, and use the layerId and formId inside the observation data to get the layer from the project, and the form from the layer, instead of searching for it as we in `getForm()` below? Firebase UUIDs are globally unique (statistically speaking), so this implementation would work w/Firebase as a backend, but might break if observationId weren't globally unique. It would also be a bit easier to follow logically, since searching through layers in `getForm` implies we don't which layer and observation belongs to, which isn't the case.

gyarill: done!

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -27,7 +27,7 @@ export class Project {
     readonly layers: Map<string, Layer>
   ) {}
 
-  getLayerForm(layerId: string, formId: string): Form {
+  getForm(layerId: string, formId: string): Form {
```
gino-m: Can we use composition here (`getLayer(layerId)!.getForm(formId)!` instead of combining the two behaviors into one?

gino-m: Also, not sure we want to throw an error if not found directly in the model. In general the model accessors return `null` for unknown members, so perhaps should do the same here for consistency?

gino-m:
@gyarill Gentle nudge!

PR #284, opened 2020-05-17
Refactor text field into its own separate component

parulraheja98:


web-ng/src/app/components/text-field/text-field.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  label: string;
+  type: FieldType;
+}
+
+@Component({
+  selector: 'app-text-field',
+  templateUrl: './text-field.component.html',
+  styleUrls: ['./text-field.component.css'],
+})
+export class TextFieldComponent implements OnInit, OnChanges {
```
gino-m, line 23: Are we sure breaking these UI components down along the lines of field type is the best approach here? The label, type, and required toggle are the same for all fields, so it would seem fitting if there was one "FieldEditorComponent", which in turn contains other [potentially hidden] subcomponents (e.g., OptionList, OptionSelector, etc) depending on the field type. Wdyt?

web-ng/src/app/components/form-field-editor/form-field-editor.component.css
```diff
@@ -0,0 +1,44 @@
...
+    margin-right: 20px;
+  }
+  
+  .question-actions {
+    margin-bottom: 20px;
+    margin-right: 10px;
+  }
+  
+  .trash-icon {
+    color: #707074;
```
gino-m, line 17: Let's use Angular Material themes instead of specifying specific colors wherever possible. In this case, the mat-button attribute would cause the trash icon to use the default foreground color.

web-ng/src/app/components/form-field-editor/form-field-editor.component.css
```diff
@@ -0,0 +1,44 @@
...
+  .container-question {
+      border-radius: 10px;
+      padding-top: 20px;
+      padding-left: 10px;
+      padding-right: 10px;
+      border: 0px thin;
+      margin-top: 10px;
+      border: 1px #e6e7ea;
+      border-style: solid;
+
```
gino-m, line 43: Nit: Extra line.

web-ng/src/app/components/form-field-editor/form-field-editor.component.css
```diff
@@ -0,0 +1,44 @@
...
+  
+  .required-toggle {
+    padding-top: 5px;
+  }
+  
+  .required-label {
+    margin-right: 20px;
+  }
+  
+  .container-question {
```
gino-m, line 34: Did you mean `.question-container`?

web-ng/src/app/components/form-field-editor/form-field-editor.component.html
```diff
@@ -0,0 +1,32 @@
+<form [formGroup]="textFieldGroup" class="container-question">
```
gino-m, line 1: Please add license headers here and throughout.

web-ng/src/app/components/form-field-editor/form-field-editor.component.html
```diff
@@ -0,0 +1,32 @@
...
+      </button>
+      <span class="vertical-divider"></span>
+      <div class="required-toggle">
+        <span class="required-label">Required</span>
+        <mat-slide-toggle formControlName="required" color="primary">
+        </mat-slide-toggle>
+      </div>
+    </div>
+  </form>
+
```
gino-m, line 32: Nit: extra LF

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  templateUrl: './form-field-editor.component.html',
+  styleUrls: ['./form-field-editor.component.css'],
+})
+export class FormFieldEditorComponent implements OnInit {
+  @Input() label?: string;
+  @Input() required?: boolean;
+  @Input() type?: string;
+  @Output() update = new EventEmitter();
+  @Output() delete = new EventEmitter();
+  fieldTypeOptions: FieldTypeOptionModel[] = [
```
gino-m, line 29: I started writing this:
> `FieldTypeOptionModel` is a bit hard to grok at first glance. Can we call it something like `FieldType`, and then the property could be "availableFieldTypes"?

But then I realized, why do we need this at all? Since it's constant, can't it be defined directly in the template?

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -0,0 +1,84 @@
...
+      icon: 'library_add_check',
+      label: 'Select multiple',
+      type: FieldType.MULTIPLE_CHOICE,
+    },
+  ];
+
+  textFieldGroup: FormGroup;
+
+  constructor(private formBuilder: FormBuilder) {
+    this.textFieldGroup = this.formBuilder.group({
```
gino-m, line 45: What does textFieldGroup represent? Won't this component be generic, not specific to text fields?

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -0,0 +1,84 @@
...
+import { FormGroup, FormBuilder } from '@angular/forms';
+import { FieldType } from '../../shared/models/form/field.model';
+
+export interface FieldTypeOptionModel {
+  icon: string;
+  label: string;
+  type: FieldType;
+}
+
+@Component({
```
gino-m, line 18: Please add basic TSDoc to each class and non-obvious public method.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -219,16 +203,14 @@ export class LayerDialogComponent implements OnDestroy {
     );
   }
 
-  convertQuestionToField(fieldId: string, question: Question): Field {
-    switch (question.fieldTypeOption.type) {
+  convertQuestionToField(fieldId: string, question: Field): Field {
```
gino-m, line 206: This is a bit confusing now - it accepts a Field arg called `question`, and returns.. a field?

web-ng/src/app/shared/converters/firebase-data-converter.ts
```diff
@@ -109,9 +109,9 @@ export class FirebaseDataConverter {
     return new Form(
       id,
       Map<string, Field>(
-        keys(data.elements).map((id: string) => [
+        keys(data.fields).map((id: string) => [
```
gino-m, line 112: Was this changed by accident? In the current spec it's still called `elements`:
https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation#document-structure

parulraheja98: No its not a mistake, toForm method is being passed fields key instead of elements, nobody noticed that earlier

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
parulraheja98: forgot to rename this , thanks!

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
parulraheja98: this method is no longer needed as it is already field type, removed this method

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
parulraheja98: done!

web-ng/src/app/components/form-field-editor/form-field-editor.component.html
parulraheja98: done!

web-ng/src/app/components/form-field-editor/form-field-editor.component.css
parulraheja98: fixed it to question-container

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
parulraheja98: I am using it for mat-select-trigger which requires the element to be index of array type

web-ng/src/app/shared/converters/firebase-data-converter.ts
gino-m: In our current data model "fields" are a special type of form element. In the schema there currently isn't a property with the name `fields`, only `elements`. If we're accidentally saving it to Firebase with the name "fields" elsewhere we should update it to "elements" so we all use the same database doc structure.

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
gino-m: An array is fine here, I was referring to the type and variable names. Option and Model are overloaded terms, and using them here arguably makes things less clear.

web-ng/src/app/components/form-field-editor/form-field-editor.component.html
gino-m: Thanks!

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -0,0 +1,117 @@
...
+  templateUrl: './form-field-editor.component.html',
+  styleUrls: ['./form-field-editor.component.css'],
+})
+export class FormFieldEditorComponent implements OnInit, OnChanges {
+  @Input() label?: string;
+  @Input() required?: boolean;
+  @Input() type?: string;
+  @Output() update = new EventEmitter();
+  @Output() delete = new EventEmitter();
+  fieldTypes: FieldTypeModel[] = [
```
gino-m, line 46: If we won't move these into the template, let's mark the field as readonly to indicate it will never be replaced. 

Also, as mentioned "Model" may be confusing here; `FieldType` may be clearer.

gino-m:
Tests seems to be failing with:
```

ERROR in src/app/shared/converters/firebase-data-converter.spec.ts:58:30 - error TS2554: Expected 5-6 arguments, but got 4.

 58   static element001: Field = new Field(
                                 ~~~~~~~~~~
 59     'element001',
    ~~~~~~~~~~~~~~~~~
... 
 62     /*required=*/ true
    ~~~~~~~~~~~~~~~~~~~~~~
 63   );
    ~~~

  src/app/shared/models/form/field.model.ts:51:5
    51     readonly index: number,
           ~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'index' was not provided.
src/app/shared/converters/firebase-data-converter.spec.ts:70:5 - error TS2345: Argument of type 'MultipleChoice' is not assignable to parameter of type 'number'.

 70     new MultipleChoice(
        ~~~~~~~~~~~~~~~~~~~
 71       Cardinality.SELECT_MULTIPLE,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 75       ])
    ~~~~~~~~
 76     )
    ~~~~~
```

PR #216, opened 2020-02-14
UI styling of header component

parulraheja98:
I just was not sure about where Sign out button will go , I included it on far right of the screen. Rest mocks uploaded were very descriptive 

Refs: #214

PR #285, opened 2020-05-18
[Project header] Clean up header and convert to Material Design

gino-m:
Closes #216.


web-ng/src/app/components/project-header/project-header.component.html
```diff
@@ -13,20 +13,23 @@
...
+<div fxLayout="row" class="header" fxLayoutAlign="flex-self center">
+  <div>
     <img class="logo" src="/assets/img/logo.png" />
-    <span class="ground app-title">Ground</span>
+    <span class="app-title">Ground</span>
     <span class="header-divider"></span>
     <app-inline-edit-title></app-inline-edit-title>
   </div>
-  <div class="header-item">
+  <div fxFlex></div>
```
parulraheja98, line 23: is there any reason why this div is empty?

gino-m: I was using the empty div to stretch the header to full width, but seems this isn't necessary. Moved `fxFlex` to the first `div` instead.

gino-m:
> Very cool! In future we can have global classes defined for margin and padding such as margin-md, padding-md in styles.scss. This will prevent creating new classes just for margin or padding purposes.

Good idea, provided we only keep the truly common ones there. I can imagine there will be a lot of cases where we tweak the padding for specific items, but a general padding/margin style sounds like a good idea.

PR #287, opened 2020-05-18
Integrate Angular Material themes into several components

gino-m:
There are a few more instances of hard-coded color codes left. This is intended as an example of how we can use the custom theme throughout out styles. @DaoyuT @gyarill FYI.

### DIFF
--- a/web-ng/src/app/components/inline-edit-title/inline-edit-title.component.scss
+++ b/web-ng/src/app/components/inline-edit-title/inline-edit-title.component.scss
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
- /* Relative top used to push title input field in line with flex row. */
+ 
+ @import '~src/custom-theme';
+ 
+/* Relative top used to push title input field in line with flex row. */
 .inline-edit {
   position: relative;
   top: 2px;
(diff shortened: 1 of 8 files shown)


web-ng/src/app/components/inline-edit-title/inline-edit-title.component.scss
```diff
@@ -31,19 +33,19 @@ input {
   border: none;
   font-size: 14pt;
   margin: 10px 10px 5px 0;
-  border: 1px solid transparent;
-  color: #afa7a7
+  border: 2px solid transparent;
```
parulraheja98, line 36: why is that we are using mat-color for some styling and not for others?

web-ng/src/custom-theme.scss
```diff
@@ -27,12 +27,17 @@
...
+$accent: mat-palette($mat-amber, 700, 600, 800);
+$warn: mat-palette($mat-pink, 600, 500, 700);
 
 // Create the theme object (a Sass map containing all of the palettes).
-$web-ng-theme: mat-light-theme($web-ng-primary, $web-ng-accent, $web-ng-warn);
+$theme: mat-light-theme($primary, $accent, $warn);
+
+// Get generated palletes from theme. Sub-palettes listed in:
+// https://github.com/angular/components/blob/16f5f793dea3d83a6f08de14e3787ce97d82ff15/src/material/core/theming/_palette.scss#L674
+$foreground: map-get($theme, foreground);
```
parulraheja98, line 39: is foreground passed in as parameter something inbuilt we can use?

gino-m: `foreground` arg is the name of the palette we want to extract.

web-ng/src/app/components/inline-edit-title/inline-edit-title.component.scss
gino-m: In cases where there is no color code like here it's because we're allowing the default global style to work for us. In cases where the hex code is still used it's just because I haven't gotten to converting them yet :)

parulraheja98: sounds good :)

PR #256, opened 2020-04-08
Add Multiple Choice Fields for the form designer

parulraheja98:
Refs: #225 
image

PR #292, opened 2020-05-24
[Form editor] Add component for editing multiple choice options

parulraheja98:
Also bumped the version of puppeteer as the local test were not running in the old version.


web-ng/src/app/components/form-field-editor/form-field-editor.component.spec.ts
```diff
@@ -33,6 +37,9 @@ describe('FormFieldEditorComponent', () => {
     TestBed.configureTestingModule({
       declarations: [FormFieldEditorComponent],
       imports: [
+        AngularFireModule.initializeApp(environment.firebaseConfig),
```
gino-m, line 40: Why initialize Firebase here? For unit tests we wouldn't want to use a real database, but rather a mocked out instance of DataStore. Can that be done here?

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -112,6 +126,55 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
...
+
+  /**
+   * Emits event to layer dialog whenvever label or code gets updated in options-editor.
+   *
+   * @param event: label and code value of the option field.
+   * @param index: index of the option to be updated.
+   * @returns void
+   *
+   */
+
```
gino-m, line 146: Extra LF

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -112,6 +126,55 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
...
+  onAddOption() {
+    const optionId = this.dataStoreService.generateId();
+    const option = new Option(optionId || '', '', StringMap({ en: '' }));
+    const index = this.formOptions?.options.size || 0;
+    this.setOption(index, option);
+  }
+
+  setOption(index: number, option: Option) {
+    const cardinality = this.formOptions?.cardinality || 2;
```
gino-m, line 169: Please use enum here instead.

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -112,6 +126,55 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
...
+    let options = this.formOptions?.options || List<Option>();
+    options = options?.set(index, option);
+    this.multipleChoice = new MultipleChoice(cardinality, options);
+    this.update.emit({
```
gino-m, line 173: It looks like the form editor and option component keep their state in sync on each change. Is this necessary? Can we better encapsulate the behavior of the options component so that the parent component can query its state as needed?

web-ng/src/app/components/options-editor/options-editor.component.html
```diff
@@ -0,0 +1,33 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<form [formGroup]="optionsGroup">
+  <div fxLayout fxLayoutGap="15px">
+    <span class="dot"></span>
```
gino-m, line 19: What is this dot for? Having trouble visualizing what relates to in the specs.

web-ng/src/app/components/options-editor/options-editor.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  OnChanges,
+} from '@angular/core';
+import { FormGroup, FormBuilder } from '@angular/forms';
+
+@Component({
+  selector: 'app-options-editor',
+  templateUrl: './options-editor.component.html',
+  styleUrls: ['./options-editor.component.css'],
+})
+export class OptionsEditorComponent implements OnInit, OnChanges {
```
gino-m, line 32: If each instance represents a single option this should be called "OptionEditComponent" (singular) or similar.

web-ng/src/app/components/options-editor/options-editor.component.html
parulraheja98: image
This dot is css styling for options

web-ng/src/app/components/form-field-editor/form-field-editor.component.html
```diff
@@ -37,6 +37,25 @@
...
+      "
+      fxLayout
+    >
+      <app-options-editor
+        [label]="option.label && option.label.get('en')"
+        [code]="option.code"
+        (update)="onOptionUpdate($event, i)"
+      ></app-options-editor>
+    </div>
+    <a class="add-option" (click)="onAddOption()">
```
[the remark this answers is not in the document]
parulraheja98: I thought it was intentional to show it as a link https://github.com/google/ground-platform/pull/256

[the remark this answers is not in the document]
gino-m: It's actually a Material button with type "basic". This guarantees consistent touch area, haptics, animations. See link for examples / usage.

web-ng/src/app/components/options-editor/options-editor.component.html
gino-m: The "option" you're referring to is a possible value in a "select" UI element. These don't make sense here since this is the form designer, not the actual final form where options would be shown. We would normally show these if we wanted the user to select one of the options from a list. Here the user isn't selecting anything, they're defining the options to be shown on the form.

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
gino-m: I actually may be wrong here - how to passive components in Angular normally expose their state? Do parents subscribe to  `output`, or do the invoke a method to query state?

parulraheja98: It works in both way you can use value change to send value to the parent and on the other way you can query child state too. Both ways are valid.

web-ng/src/app/components/options-editor/options-editor.component.html
gino-m: Deleted my previous comment, apologies for the confusion. Could you kindly indent it further to leave room for the eventual drag widget, and adjust the color to grey (e.g. `mat-color($foreground, light)`

web-ng/.prettierrc.js
```diff
@@ -1,5 +1,4 @@
 module.exports = {
-  ...require('gts/.prettierrc.json'),
```
parulraheja98, line 2: This gives me error locally that cannot find gts/.prettierrc.json. Also why are we importing and adding properties both?

PR #306, opened 2020-05-28
Fix reload issue on projects

### LINKED ISSUE #305: Layer dialog doesn't load fields on the first attempt sometimes despite project being loaded

parulraheja98:
Closes #305

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -112,7 +112,7 @@ export class LayerDialogComponent implements OnDestroy {
     });
   }
 
-  init(projectId: string, createLayer: boolean, layer: Layer | undefined) {
+  init(projectId: string, createLayer: boolean, layer?: Layer) {
     this.projectId = projectId;
     if (!createLayer && !layer) {
       throw Error('Layer not found');
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -15,8 +15,6 @@
  */
 
 import { Component, Inject, OnDestroy } from '@angular/core';
-import { Observable } from 'rxjs';
-import { ProjectService } from '../../services/project/project.service';
 import {
   MatDialogRef,
   MAT_DIALOG_DATA,
@@ -50,15 +48,13 @@ export class LayerDialogComponent implements OnDestroy {
   layer?: Layer;
   layerName!: string;
   projectId?: string;
-  activeProject$: Observable<Project>;
   subscription: Subscription = new Subscription();
   fieldTypes = FieldType;
   fields: List<Field>;
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) data: { layerId: string },
+    @Inject(MAT_DIALOG_DATA) data: { layerId: string; project: Project },
     private dialogRef: MatDialogRef<LayerDialogComponent>,
-    private projectService: ProjectService,
     private dataStoreService: DataStoreService,
     private router: Router,
     private confirmationDialog: MatDialog
@@ -66,27 +62,9 @@ export class LayerDialogComponent implements OnDestroy {
     this.lang = 'en';
     // Disable closing on clicks outside of dialog.
     dialogRef.disableClose = true;
+    this.fields = List<Field>();
     this.layerId = data.layerId;
-    this.activeProject$ = this.projectService.getActiveProject$();
-    this.subscription.add(
-      this.activeProject$.subscribe(project => {
-        this.onProjectLoaded(project);
-      })
-    );
-    const fields = List<Field>();
-    const fieldId = this.dataStoreService.generateId();
-    this.fields = fields.push(
-      new Field(
-        fieldId,
-        1,
-        StringMap({
-          en: '',
-        }),
-        false,
-        0,
-        undefined
-      )
-    );
+    this.onProjectLoaded(data.project);
   }
 
   addQuestion() {
@@ -135,9 +113,22 @@ export class LayerDialogComponent implements OnDestroy {
   onProjectLoaded(project: Project) {
     if (this.layerId === ':new') {
       this.layerId = this.dataStoreService.generateId();
+      const fieldId = this.dataStoreService.generateId();
+      this.fields = this.fields.push(
+        new Field(
+          fieldId,
+          1,
+          StringMap({
+            en: '',
+          }),
+          false,
+          0,
+          undefined
+        )
+      );
       this.layer = new Layer(this.layerId, /* index */ -1);
     } else {
-      this.layer = project.layers.get(this.layerId);
+      this.layer = project?.layers?.get(this.layerId);
     }
     this.layerName = this.layer?.name?.get(this.lang) || '';
     const form = this.getForms();
@@ -147,10 +138,7 @@ export class LayerDialogComponent implements OnDestroy {
           ?.fields.toList()
           .sortBy(field => field.index) || List<Field>();
     }
-    if (!this.layer) {
-      throw Error('No layer exists');
-    }
-    this.projectId = project.id;
+    this.projectId = project?.id;
   }
 
   private getForms(): Form | undefined {
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -72,7 +72,7 @@ export class LayerDialogComponent implements OnDestroy {
     this.fields = this.fields.push(
       new Field(
         fieldId,
-        1,
+        FieldType.TEXT,
         StringMap({
           en: '',
         }),
@@ -117,7 +117,7 @@ export class LayerDialogComponent implements OnDestroy {
       this.fields = this.fields.push(
         new Field(
           fieldId,
-          1,
+          FieldType.TEXT,
           StringMap({
             en: '',
           }),
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -53,7 +53,8 @@ export class LayerDialogComponent implements OnDestroy {
   fields: List<Field>;
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) data: { layerId: string; projectId: string; layers: Map<string,Layer> },
+    @Inject(MAT_DIALOG_DATA)
+    data: { layerId: string; projectId: string; layers: Map<string, Layer> },
     private dialogRef: MatDialogRef<LayerDialogComponent>,
     private dataStoreService: DataStoreService,
     private router: Router,
@@ -76,6 +77,7 @@ export class LayerDialogComponent implements OnDestroy {
         StringMap({
           en: '',
         }),
+        /* required= */
         false,
         this.fields.size,
         undefined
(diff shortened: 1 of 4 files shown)


web-ng/src/app/services/project/project.service.ts
```diff
@@ -36,12 +36,12 @@ export class ProjectService {
         this.activeProjectId$.pipe(
           // Asynchronously load project. switchMap() internally disposes
           // of previous subscription if present.
-          switchMap(id => this.dataStore.loadProject$(id))
+          switchMap(id => this.dataStore.loadProject$(id)),
+          shareReplay(1)
```
gino-m, line 40: Have you verified that `loadProject` is only getting called once? (eg add `doOnNext(p -> console.log(p))` instead loadProject$, for ex)? My guess is that it would called multiple times with `shareReplay` being placed here, but I could be wrong.

web-ng/src/app/services/project/project.service.ts
```diff
@@ -36,12 +36,12 @@ export class ProjectService {
...
+          // Cache last loaded project so that late subscribers don't cause
```
gino-m, line 41: If this sol'n work, please move this comment above the call to `shareReplay`.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -50,43 +48,23 @@ export class LayerDialogComponent implements OnDestroy {
...
   layerName!: string;
   projectId?: string;
-  activeProject$: Observable<Project>;
   subscription: Subscription = new Subscription();
   fieldTypes = FieldType;
   fields: List<Field>;
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) data: { layerId: string },
+    @Inject(MAT_DIALOG_DATA) data: { layerId: string; project: Project },
```
gino-m, line 56: To minimize the surface area of the API of this component, can we pass in only what's needed, i.e. projectId and Layer?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -50,43 +48,23 @@ export class LayerDialogComponent implements OnDestroy {
...
-        1,
-        StringMap({
-          en: '',
-        }),
-        false,
-        0,
-        undefined
-      )
-    );
+    this.onProjectLoaded(data.project);
```
gino-m, line 67: This is a big of a misnomer within the context of this component now. Maybe something like `init()` would be more appropriate?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -135,9 +113,22 @@ export class LayerDialogComponent implements OnDestroy {
   onProjectLoaded(project: Project) {
     if (this.layerId === ':new') {
       this.layerId = this.dataStoreService.generateId();
+      const fieldId = this.dataStoreService.generateId();
```
gino-m, line 116: If we update the component to accept `Layer?` instead of Project and Id, a null `Layer` could mean "new" here. This also makes sense since `:new` is a detail of the URL/router that the component shouldn't know/care about.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -135,9 +113,22 @@ export class LayerDialogComponent implements OnDestroy {
...
+      this.fields = this.fields.push(
+        new Field(
+          fieldId,
+          1,
```
gino-m, line 120: Please use enums for field types.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -135,9 +113,22 @@ export class LayerDialogComponent implements OnDestroy {
...
+          StringMap({
+            en: '',
+          }),
+          false,
+          0,
```
gino-m, line 125: Also, in Google style, constant args should be prefixed with /* argname= */.
https://google.github.io/styleguide/jsguide.html#formatting-param-name-comments

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -87,8 +88,12 @@ export class MainPageComponent implements OnInit {
   }
 
   private showEditLayerDialog(layerId: string) {
-    this.dialog.open(LayerDialogComponent, {
-      data: { layerId },
-    });
+    this.subscription.add(
+      this.activeProject$.pipe(take(1)).subscribe(project => {
```
gino-m, line 92: I did a [little research](https://medium.com/@benlesh/rxjs-dont-unsubscribe-6753ed4fda87), and you can do this to have the subscription automatically disposed of for you after one item is emitted:
```
this.activeProject$.take(1).subscribe(project => this.dialog.open...);
```

`this.subscription.add` can the be removed. Note that `{}` is not needed in the lambda in this case.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -50,54 +47,46 @@ export class LayerDialogComponent implements OnDestroy {
...
   fieldTypes = FieldType;
   fields: List<Field>;
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) data: { layerId: string },
+    @Inject(MAT_DIALOG_DATA)
+    data: {
+      layerId: string;
+      projectId: string;
+      layers: Map<string, Layer>;
```
gino-m, line 59: Why does the layer dialog need access to all the layers? Can we just pass the layer that's being edited (optional)? See https://en.wikipedia.org/wiki/Principle_of_least_privilege

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -50,54 +47,46 @@ export class LayerDialogComponent implements OnDestroy {
...
-        1,
-        StringMap({
-          en: '',
-        }),
-        false,
-        0,
-        undefined
-      )
-    );
+    if (data.layers) {
```
gino-m, line 72: Why not just `this.initProject(data.projectId, data.layer, data.createLayer)`?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -50,54 +47,46 @@ export class LayerDialogComponent implements OnDestroy {
...
     const fieldId = this.dataStoreService.generateId();
     this.fields = this.fields.push(
       new Field(
         fieldId,
-        1,
+        FieldType.TEXT,
         StringMap({
           en: '',
         }),
+        /* required= */
```
gino-m, line 89: Thanks! What is the last arg in this call? If it's optional just leave it out instead of specifying `undefined`.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -132,12 +121,36 @@ export class LayerDialogComponent implements OnDestroy {
     });
   }
 
-  onProjectLoaded(project: Project) {
-    if (this.layerId === ':new') {
+  initProject(
```
gino-m, line 124: This method doesn't actually initialize the project, it initializes the LayerDialogComponent, so `init()` or similar would be more appropriate here. But really, this is just part of the constructor logic logic and can be inlined there if you like.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -50,54 +47,46 @@ export class LayerDialogComponent implements OnDestroy {
...
     private dialogRef: MatDialogRef<LayerDialogComponent>,
-    private projectService: ProjectService,
     private dataStoreService: DataStoreService,
     private router: Router,
     private confirmationDialog: MatDialog
   ) {
     this.lang = 'en';
     // Disable closing on clicks outside of dialog.
     dialogRef.disableClose = true;
+    this.fields = List<Field>();
```
gino-m, line 70: It looks like initialization of fields overlaps with `init()`. Inlining the init method here will make that clearer (see comment below).

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -132,12 +121,36 @@ export class LayerDialogComponent implements OnDestroy {
...
+          false,
+          /* index= */
+          0,
+          undefined
+        )
+      );
       this.layer = new Layer(this.layerId, /* index */ -1);
     } else {
-      this.layer = project.layers.get(this.layerId);
+      this.layer = layers?.get(this.layerId);
```
gino-m, line 149: Based on previous comments, this logic would go away in favor of throwing an error if the component is called with both `createNew` and `layer`, or with `!createNew` and `!layer`.

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -69,6 +70,11 @@ export class MainPageComponent implements OnInit {
       return;
     }
     const params = new HttpParams({ fromString: fragment });
+    // The 'l' param with value ':new" is used to create new layer.
```
gino-m, line 73: If this logic is handled inside showEditLayerDialog(), this block can be deleted.

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -86,9 +92,24 @@ export class MainPageComponent implements OnInit {
     }
   }
 
-  private showEditLayerDialog(layerId: string) {
-    this.dialog.open(LayerDialogComponent, {
-      data: { layerId },
+  private showEditLayerDialog(layerId?: string) {
```
gino-m, line 95: ..in that case layerId would be required (it can be a value or ":new").

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -86,9 +92,24 @@ export class MainPageComponent implements OnInit {
...
+    this.activeProject$.pipe(take(1)).subscribe(project => {
+      let layerDialogData = {};
+      if (layerId) {
```
gino-m, line 98: No conditional needed here; you can just set `createNew: layerId == ":new"` in the map.  The local vars could then be removed and the logic inlined below in the call to `dialog.open()`.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -46,61 +43,44 @@ const DEFAULT_LAYER_COLOR = '#ff9131';
...
   subscription: Subscription = new Subscription();
   fieldTypes = FieldType;
   fields: List<Field>;
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) data: { layerId: string },
+    @Inject(MAT_DIALOG_DATA)
+    data: {
+      projectId: string;
+      layer: Layer;
```
gino-m, line 57: Since this can be null, should use `layer?` here?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -132,12 +112,30 @@ export class LayerDialogComponent implements OnDestroy {
...
   }
 
-  onProjectLoaded(project: Project) {
-    if (this.layerId === ':new') {
-      this.layerId = this.dataStoreService.generateId();
-      this.layer = new Layer(this.layerId, /* index */ -1);
+  init(projectId: string, createLayer: boolean, layer: Layer) {
+    this.projectId = projectId;
+    if (!createLayer && !layer) {
+      throw Error('No layer exists.');
```
gino-m, line 118: Nit: => 'Layer not found'

PR #333, opened 2020-06-20
[Layer dialog] Create new layer when unknown layerId passed in URL

parulraheja98:
Allow users to pass any layer id when trying to create new layer. New layer should be created for id that doesn't exist for any layer.

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -72,19 +72,8 @@ export class LayerDialogComponent implements OnDestroy {
   }
 
   addQuestion() {
-    const fieldId = this.dataStoreService.generateId();
-    this.fields = this.fields.push(
-      new Field(
-        fieldId,
-        FieldType.TEXT,
-        StringMap({
-          en: '',
-        }),
-        /* required= */
-        false,
-        this.fields.size
-      )
-    );
+    const newField = this.createNewField();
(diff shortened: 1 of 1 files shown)


web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -116,10 +116,7 @@ export class LayerDialogComponent implements OnDestroy {
 
   init(projectId: string, createLayer: boolean, layer?: Layer) {
     this.projectId = projectId;
-    if (!createLayer && !layer) {
-      throw Error('Layer not found');
-    }
-    if (createLayer) {
+    if ((!createLayer && !layer) || createLayer) {
```
gino-m, line 119: To avoid nesting the entire body of the function of the body in a block, you can use short circuiting (i.e. return instead).

Also, is there a way to make this expression more self explanatory? Are any of these possible states actually errors that we'd like to log?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -114,31 +114,31 @@ export class LayerDialogComponent implements OnDestroy {
     });
   }
 
+  createNewLayer() {
```
gino-m, line 117: Readers [like myself] would assume a `create` methods would implement the factory method design pattern; that they'd be static factories which return a new instance and do not modify state, so this method should update internal state or return a new instance, but not both.

PR #338, opened 2020-06-21
Fix enter key to focus on mat-select

### LINKED ISSUE #334: [Layer dialog] When editing Question field, the Return key triggers "Delete" action
**Describe the bug**
In the "Create New Layer" dialog, when editing Question field, the Return key triggers "Delete" action

**To Reproduce**
1. Go to 'Create New Layer'
2. Type anything in the Question field text
3. Hit Return key
4. See Warning dialog for "Delete Question" action

**Expected behavior**
Return key moves the cursor from "Question" to the text field, i.e. Question type selector

**Screenshots**
image

parulraheja98:
Closes #334


parulraheja98:
@gino-m I Changed the type="button" this will prevent submit action and I removed the navigation to select on enter press.

PR #339, opened 2020-06-21
Styling Changes for the Layer-Dialog

parulraheja98:
- Update hover and focus actions of layer dialog title.
- Fix margin bottom of mat-dialog-actions
- Update add-question container to be a button instead of div.
- Move inline-edit-title and inline-editor styling to common scss file

image


parulraheja98:
image

Issue #349, opened 2020-06-28
[DI] Define best practice and/or simplify structure of Angular modules


gino-m:
There's currently one module per component, leading to a lot of extra boilerplate. Is this the recommended approach?

gino-m:
@parulraheja98 @DaoyuT @gyarill in case you have any insight!

gino-m:
One module per component seems to be working for us; let's close this for now and revisit if it comes up again.

PR #350, opened 2020-06-28
[Route] Refactor route state into custom service

gino-m:
Work towards #351.


web-ng/src/app/services/router/router.service.ts
```diff
@@ -0,0 +1,78 @@
...
+import { map } from 'rxjs/operators';
+import { ActivatedRoute } from '@angular/router';
+import { HttpParams } from '@angular/common/http';
+
+/**
+ * Exposes application state in the URL as streams to other services
+ * and components, and provides methods for altering said state.
+ */
+@Injectable({
+  providedIn: 'root',
```
parulraheja98, line 28: Can this be injected to MainPageModule?

gino-m: I think this is probably a good idea, but at present all of our services are provided in root instead of as service modules. According to https://angular.io/guide/module-types, they should only be imported into App Module. Let's follow up on this in #349?

PR #353, opened 2020-06-28
[Router] Refactor URL fragment manipulation into custom RouterService

### LINKED ISSUE #351: [Router] Refactor interactions with URL into custom service
- [x] Always read URL/route state via service
- [x] Always update URL/route state via service

gino-m:
Fixes #351.

### DIFF
--- a/web-ng/src/app/services/router/router.service.ts
+++ b/web-ng/src/app/services/router/router.service.ts
@@ -17,8 +17,12 @@
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router, NavigationExtras } from '@angular/router';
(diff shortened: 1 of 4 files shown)


web-ng/src/app/services/router/router.service.ts
```diff
@@ -75,4 +84,50 @@ export class RouterService {
   getObservationId$(): Observable<string | null> {
     return this.observationId$!;
   }
+
+  /**
+   * Returns the current URL fragment, parsed as if their were normal HTTP
+   * query parameter key/value pairs.
+   */
+  private getFragmentParams(): HttpParams {
+    const fragment = this.activatedRoute!.snapshot.fragment;
```
parulraheja98, line 93: ! -> ?  for undefined type

gino-m: `!` is correct - we want to throw an error if `getFragmentParams()` was called before `init()` (i.e. activatedRoute was not yet set).

Issue #348, opened 2020-06-28
[Layer list] Finalize styling and UI elements to match spec


gino-m:
This includes:
- [x] Add "Layers" heading
- [x] Eliminate lines between 
- [x] Fix padding/spacing of layers in list
- [x] Fix padding/spacing of marker next to each layer
- [x] Add "Add layer" button
- [x] Add three-dots overflow button (for now click won't do anything)
- [ ] Style and behavior of expand/collapse tab (tab shouldn't slide in and out on hover)

Please submit each as a separate PR to help make reviewing faster and easier. Reference this issue in the PR description with "Contributes to #348".

Before:
![image](https://user-images.githubusercontent.com/228050/85954278-3a955000-b944-11ea-8102-ceef987e96f3.png)

After:
![image](https://user-images.githubusercontent.com/228050/85954372-ee96db00-b944-11ea-9dc8-40b3e83b6d1f.png)

gauravchalana1:
Hi @gino-m can I start working on this task? Thank you!

gino-m:
Hi @gauravchalana1, happy to have your contributions on this!  Please check out https://github.com/google/ground-platform/blob/master/CONTRIBUTING.md, and once ready, send a pull request with one of the above listed items to get started. Glad to have you here!

gino-m:
@gauravchalana1 is this one done and ok to close?

gino-m:
A few more tweaks coming shortly.

PR #357, opened 2020-06-30
Fix padding/spacing of layers in list

gauravchalana1:
"Contributes to #348"
image


gino-m:
Change LGTM! Please merge with latest upstream and resolve conflict so we can merge. Thanks!

gauravchalana1:
@gino-m , good to look now

PR #362, opened 2020-07-01
Add layer button

gauravchalana1:
"Contributes to #348"
image

### DIFF
--- a/web-ng/src/app/components/layer-list/layer-list.component.html
+++ b/web-ng/src/app/components/layer-list/layer-list.component.html
@@ -22,10 +22,17 @@ limitations under the License.
   <div *ngFor="let entry of layers">
     <ground-layer-list-item [layer]="entry"> </ground-layer-list-item>
   </div>
-  <div class="add-layer">
-    <button mat-raised-button class="add-layer-btn">
-      <mat-icon class="add-layer-text">add</mat-icon>
-      Add Layer
-    </button>
+
+  <div fxLayout="row" class="add-Layer">
+    <div fxLayout="column" fxFlex="fill">
+      <button
+        mat-stroked-button
+        class="add-Layer-btn"
+        (click)="onAddLayerClick(layerId)"
+      >
+        <mat-icon color="primary" class="add-Layer-icon">add</mat-icon>
+        Add Layer
+      </button>
+    </div>
   </div>
 </mat-list>
--- a/web-ng/src/app/components/layer-list/layer-list.component.html
+++ b/web-ng/src/app/components/layer-list/layer-list.component.html
@@ -20,19 +20,18 @@ limitations under the License.
   </div>
   <mat-divider></mat-divider>
   <div *ngFor="let entry of layers">
-    <ground-layer-list-item [layer]="entry"> </ground-layer-list-item>
+    <ground-layer-list-item [layer]="entry"></ground-layer-list-item>
   </div>
-
   <div fxLayout="row" class="add-Layer">
-    <div fxLayout="column" fxFlex="fill">
-      <button
-        mat-stroked-button
-        class="add-Layer-btn"
-        (click)="onAddLayerClick(layerId)"
-      >
-        <mat-icon color="primary" class="add-Layer-icon">add</mat-icon>
-        Add Layer
-      </button>
-    </div>
+    <button
+      fxFlex="fill"
+      mat-stroked-button
+      color="primary"
+      class="add-Layer-btn"
+      (click)="onAddLayer(':new')"
+    >
+      <mat-icon class="add-Layer-icon">add</mat-icon>
+      Add Layer
+    </button>
   </div>
 </mat-list>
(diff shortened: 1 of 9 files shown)


web-ng/src/app/components/layer-list/layer-list.component.scss
```diff
@@ -14,6 +14,22 @@
...
+  padding: 10px;
+}
+
+.add-layer-btn {
+  text-align: left;
+  width: 100%;
+}
+
+.add-layer-text {
+  color: mat-color($primary);
```
gino-m, line 33: This shouldn't be necessary if you specify color on the Material widget instead.

web-ng/src/app/components/layer-list/layer-list.component.html
```diff
@@ -22,4 +22,10 @@
   <div *ngFor="let entry of layers">
     <ground-layer-list-item [layer]="entry"> </ground-layer-list-item>
   </div>
+  <div class="add-layer">
+    <button mat-raised-button class="add-layer-btn">
```
gino-m, line 26: Please use `<button mat-stroked-button color="primary">` instead to match mocks more closely.

web-ng/src/app/components/layer-list/layer-list.component.html
```diff
@@ -22,4 +22,10 @@
...
+      <mat-icon class="add-layer-text">add</mat-icon>
+      Add Layer
```
gino-m, line 28: "layer" => "Layer" (lowercase)

web-ng/src/app/components/layer-list/layer-list.component.scss
```diff
@@ -14,6 +14,22 @@
...
   padding: 20px;
 }
+
+.add-layer {
+  padding: 10px;
+}
+
+.add-layer-btn {
+  text-align: left;
+  width: 100%;
```
gino-m, line 29: Would it make sense to use Flex layouts here instead of 100% width and hard-coded padding? I did this recently for the sharing dialog:

https://github.com/google/ground-platform/blob/49fbb5076e64a7850d2e6f6094762b5f556b730d/web-ng/src/app/components/share-dialog/share-dialog.component.html#L21-L44

Could this also work here?

gauravchalana1:
@googlebot I signed it!

gauravchalana1:

web-ng/src/app/components/layer-list/layer-list.component.scss
```diff
@@ -14,6 +14,20 @@
...
+
+.add-Layer {
+  padding: 10px;
+}
+
+.add-Layer-btn {
+  text-align: left;
+}
+
+.add-Layer-icon {
```
gino-m, line 31: CSS styles are generally all lower case

web-ng/src/app/components/layer-list/layer-list.component.html
```diff
@@ -20,6 +20,18 @@
...
-    <ground-layer-list-item [layer]="entry"> </ground-layer-list-item>
+    <ground-layer-list-item [layer]="entry"></ground-layer-list-item>
+  </div>
+  <div fxLayout="row" class="add-Layer">
+    <button
+      fxFlex="fill"
+      mat-stroked-button
+      color="primary"
+      class="add-Layer-btn"
+      (click)="onAddLayer(':new')"
```
gino-m, line 31: Since this function will always be called with the same value, can we can remove this arg?

web-ng/src/app/components/layer-list/layer-list.component.ts
```diff
@@ -41,4 +45,8 @@ export class LayerListComponent {
         )
       );
   }
+
+  onAddLayer(layerId: string) {
```
gino-m, line 49: Instead of taking this as an arg, can we add it as a constant in RouterService?

gauravchalana1:
Hi @gino-m , thank you for the comments, can you please review the changes? Thank you 😄

web-ng/src/app/components/layer-list/layer-list.component.html
gino-m: I see what happened- I reversed my comment, very sorry! I meant to say the text should be "Add layer", with "layer" being lowercase. Could you kindly adjust?

web-ng/src/app/components/layer-list/layer-list.component.ts
```diff
@@ -41,4 +45,8 @@ export class LayerListComponent {
         )
       );
   }
+
+  onAddLayer() {
+    this.routerService.setLayerId(':new');
```
gino-m, line 50: Please define':new' as a static constant in RouterService and reference it here.

gauravchalana1:

web-ng/src/app/components/layer-list/layer-list.component.html
gino-m: Please replace button text "Add Layer" with "Add layer" as per previous comment

gauravchalana1:
Thank you @gino-m , it's alright, I saw your message on the channel that you will be away for a vacation.

web-ng/src/app/components/layer-list/layer-list.component.html
gauravchalana1: I got it, I will make the changes as you have suggested. Thank you! 😊

gauravchalana1:
Hi @gino-m , can I please get a re-review, have resolved the conflicts. Thank you!

PR #363, opened 2020-07-02
Fix navigation of project with id :new

### LINKED ISSUE #313: [Project header] Editing project title on :new project throws errors
Instead, a new project should be created and the URL updated with the new uuid

### LINKED ISSUE #368: [Project] Create project and redirect to new id on first save of unsaved new project

parulraheja98:
Fixes #313, closes #368.


web-ng/src/app/services/router/router.service.ts
```diff
@@ -52,7 +58,17 @@ export class RouterService {
     this.activatedRoute = route;
     // Pipe values from URL query parameters.
     this.projectId$ = route.paramMap.pipe(
-      map(params => params.get('projectId'))
+      map(params => {
+        if (params.get('projectId') === ':new') {
+          const projectId = this.dataStoreService.generateId();
+          this.projectService.createProject(projectId).then(() => {
```
gino-m, line 64: In functional programming, side-effects (functions, or in this case streams, that modifying state) are generally avoided since they break "purity", one of the key qualities of the paradigm, including safe multi-threading and easy debuggability.

Also, following the Single Responsibility Principle, the RouterService's one responsibility is to abstract the URL state. Saving the project is out of scope.

To fix this, the main page component should subscribe to the `projectId$` stream, and when the id is `:new`, it should create a new project there instead.

web-ng/src/app/services/project/project.service.ts
```diff
@@ -67,4 +67,8 @@ export class ProjectService {
   updateRole(projectId: string, email: string, role: Role): Promise<void> {
     return this.dataStore.updateRole(projectId, email, role);
   }
+
+  createProject(projectId: string) {
+    return this.dataStore.updateProjectTitle(projectId, '');
```
gino-m, line 72: It's not clear from local context why we'd be calling `updateProjectTitle` here as opposed to something more general or specific. Also, if we're creating a new project, I assume we wouldn't have a projectId yet, so this would be no-arg.

Could we add a no-arg method to `DataStore`, something like `createEmptyProject()` that returns a Promise w/the new project id?

web-ng/src/app/services/router/router.service.ts
gino-m: Actually, this isn't the expected behavior; as described in #313, we'd expect the project to be created in the db only when the title is saved. The same would apply for saving layers and sharing; the project would get created on the first change an no sooner. This prevents the db from getting filled with empty abandoned projects. I'll look at the code now to see where the hook points might be.

web-ng/src/app/services/router/router.service.ts
```diff
@@ -53,7 +59,17 @@ export class RouterService {
     this.activatedRoute = route;
     // Pipe values from URL query parameters.
     this.projectId$ = route.paramMap.pipe(
-      map(params => params.get('projectId'))
+      map(params => {
+        if (params.get('projectId') === ':new') {
```
gino-m, line 63: Please use new constant in RouterService for `:new`.

web-ng/src/app/services/router/router.service.ts
gino-m: I think it's ok the redirect inside `ProjectService` or `LayerService` once a new project has been stored to the db, i.e.:

* `updateTitle` in `ProjectService` generates new id if necessary, saves, and calls `RouterService.setProjectId(..)` to new id.
* `updateLayer` in `LayerService` (#369) does the same.

Wdyt?

parulraheja98: Thanks I will work on them soon :)

gino-m:
Hi @parulraheja98! Gentle ping :) Please let me know if you have any questions!

parulraheja98:

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -184,9 +186,22 @@ export class LayerDialogComponent implements OnDestroy {
         : Map<string, Form>()
     );
 
+    if (this.projectId === ':new') {
+      this.projectService
+        .createProject(/* title= */ '')
+        .then((projectId: string) => {
```
gino-m, line 192: Can this be replaced with `projectId: => {`? (i.e. inferred type?)

web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -89,7 +89,7 @@ export class MainPageComponent implements OnInit {
     this.activeProject$.pipe(take(1)).subscribe(project =>
       this.dialog.open(LayerDialogComponent, {
         data: {
-          projectId: project.id,
+          projectId: project.isEmpty() ? ':new' : project.id,
```
gino-m, line 92: Using magic values (":new") for APIs between classes exposes implementation details across the boundaries of components (so called "leaky abstractions") and increases complexity and fragility of the system.

Since this is still a prototype, rather than design a special representation for this state, can we  move all instance of ':new' into a static const member of Project?

web-ng/src/app/shared/models/project.model.ts
```diff
@@ -31,4 +44,14 @@ export class Project {
   getLayer(layerId: string): Layer | undefined {
     return this.layers.get(layerId);
   }
+
+  isEmpty() {
```
gino-m, line 48: It might not be clear why we have `EMPTY` and `isEmpty` from the name - maybe we can call these `UNSAVED_NEW` and `isUnsavedNew()`, respectively?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
parulraheja98: Done.

web-ng/src/app/components/main-page/main-page.component.ts
parulraheja98: Done.

web-ng/src/app/shared/models/project.model.ts
parulraheja98: Done.

parulraheja98:
@gino-m , Ready for re-review.

PR #366, opened 2020-07-06
[Map] Clicking on the map should deselect selected feature

DaoyuT:


web-ng/src/app/components/main-page/main-page.component.html
```diff
@@ -20,7 +20,7 @@
   <div class="content">
     <mat-sidenav-container class="main-page-container">
       <mat-sidenav-content class="mat-sidenav-content">
-        <ground-map></ground-map>
+        <ground-map (click)="sideNavOpened = false;"></ground-map>
```
gino-m, line 23: Hey, that was easier than I thought.. cool!

gino-m: By the way, I assume clicks on map markers still work? (i.e. subcomponent events get intercepted?)

DaoyuT: Sorry, I totally overlooked it and thought it was "[Map] Clicking on the map should deselect side panel" and did that. Just added a new commit to deselect selected feature. Please review.

web-ng/src/app/components/map/map.component.html
```diff
@@ -17,7 +17,8 @@
 <google-map
   height="100%"
   width="100%"
-  [options]="mapOptions">
+  [options]="mapOptions"
+  (mapClick)="onFeatureClick('')">
```
gino-m, line 21: Can we abstract this action a little better, eg by adding a no-arg method `onFeatureBlur()`?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -59,9 +59,7 @@ export class MapComponent implements OnInit {
   ngOnInit() {
     this.subscription.add(
       this.routerService.getFeatureId$().subscribe(id => {
-        if (id) {
-          this.focusedFeatureId = id;
-        }
+        this.focusedFeatureId = id || '';
```
gino-m, line 62: Let's use `null` instead of `''` to make sure subscribers handle this special case correctly.

web-ng/src/app/components/main-page/main-page.component.html
gino-m: Ah, I hadn't had my morning coffee yet and also completely overlooked that collapsing the side panel doesn't deselect the feature! Thanks for the fix!

Does clicking on map markers still work with this fix?

DaoyuT: I reverted this change. Because with this, clicking on the markers would close the side panel. So I made this PR focus just on deselecting selected feature.

gino-m: I saw that, thanks! My question is whether registering the click event on the map component also traps clicks on map markers, in which case click events on markers themselves would not get triggered. Have you tried that and it still works?

web-ng/src/app/components/map/map.component.html
DaoyuT: Done.

web-ng/src/app/components/map/map.component.ts
DaoyuT: Done

web-ng/src/app/components/main-page/main-page.component.html
DaoyuT: Yes, I've tried. If you register the click event on the map component, clicking the markers will trigger the event.

But if you use (mapClick) like I used in the other file on the map, it means map excluding the markers (and I guess any other components as well).

PR #377, opened 2020-07-09
Fix template build errors and enable AOT compiler for all builds

gino-m:
Multiple errors were caught when running `ng build --prod`, which used the AOT compiler instead of JIT. This led to type checking being done on templates only when building for production. Switching to AOT for all builds will give us more consistent results and help us catch errors during development time.

### DIFF
--- a/web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
+++ b/web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
@@ -233,6 +233,6 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
   }
 
   get labelControl() {
-    return this.formFieldGroup.get('label');
+    return this.formFieldGroup.get('label')!;
   }
 }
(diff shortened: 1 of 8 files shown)


web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -233,6 +233,6 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
   }
 
   get labelControl() {
-    return this.formFieldGroup.get('label');
+    return this.formFieldGroup.get('label')!;
```
parulraheja98, line 236: ! might not be required here as it is always going to return control as it is present part of formFieldGroup

gino-m: Unfortunately formFieldGroup.get() is defined as being able to can return undefined, so we need to check this someplace or we get compile errors.

gino-m:
Checks passed but status not reported back to PR - forcing merge.

PR #378, opened 2020-07-10
Refactor dynamic stylesheets to reduce size of SCSS below default budget limit

gino-m:
- [x] Refactor scss color vars in separate include
- [x] Use Material component color attribute instead of custom CSS

### DIFF
--- a/web-ng/src/app/components/form-field-editor/form-field-editor.component.html
+++ b/web-ng/src/app/components/form-field-editor/form-field-editor.component.html
@@ -65,7 +65,13 @@
         (delete)="onOptionDelete(i)"
       ></app-option-editor>
     </div>
-    <button mat-button type="button" (click)="onAddOption()" class="add-option">
+    <button
+      mat-button
+      type="button"
+      (click)="onAddOption()"
+      color="primary"
+      class="add-option"
+    >
       Add option
     </button>
   </div>
--- a/web-ng/angular.json
+++ b/web-ng/angular.json
@@ -47,7 +47,7 @@
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "2mb",
+                  "maximumWarning": "3mb",
                   "maximumError": "5mb"
                 },
                 {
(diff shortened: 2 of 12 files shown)


web-ng/src/app/components/form-field-editor/form-field-editor.component.html
```diff
@@ -65,7 +65,13 @@
         (delete)="onOptionDelete(i)"
       ></app-option-editor>
     </div>
-    <button mat-button type="button" (click)="onAddOption()" class="add-option">
+    <button
+      mat-button
+      type="button"
+      (click)="onAddOption()"
+      color="primary"
```
DaoyuT, line 72: Can we put the color setting in scss?

gino-m: The color is defined on the component by design; it pulls in the theme colors (higher level) without requiring tweaking individual colors (lower level). This is a feature, not a bug :)

gino-m:
Thanks for the review!

web-ng/src/app/components/form-field-editor/form-field-editor.component.html
DaoyuT: Yes, I appreciate having themes over magic numbers here and there. I meant to have:

.add-option {
  margin-left: 30px;	  
  color: primary;
}

Does that work? Or you have to import `_variables.scss` in the component scss to do so?

gino-m: Unfortunately it wouldn't. You'd have to import `_variables.scss` and use `mat-color($primary)`. There's no reason to tweak individual DOM elements since the Material Angular component is already integrated with themes:
https://material.angular.io/components/button/overview

image

e.g.:
```html
    <button mat-stroked-button>Basic</button>
    <button mat-stroked-button color="primary">Primary</button>
    <button mat-stroked-button color="accent">Accent</button>
    <button mat-stroked-button color="warn">Warn</button>
    <button mat-stroked-button disabled>Disabled</button>
```

PR #386, opened 2020-07-14
[Layer] Delete layer

### LINKED ISSUE #15: [Layer] Delete layer
Add an option hidden behind an overflow menu on the "Customize Layer" dialog, "Delete layer".

Since this would also delete all associated data, we should show a big scary warning before deleting:

> *Delete layer?*
> This will delete all data in this layer. Are you sure you? This change cannot be undone.
>
> Cancel | Delete layer

In the future we'll start marking feature types and data as deleted rather than actually deleting it to allow 'Undo'.

DaoyuT:
Closes #15


web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -75,6 +76,25 @@ export class DataStoreService {
...
+    await this._deleteAllFeaturesInLayer(projectId, layerId);
+    return await this.db
+      .collection('projects')
+      .doc(projectId)
+      .update({
+        [`layers.${layerId}`]: firebase.firestore.FieldValue.delete(),
+      });
+  }
+
+  async _deleteAllFeaturesInLayer(projectId: string, layerId: string) {
```
gino-m, line 89: As a matter of convention we haven't been using `_` prefix or suffix for private methods; instead, please use the `private` keyword.

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -26,6 +26,7 @@ import { Layer } from './../../shared/models/layer.model';
 import { List, Map } from 'immutable';
 import { Observation } from '../../shared/models/observation/observation.model';
 import { Role } from '../../shared/models/role.model';
+import * as firebase from 'firebase/app';
```
gino-m, line 29: Can we instead import only what's needed here?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -244,4 +244,29 @@ export class LayerDialogComponent implements OnDestroy {
...
+
+    dialogRef.afterClosed().subscribe(async dialogResult => {
+      if (dialogResult) {
+        await this.deleteLayer();
+      }
+    });
+  }
+
+  async deleteLayer() {
+    await this.dataStoreService.deleteLayer(this.projectId!, this.layer!.id);
```
gino-m, line 269: Cool!

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -75,6 +76,25 @@ export class DataStoreService {
       });
   }
 
+  async deleteLayer(projectId: string, layerId: string) {
+    await this._deleteAllFeaturesInLayer(projectId, layerId);
+    return await this.db
```
gino-m, line 81: I guess we need to `return await` here instead of `await` because the caller returns a `Promise`?

gino-m:
Gentle ping - only one nit and we can merge!

web-ng/src/app/services/data-store/data-store.service.ts
DaoyuT: Done.

web-ng/src/app/services/data-store/data-store.service.ts
DaoyuT: Done.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
DaoyuT: =D

web-ng/src/app/services/data-store/data-store.service.ts
DaoyuT: Correct

DaoyuT:
> LGTM! One bit is missing - deleting all observations associated with each feature! Would you mind doing this in a follow up PR? Perhaps we need to add layerId on features to make this easier?

Cool, I'll do it next. BTW, we already have layerId field in feature.

PR #398, opened 2020-07-16
Move inline-edit-title and inline-editor into single component inline-editor.

### LINKED ISSUE #352: Move inline-edit-title and inline-editor into single component inline-editor.
- Make Styling Common between two components.
- Move the styling from styles.scss to the inline-editor component

parulraheja98:
Closes #352


web-ng/src/app/components/inline-editor/inline-editor.component.scss
```diff
@@ -14,12 +14,33 @@
...
-  margin-top: 0.83em;
-  margin-bottom: 0.83em;
-  margin-left: -8px;
-  margin-right: 0;
   font-weight: bold;
+  font-size: 14pt;
+}
+
+// Share styling of inline edit with inline-edit-title and inline-editor component.
+.inline-editor-input {
```
gino-m, line 25: I believe this class is now the same as `:host`, so this style rule can be merged into `:host` and removed.

web-ng/src/app/components/inline-editor/inline-editor.component.scss
```diff
@@ -14,12 +14,33 @@
...
+  font-family: inherit;
+  font-size: inherit;
+  border: 2px solid transparent;
+  outline: none;
+  border-radius: 3px;
+  padding: 2px 1px 2px 4px;
+}
+
+// !important used to override hover border color.
+.inline-editor-input:focus {
```
gino-m, line 35: Does `:host:focus` and `:host:hover` work? If not please ignore previous comment.

web-ng/src/app/components/project-header/project-header.component.ts
```diff
@@ -14,19 +14,42 @@
...
+
+  subscription: Subscription = new Subscription();
+  constructor(
+    public auth: AuthService,
+    private dialog: MatDialog,
+    private projectService: ProjectService
+  ) {
+    this.lang = 'en';
+    this.title = '';
+    this.activeProject$ = this.projectService.getActiveProject$();
```
gino-m, line 45: Since `activeProject$` isn't used elsewhere in this case, can we make it a `const` local variable?

web-ng/src/app/components/project-header/project-header.component.ts
```diff
@@ -43,4 +66,21 @@ export class ProjectHeaderComponent implements OnInit {
...
+  /**
+   * Updates the project title with input element value.
+   *
+   * @param evt the event emitted from the input element on blur.
+   */
+  updateProjectTitle(value: string) {
+    if (!this.projectId) {
+      return Promise.reject(new Error('Project not loaded'));
+    }
+    if (value === this.title) return;
```
gino-m, line 79: This would return `undefined`, while other paths return `Promise`. Not sure what the event handler does with the return value, but can we return `Promise.resolve()` here instead for consistency?

web-ng/src/app/components/inline-editor/inline-editor.component.scss
parulraheja98: Sorry this doesn't work as host is for parent container but we are trying to give styling to input specifically

PR #399, opened 2020-07-16
Fix layer title and saving of layer.

### LINKED ISSUE #367: New Layer shows title undefined
When a new layer is created using #l=:new route parameter it shows layer title undefined.
image

parulraheja98:
This fixes layer title undefined and layer not being to able save issue.
Closes #367


gino-m:
Thanks for the fix!

PR #404, opened 2020-07-17
[User profile] Avatar image broken when user has no profile pic

### LINKED ISSUE #236: [User profile] Avatar image broken when user has no profile pic
We may need to implement the circle w/the first letter of the user name as is done in other profile/avatar components.

parulraheja98:
image

image

Closes #236


web-ng/src/app/components/project-header/project-header.component.html
```diff
@@ -30,12 +30,18 @@
...
+        *ngIf="user.photoURL; else defaultPhoto"
         src="{{ user.photoURL }}=s64"
-        alt="{{ user.displayName }}"
         class="user-thumb"
-        style="cursor: pointer;"
         (click)="openProfileDialog($event)"
       />
+      <ng-template #defaultPhoto>
+        <span
+          [attr.data-letters]="user.displayName.charAt(0)"
```
gino-m, line 40: How does this actually work? What component causes the letter to be read from the data-letters and attribute? And why do we need :before pseudo-elements in CSS?

web-ng/src/app/components/user-profile-popup/user-profile-popup.component.scss
```diff
@@ -42,10 +45,24 @@
...
-  margin-top: 10px ;
+  margin-top: 10px;
   text-align: center;
 }
 
 .btn-signout {
   margin-top: 20px;
 }
+
+[data-letters]:before {
```
gino-m, line 56: Duplicated UI widgets should implemented as a separate component (e.g., user-avatar-component).

parulraheja98: Done..

web-ng/src/app/components/project-header/project-header.component.html
parulraheja98: Remove pseudo attr.

PR #407, opened 2020-07-18
Refactor access to data store out of UI component into new LayerService

### LINKED ISSUE #369: [Layer dialog] Refactor access to data store out of UI component into new LayerService
UI widgets should not reference `DataStore`, which is datastore-specific, directly, but rather via domain-specific services. See ProjectService for an example.

parulraheja98:
Closes #369


web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -132,60 +122,52 @@ export class LayerDialogComponent implements OnDestroy {
...
-      const fieldId = layerFieldId
-        ? layerFieldId
-        : this.dataStoreService.generateId();
-      fields = fields.set(fieldId, field);
-    });
-    const form = this.getForms();
-    const formId = form ? form.id : this.dataStoreService.generateId();
+    const fields = this.layerService.convertFieldsListToMap(this.fields);
+    const form = this.layerService.getForm(this.layer);
+    const forms = form
```
gino-m, line 157: I think the current logic could be further simplified - instead of extracting the form id from the layer using the service, can we add it as a member to this component? Then we wouldn't need to merge the previous form with the new fields; the component will have id and fields that could be passed to the LayerService to create or update the form.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -132,60 +122,52 @@ export class LayerDialogComponent implements OnDestroy {
...
+      ? this.layerService.setForms(form.id, new Form(form.id, fields))
+      : this.layerService.createForm(fields);
```
gino-m, line 159: In that case we wouldn't need the ternary; you'd just have `forms =  this.layerService.createForm(formId, fields)`.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -132,60 +122,52 @@ export class LayerDialogComponent implements OnDestroy {
...
       /* index */ -1,
       this.color,
       // TODO: Make layerName Map
       StringMap({ [this.lang]: this.layerName }),
-      this.fields && this.fields.size > 0
-        ? Map({
-            [formId]: new Form(formId, fields),
-          })
-        : Map<string, Form>()
+      forms ? forms : Map<string, Form>()
```
gino-m, line 166: Forms is never undefined, so the ternary isn't required here.

web-ng/src/app/services/layer/layer.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  createForm(
+    id: string | undefined,
+    fields: Map<string, Field>
+  ): Map<string, Form> {
+    const formId = id || this.dataStoreService.generateId();
+    return this.setForms(formId, new Form(formId, fields));
+  }
+
+  setForms(id: string, form: Form): Map<string, Form> {
```
gino-m, line 93: This method name is still confusing; `setForms` implies this will somehow alter the state of the service, since `set` is frequently used for mutator methods. 

Maybe `createFormMap()` or similar?

Also, since this method doesn't touch internal state, it should be marked `static`.

web-ng/src/app/services/layer/layer.service.ts
```diff
@@ -0,0 +1,103 @@
...
+import { StringMap } from '../../shared/models/string-map.model';
+import { Option } from '../../shared/models/form/option.model';
+import { MultipleChoice } from '../../shared/models/form/multiple-choice.model';
+import { List, Map } from 'immutable';
+import { Form } from '../../shared/models/form/form.model';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LayerService {
```
gino-m, line 30: Please add essential doc comments throughout (class and public methods).

parulraheja98:

web-ng/src/app/services/layer/layer.service.ts
```diff
@@ -30,11 +30,23 @@ import { Form } from '../../shared/models/form/form.model';
...
 
+  /**
+   * Creates a new field with a given type, label, required, index and multipleChoice values.
+   *
+   * @param type the type of the new field.
+   * @param label the label of the new field.
+   * @param required the required value of the new field.
+   * @param index the index of the new field.
+   * @param multipleChoice the multipleChoice value of the new field.
+   */
```
gino-m, line 49: Thanks for adding these! Suggestion for adding comments in general: it's generally preferable to not repeat what's already clear from code in the comments. The idea being the extra text should add clarity rather than additional work the reader, and that if something can be clarified directly in code, even better. For example, instead of the block, we might simply write:

```
  /**
   * Creates and returns a new field with a generated unique identifier and a single English label.
   */
```

Not that method arg JSDoc are not required here, since TypeScript already lets us declare types, and there's nothing interesting to add to the comments. Could you kindly update similarly throughout?

parulraheja98: Done.

parulraheja98:

parulraheja98:
@gino-m  , Resolved Conflicts :)

PR #418, opened 2020-07-23
Change how side panel selects content based on RouterService

### LINKED ISSUE #374: Routing from a feature to primary url does not change the side panel back to list of layers view
**Describe the bug**
Routing from a feature to primary url (e.g. http://localhost:4200/p/example)  (deselecting the feature) does not change the side panel back to list of layers view.

**To Reproduce**
1. Go to home page.
2. Shows list of layers in side panel.
3. Click on a feature.
4. Shows details of the feature in side panel.
5. Click on the map to deselect the feature.
6. Still shows the details of that feature in side panel.

**Expected behavior**
Should show the list of layers view in side panel.

gyarill:
fixes #374


web-ng/src/app/components/side-panel/side-panel.component.ts
```diff
@@ -14,36 +14,29 @@
...
 
 @Component({
   selector: 'ground-side-panel',
   templateUrl: './side-panel.component.html',
   styleUrls: ['./side-panel.component.css'],
 })
 export class SidePanelComponent {
-  readonly sideNavContentType = SideNavContentType;
-  readonly sideNavContentType$: Observable<SideNavContentType>;
+  readonly sideNavContentMode = SideNavContentMode;
```
gino-m, line 31: Nit: s/contentMode/Mode/g

web-ng/src/app/components/side-panel/side-panel.component.ts
```diff
@@ -14,36 +14,29 @@
...
+  readonly sideNavContentMode$: Observable<SideNavContentMode>;
   readonly lang: string;
 
-  constructor(
-    private routerService: RouterService,
-    route: ActivatedRoute) {
+  constructor(private routerService: RouterService, route: ActivatedRoute) {
     routerService.init(route);
```
gino-m, line 36: `routerService` should be a singleton, and `init()` is already called in `MainPageComponent`, so should probably be removed.

web-ng/src/app/services/router/router.service.ts
```diff
@@ -69,19 +69,17 @@ export class RouterService {
...
     );
-    this.sideNavContentType$ = fragmentParams$.pipe(
-      map(params => { 
-        console.log(params);
-        console.log(params.get(RouterService.OBSERVATION_ID_FRAGMENT_PARAM));
+    this.sideNavContentMode$ = fragmentParams$.pipe(
+      map(params => {
         if (params.get(RouterService.OBSERVATION_ID_FRAGMENT_PARAM)) {
-          return SideNavContentType.OBSERVATION;
+          return SideNavContentMode.OBSERVATION;
```
gino-m, line 75: Can you extract this logic in a static function that maps params to SideNavContentMode? It would make this pipe a little more friendly to read since it would separate the Rx plumbing (no pun intended) from the conversion logic.

web-ng/src/app/services/router/router.service.ts
```diff
@@ -164,8 +162,8 @@ export class RouterService {
   }
 }
 
-export enum SideNavContentType {
-  LAYER_LIST = 0,
-  OBSERVATION = 1,
-  FEATURE = 2,
+export enum SideNavContentMode {
+  LAYER_LIST = 1,
```
gino-m, line 166: Can we return this to 0, 1, 2 now?

gyarill: unfortunately still does not work with 0 (even with switch)

web-ng/src/app/components/side-panel/side-panel.component.ts
gyarill: done

web-ng/src/app/components/side-panel/side-panel.component.ts
gyarill: done

web-ng/src/app/services/router/router.service.ts
gyarill: done

web-ng/src/app/services/router/router.service.ts
gino-m: Very odd.. I don't think `switch` would ignore falsey values, so something else is probably amiss. Let's leave it like this for now, but would good to understand why it's happening!

Issue #251, opened 2020-03-27
[Map] Add new point feature to map


gino-m:
Ability to add a new point to the map. @jacobmclaws, how will the user select the layer and location for new points? Whatever solution we come up with should also scale to drawing polygons eventually.

gino-m:
Reassigning to PM @sergeydolgov1. @jacobmclaws please accept new invite to repo so I can also assign to you. Will assign back to eng once the CUJ is defined.

gino-m:
@sergeydolgov1 @jacobmclaws Gentle ping:

> How will the user select the layer to which to add new points?

Please paste mocks here for @DaoyuT.

sergeydolgov1:
@gino-m @DaoyuT 

A) User clicks on the layer for which they want to add Point:
![image](https://user-images.githubusercontent.com/20148684/87334503-9ab6f500-c4f3-11ea-998f-721443ab32a7.png)

sergeydolgov1:
@gino-m @DaoyuT 

B) To add new Point user clicks on the Point button in the lower left corner 
![image](https://user-images.githubusercontent.com/20148684/87334632-cfc34780-c4f3-11ea-9594-ac7f9f7efdec.png)

gino-m:
@sergeydolgov1 Could you kindly paste the mock here?

sergeydolgov1:
@gino-m @DaoyuT 
This is the latest mock (see the link below)
Question from Jacob: _"Is it possible to have some smarts about which layer a user was looking at in the side panel and defaulting to that layer when they create or will it always just be the top layer in the list?"_ [This feature would be optional]
[Add New Point.pdf](https://github.com/google/ground-platform/files/4963832/Add.New.Point.pdf)

gino-m:
@sergeydolgov1 Thanks again for the wireframe! @jacobmclaws Could you a provide actual updated mocks for this flow and the "polygon" icon asset as well?

gino-m:
From @sergeydolgov1 (moved from https://github.com/google/ground-android/issues/358#issuecomment-671541451):

> Prioritized drawing a polyline (Lines + Polygons) for RC1. Latest mocks below:
>
> ![image](https://user-images.githubusercontent.com/20148684/89821845-18c2e780-db04-11ea-8c31-b07b0fdfb8e0.png)
>
> ![image](https://user-images.githubusercontent.com/20148684/89821861-211b2280-db04-11ea-9f66-85103f5ca9b7.png)

[...]

PR #425, opened 2020-08-06
[Map] Add point feature on click

gino-m:
This is a temporary solution to allow us to add features to the map for manual testing. It will cause a new feature to be added to an arbitrary layer when the map is clicked and when no feature is selected.

Contributes to #251.

### DIFF
--- a/web-ng/src/app/services/feature/feature.service.ts
+++ b/web-ng/src/app/services/feature/feature.service.ts
@@ -85,6 +85,7 @@ export class FeatureService {
     }
     const newFeature = new Feature(
       this.dataStore.generateId(),
+      // TODO(#251): When we implement the real "add point" flow, use selected layer instead of first.
       project.layers.keySeq().first(),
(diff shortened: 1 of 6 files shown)


web-ng/src/app/services/feature/feature.service.ts
```diff
@@ -66,4 +68,28 @@ export class FeatureService {
...
+      .then(project => this.addPointInternal(project, lat, lng));
+  }
+
+  private addPointInternal(project: Project, lat: number, lng: number) {
+    if (project.layers.isEmpty()) {
+      return Promise.resolve();
+    }
+    const newFeature = new Feature(
+      this.dataStore.generateId(),
+      project.layers.keySeq().first(),
```
DaoyuT, line 88: Can we add a comment or TODO here since adding to arbitrary layer is a temporary solution?

gino-m: Sorry, could have sworn I did this.. adding one now!

PR #429, opened 2020-08-06
Select One field added in drop down

### LINKED ISSUE #347: [Form designer] Add "Select one" field type to drop down
This will look the same as "Select many", except using the "radio button" Material icon in place of the "checkbox" icon:
https://material.io/resources/icons/?icon=radio_button_checked&style=baseline

The result of selecting this field is that the current form field will get `type: 'multiple_choice'` and `cardinality: 'select_many'` as described in https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation.

The behavior of the field will be identical to `type: 'multiple_choice'` with `cardinality: 'select_one'`.

gauravchalana1:
Fix #347
@gino-m Please review, ty :) 
image

### DIFF
--- a/web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
+++ b/web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
@@ -106,9 +106,9 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
     });
   }
 
-  /*
-    Input params change from the parent are handled by this method.
-  */
+  /**
+   * Input params change from the parent are handled by this method.
+   */
   ngOnChanges(changes: SimpleChanges) {
     if (changes.multipleChoice) {
       this.formOptions = this.multipleChoice;
@@ -139,7 +139,6 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
    * Emits the delete field event to the layer dialog component.
    *
    * @returns void
-   *
    */
   onFieldDelete() {
     this.delete.emit();
@@ -154,7 +153,6 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
    *
    * @param event: FieldTypeSelectOption
    * @returns void
-   *
    */
   onFieldTypeSelect(event: FieldTypeSelectOption) {
     this.type = event.type;
@@ -185,9 +183,7 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
    * @param event: label and code value of the option field.
    * @param index: index of the option to be updated.
    * @returns void
-   *
    */
-
   onOptionUpdate(event: { label: string; code: string }, index: number) {
     const option = this.layerService.createOption(
       event.code,
--- a/web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
+++ b/web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
@@ -42,6 +42,7 @@ export interface FieldTypeSelectOption {
   icon: string;
   label: string;
   type: FieldType;
+  cardinality?: Cardinality;
 }
 
 @Component({
@@ -54,6 +55,7 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
   @Input() required?: boolean;
   @Input() type?: string;
   @Input() multipleChoice?: MultipleChoice;
+  @Input() cardinality?: Cardinality;
   @Output() update = new EventEmitter();
   @Output() delete = new EventEmitter();
   formOptions: MultipleChoice | undefined;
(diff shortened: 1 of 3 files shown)


gauravchalana1:
@gino-m Can I please get a re-review? ty

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -98,9 +107,13 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    const type = this.fieldTypes.find(
-      fieldType => fieldType.type === Number(this.type)
-    );
+    const type = !this.cardinality
```
gino-m, line 110: Cardinality only applies in when "Select one" or "Select many" is selected, but here it seems it's always evaluated. Is that intentional?

parulraheja98: I think this is for property passed from parent to child for fields with stored values of "Select one" or "Select many". @gauravchalana1  am I right?

gino-m: A few things seem to be muddling my understanding here:
1. I see we also have `onFieldTypeSelect` defined. If that's the case, why do we also have `ngOnChanges`? I assume it's because @Input can be changed by the parent after the component has been constructed? If so let's add a comment to clarify.
2. Assuming my understanding in (1) correct, we should first check the type provided; if it's anything other than `multiple_choice`, we should ignore cardinality. If it is `multiple_choice`, only then do we care about `cardinality`. Based on this the current logic may work, but it's counterintuitive given the logical hierarchy is field type->cardinality, not cardinality->field type.

gauravchalana1: Hi @gino-m , I have now tried and updated the function as you have suggested. Please review and share your comments. :) ty

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -97,10 +106,18 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
     });
   }
 
+  /*
+    Input params change from the parent are handled by this method.
```
gino-m, line 110: Nit: s/change/changes/

Also, please use method doc style:
/**
 * ...
 */

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -97,10 +106,18 @@ export class FormFieldEditorComponent implements OnInit, OnChanges {
...
+  */
   ngOnChanges(changes: SimpleChanges) {
-    const type = this.fieldTypes.find(
-      fieldType => fieldType.type === Number(this.type)
-    );
+    const type =
```
gino-m, line 113: This is better, but I think a switch statement would be clearer here; find+lambda is overkill considering there are only two items that would potentially match. Also, the fact that "type" the local var refers to something different than "type" the class member is confusing. I'd simplify as follows:

1. Rename class member "type" to "fieldType".
2. Extract switch block into a helper function and refer to it directly when calling setValue below.

For ex:
```
this.formFieldGroup.setValue({
  label: this.label,
  required: this.required,
  type: this.getSelectedFieldTypeOption()
});
```

Then getSelectedFieldTypeOption() would be something like:
```
switch (this.type) {
  case FieldType.TEXT:
    return FormFieldEditor.FIELD_TYPE_OPTIONS[0];
  case FieldType.MULTIPLE:
    return FormFieldEditor.FIELD_TYPE_OPTIONS[this.cardinality === FieldCardinality.SINGLE ? 1 : 2];
  default:
    throw new Error(`Unsupported field type${this.type}`);
}
```

I'd also make "this.fieldTypes" into a private static constant "FIELD_TYPE_OPTIONS" to avoid further confusion.

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -52,12 +53,19 @@ export interface FieldTypeSelectOption {
...
   @Output() update = new EventEmitter();
   @Output() delete = new EventEmitter();
   formOptions: MultipleChoice | undefined;
   fieldTypes: FieldTypeSelectOption[] = [
+    {
+      icon: 'radio_button_checked',
+      label: 'Select One',
+      type: FieldType.MULTIPLE_CHOICE,
+      cardinality: Cardinality.SELECT_ONE,
+    },
```
gino-m, line 68: The correct order should be "text', "select one", "select multiple".

gauravchalana1:
Hi @gino-m , I have made the changes as suggested. Please check and share your feedback. ty :)

gino-m (approved):

gino-m:
LGTM! Thanks for your patience on this super long review!

gauravchalana1:
Hi @gino-m , thank you for your patience and feedback. :)

PR #432, opened 2020-08-10
Render GeoJson in features.

DaoyuT:
- [x] Render arbitrary GeoJSON stored in Feature in Firebase
- [x] Use default style of layer for styling

Screenshot:
Screen Shot 2020-08-10 at 1 13 26 AM

### DIFF
--- a/web-ng/src/app/shared/converters/firebase-data-converter.ts
+++ b/web-ng/src/app/shared/converters/firebase-data-converter.ts
@@ -176,8 +176,9 @@ export class FirebaseDataConverter {
         geoJson,
       };
     } else {
-      const { layerId } = feature;
-      return { layerId };
+      throw new Error(
+        `Cannot convert unexpected feature class ${feature.constructor.name} to json.`
+      );
     }
   }
   /**
--- a/web-ng/src/app/shared/converters/firebase-data-converter.ts
+++ b/web-ng/src/app/shared/converters/firebase-data-converter.ts
@@ -24,7 +24,11 @@ import {
   MultipleChoice,
   Cardinality,
 } from '../models/form/multiple-choice.model';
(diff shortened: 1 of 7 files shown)


web-ng/src/app/shared/converters/firebase-data-converter.ts
```diff
@@ -158,12 +162,23 @@ export class FirebaseDataConverter {
...
+        location,
+      };
+    } else if (feature instanceof GeoJsonFeature) {
+      const { layerId, geoJson } = feature;
+      return {
+        layerId,
+        geoJson,
+      };
+    } else {
+      const { layerId } = feature;
```
gino-m, line 179: If the application itself is generating features of unknown types, we should probably throw an exception here. Wdyt?

DaoyuT:
> Great work, really excited to see GeoJSON support working! Iiuc this impl only updates GeoJson features when the map is initialized, but not on remote changes. Will you implement that bit in a follow-up PR?

What do you mean? I tried adding/updating features in firebase, and then website reflects the update without me refreshing it. Is that what you meant?

web-ng/src/app/shared/converters/firebase-data-converter.ts
DaoyuT: Done.

gino-m:
> > Great work, really excited to see GeoJSON support working! Iiuc this impl only updates GeoJson features when the map is initialized, but not on remote changes. Will you implement that bit in a follow-up PR?
> 
> What do you mean? I tried adding/updating features in firebase, and then website reflects the update without me refreshing it. Is that what you meant?

Sorry, you're right; I see you're subscribing in ngOnViewInit, not actually adding/removing features directly there. Please disregard!

Issue #266, opened 2020-04-20
[Side panel] Add "back" button to deselect current feature and return to layers list


gino-m:

gino-m:
@sergeydolgov1 @jacobmclaws updated as discussed.

PR #437, opened 2020-08-13
add close icon to layer header on Side Panel when Feature or Observation is selected

gyarill:
Clicking it makes it go back to Layer List.

Closing #266


web-ng/src/app/components/layer-list-item/layer-list-item.component.html
```diff
@@ -19,10 +19,19 @@
...
+  <div [ngSwitch]="actionsType">
+    <div *ngSwitchCase="layerListItemActionsType.MENU" >
+      <button mat-icon-button [matMenuTriggerFor]="layerMenu">
+        <mat-icon>more_vert</mat-icon>
+      </button>
+      <mat-menu #layerMenu="matMenu">
+        <button mat-menu-item (click)="onCustomizeLayer()">Customize layer</button>
+      </mat-menu>
+    </div>
+    <div *ngSwitchCase="layerListItemActionsType.BACK" >
```
gino-m, line 31: Does this conditional appear in the LayerListItemComponent because we reuse it for both the list and the feature details header? If so, can we refactor this component so that we don't reuse it directly in both places? I assume the UI is going to diverge a lot between the list items and the header, so this might make more sense. Wdyt?

gino-m: Please disregard - we'll return to this once we know what the header will look like in other panels.

gyarill: I added TODO to change it when we know UI as we discussed ofline

PR #447, opened 2020-08-31
Add workflow for adding a point feature.

DaoyuT:
For issue #251

Screenshots:
Screen Shot 2020-08-31 at 8 13 04 PM
Screen Shot 2020-08-31 at 8 12 24 PM
Screen Shot 2020-08-31 at 5 09 33 PM

 - Clicking on the buttons enables the dropdown;
 - The dropdown pick the first layer as default;
 - Clicking on `add point` deselects `add polygon`, and vice versa;
 - Clicking on `add point` when `add point` is selected exits the edit mode, same for `add polygon`;
 - Change cursor to `crosshair` when adding a point (cannot show from screenshot but you can try).


web-ng/src/app/components/drawing-kit/drawing-kit.component.css
```diff
@@ -0,0 +1,29 @@
...
+  position: absolute;
+  bottom: 3em;
+  left: 3em;
+}
+
+.drawing-layer-selector {
+  position: absolute;
+  top: 2em;
+  left: 50%;
+  background-color: #fff;
```
gino-m, line 27: Can we use Angular Material background color here instead of hex literal?

web-ng/src/app/components/drawing-kit/drawing-kit.component.html
```diff
@@ -0,0 +1,37 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<mat-button-toggle-group 
+class="drawing-buttons"
+[(value)] = "selectedValue">
+  <mat-button-toggle value = "{{pointValue}}" (click)="onButtonClick()">
```
gino-m, line 19: Please add a TODO here to replace with icons. Could you also please add a GitHub issue to track?

web-ng/src/app/components/drawing-kit/drawing-kit.component.ts
```diff
@@ -0,0 +1,82 @@
...
+import { getPinImageSource } from '../map/ground-pin';
+import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
+
+@Component({
+  selector: 'drawing-kit',
+  templateUrl: './drawing-kit.component.html',
+  styleUrls: ['./drawing-kit.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DrawingKitComponent {
```
gino-m, line 36: It's not immediately obvious what "drawing kit" and related service refers to; is it all the drawing tools, the drop down, or the toolbar? If we can choose a more specific name let's do that, if not please add JSDoc to clarify.

web-ng/src/app/components/main-page/main-page.component.html
```diff
@@ -21,6 +21,7 @@
     <mat-sidenav-container class="main-page-container">
       <mat-sidenav-content class="mat-sidenav-content">
         <ground-map></ground-map>
+        <drawing-kit></drawing-kit>
```
gino-m, line 24: Maybe `drawing-tools`?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -94,20 +106,32 @@ export class MapComponent implements OnInit, AfterViewInit {
         };
       })
     );
+
+    this.drawingKitService.getEditMode$().subscribe(editMode => {
```
gino-m, line 110: Please save the subscription and dispose in `ngOnDestroy()` to prevent memory leaks.

web-ng/src/app/services/drawing-kit/drawing-kit.service.ts
```diff
@@ -0,0 +1,50 @@
...
+import { BehaviorSubject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DrawingKitService {
+  private editMode$: BehaviorSubject<EditMode> = new BehaviorSubject<EditMode>(
+    EditMode.None
+  );
+  private layerId = '';
```
gino-m, line 27: `selectedLayerId`?

Also, we should use `null` or `undefined` so the type system can protect us from cases where we don't handle the default state.

web-ng/src/app/services/drawing-kit/drawing-kit.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  getLayerId(): string {
+    return this.layerId;
+  }
+
+  setLayerId(layerId: string) {
+    this.layerId = layerId;
+  }
+
+  getEditMode$(): BehaviorSubject<EditMode> {
```
gino-m, line 37: Return `Observable<EditMode>` here instead to discourage callers from writing to the stream.

gino-m (changes requested):
Sorry for the slow reply - looks great! A few tweaks, potential improvements, HTH!

gino-m:
@DaoyuT Gentle ping!

web-ng/src/app/components/drawing-tools/drawing-tools.component.css
```diff
@@ -24,6 +24,6 @@
   position: absolute;
   top: 2em;
   left: 50%;
-  background-color: #fff;
+  background-color: white;
```
gino-m, line 27: I was referring to using the `mat-color()` directive, but since this is a temp UI we can leave it as-is for now.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -83,56 +89,68 @@ export class MapComponent implements OnInit, AfterViewInit {
...
-        event.latLng.lng(),
-        this.drawingKitService.getLayerId()
-      );
-    } else if (editMode === EditMode.AddPolygon) {
-      // TODO: Implement adding polygon.
-      return Promise.resolve();
-    } else {
+    const editMode = this.drawingToolsService.getEditMode$().getValue();
+    const selectedLayerId = this.drawingToolsService.getSelectedLayerId();
+    if (selectedLayerId === undefined) {
```
gino-m, line 137: I think `!selectedLayerId` would be more idiomatic.

web-ng/src/app/services/drawing-tools/drawing-tools.service.ts
```diff
@@ -20,18 +20,18 @@ import { BehaviorSubject } from 'rxjs';
 @Injectable({
   providedIn: 'root',
 })
-export class DrawingKitService {
+export class DrawingToolsService {
   private editMode$: BehaviorSubject<EditMode> = new BehaviorSubject<EditMode>(
     EditMode.None
   );
-  private layerId = '';
+  private selectedLayerId: string | undefined;
```
gino-m, line 27: Does `string?` work here and below instead of `| undefined`?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -83,56 +89,68 @@ export class MapComponent implements OnInit, AfterViewInit {
...
-
-    this.activeProject$.subscribe(project =>
-      this.map.data.setStyle(feature => {
-        const layerId = feature.getProperty('layerId');
-        const color = project.layers.get(layerId)?.color;
-        return {
-          fillColor: color,
-        };
+    this.subscription.add(
+      this.features$.subscribe(features => {
```
gino-m, line 93: Extracting these three handlers into their own methods might be more readable since it'll separate subscription/binding from the handling logic. Carefully named methods like `onFeaturesUpdate()`, `onProjectActivation` and `onEditModeChange` may also make the code more self-describing.

gino-m (dismissed):
Thanks for the edits, much better! Just a few nits, otherwise ready to go.

web-ng/src/app/components/drawing-tools/drawing-tools.component.css
DaoyuT: Cool, I'll leave it as-is

web-ng/src/app/services/drawing-tools/drawing-tools.service.ts
DaoyuT: No the syntax doesn't work

gino-m (approved):
Great work, congrats on getting this done!

PR #451, opened 2020-09-07
[Feature panel] Delete feature

### LINKED ISSUE #11: [Feature panel] Delete feature
Let's add the "Delete" entry to an overflow menu in the top-right of the feature details in the side panel when the feature is selected. Deleting the feature should also delete all related observations; we should show a BSW (big scary warning) confirmation dialog before doing this, i.e. "Deleting this map feature will delete all associated data, including any collected observations. This cannot be undone. Are you sure?" @sergeydolgov1 fyi/for comment.

gauravchalana1:
Closes #11.

Hi @gino-m 
Please review the Delete feature #11 function
Please share your comments on it and what can be improved.
A few pointers:
1. Where should be the positioning of this function?
2. I have commented on a few lines of prod code for testing of the function.

PFA screenshots
image
image


web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -95,6 +95,28 @@ export class DataStoreService {
     return await Promise.all(querySnapshot.docs.map(doc => doc.ref.delete()));
   }
 
+  async deleteFeature(projectId: string, featureId: string) {
+    await this.deleteAllObservationsInFeature(projectId, featureId);
+    return await this.db
```
gino-m, line 100: Since .delete() itself returns a Promise, I think this `await` is redundant (it will unwrap the result from the Promise, which will get rewrapped by being returned from an `async` function.

web-ng/src/app/services/data-store/data-store.service.ts
```diff
@@ -95,6 +95,28 @@ export class DataStoreService {
...
+  private async deleteAllObservationsInFeature(
+    projectId: string,
+    featureId: string
+  ) {
+    const observationsInFeature = this.db.collection(
+      `projects/${projectId}/observations`,
+      ref => ref.where('featureId', '==', featureId)
+    );
+    const querySnapshot = await observationsInFeature.get().toPromise();
+    return await Promise.all(querySnapshot.docs.map(doc => doc.ref.delete()));
```
gino-m, line 117: Same here.

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
```diff
@@ -29,13 +33,18 @@ export class LayerListItemComponent implements OnInit {
   @Input() layer: Layer | undefined;
   @Input() actionsType: LayerListItemActionsType =
     LayerListItemActionsType.MENU;
+  projectId!: string | null;
+  featureId!: string | null;
```
gino-m, line 37: Any reason not to just define these two as optional with the `?` type modifier?

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
```diff
@@ -46,6 +55,12 @@ export class LayerListItemComponent implements OnInit {
     this.layerPinUrl = this.sanitizer.bypassSecurityTrustUrl(
       getPinImageSource(this.layer?.color)
     );
+    this.routerService.getFeatureId$().subscribe(id => {
+      this.featureId = id;
+    });
+    this.routerService.getProjectId$().subscribe(id => {
```
gino-m, line 61: We should add the subscriptions to a composite Subscription as we do in other classes and unsubscribe on ngOnDestroy().

gino-m (changes requested):
Nice job, Gaurav, looks great so far!

Please place the delete option in an overflow menu (three dots) shown at the top-right of the feature panel (i.e., in the header next to the layer name) as described in #11. A few other comments inline. Thanks!

gino-m:
@gauravchalana1 gentle ping! Note that tests are still failing, but I assume that's expected for the draft.

gauravchalana1:
Hi @gino-m , yes I am supposing that the tests are failing because it’s a draft PR. Ty

gauravchalana1:
Hi @gino-m , the changes have been made as you have suggested. PFB screenshots and share your reviews. ty :) 
image
image

web-ng/src/app/components/feature-panel/feature-panel.component.html
```diff
@@ -29,9 +31,13 @@
       </mat-card-title>
       <!--TODO: figure out why serverTime is not working-->
       <mat-card-subtitle>
-        {{ observation.created.clientTime | date:'medium' }}
+        {{ observation.created.clientTime | date: 'medium' }}
       </mat-card-subtitle>
-      <button mat-icon-button [matMenuTriggerFor]="menu" aria-label="actions">
+      <button
```
gino-m, line 36: Looks like you reverted the change that adds the overflow menu (three dots). I assume this wasn't intentional?

web-ng/src/app/components/feature-panel/feature-panel.component.html
```diff
@@ -42,17 +48,28 @@
...
-          <label class="field-label">{{ field_entry[1].label?.get(lang) }}</label>
-          <div class="field-response" *ngIf="field_entry[1].type == fieldTypes.TEXT">
-            {{ observation.responses?.get(field_entry[1].id)?.value }}
-          </div>
-          <div class="field-response" *ngIf="field_entry[1].type == fieldTypes.MULTIPLE_CHOICE">
-            <span class="option-label" *ngFor="let option of observation.responses?.get(field_entry[1].id)?.value">
-              {{ option.label?.get(lang) }}
-            </span>
-          </div>
+      <div *ngFor="let field_entry of observation.form.fields">
```
gino-m, line 51: Out of scope of this PR, but while we're changing these lines anyway -  looks like we only use field_entry[1] , which contains the values in the `fields` map. Is there a way to just iterate over those instead? Maybe we can expose them as a property of the component?

web-ng/src/app/components/layer-list-item/layer-list-item.component.html
```diff
@@ -19,26 +19,26 @@
...
-      </button>
-    </div>
-  </div>
+
+  <button mat-icon-button [matMenuTriggerFor]="layerMenu">
+    <mat-icon>more_vert</mat-icon>
+  </button>
+  <mat-menu #layerMenu="matMenu">
+    <button mat-menu-item (click)="onDeleteFeature()">
+      <mat-icon>delete_forever</mat-icon>
```
gino-m, line 28: Let's put Delete as the last item in the menu.

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
```diff
@@ -31,16 +35,19 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
   @Input() layer?: Layer;
   @Input() actionsType: LayerListItemActionsType =
     LayerListItemActionsType.MENU;
+  projectId?: string | null;
```
gino-m, line 38: Do we still need `| null` with `?`?

gino-m (changes requested):
Nice work! A few suggestions, otherwise LG!

web-ng/src/app/components/feature-panel/feature-panel.component.html
gauravchalana1: It was intentional for testing this feature, fixed it. ty :)

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
gauravchalana1: yes we do need | null with ? as in the router service we are projectId?: string | null because paramap returns an observable of string | null type.

web-ng/src/app/components/feature-panel/feature-panel.component.html
gino-m: I tried to fix this directly in your PR, but it looks like I don't have write access to your branch; I'll send a follow-up PR once this is merged.

gino-m (approved):
LGTM.. 🚀  Nice work! 💯

PR #457, opened 2020-10-08
[Map] Move/reposition points

DaoyuT:
- [x] make marker draggable when it is focused.
 - [x] update feature's location in firebase.

### DIFF
--- a/web-ng/src/app/components/map/map.component.ts
+++ b/web-ng/src/app/components/map/map.component.ts
@@ -14,13 +14,7 @@
  * limitations under the License.
  */
 
-import {
-  Component,
-  OnInit,
-  AfterViewInit,
-  ViewChild,
-  OnDestroy,
-} from '@angular/core';
+import { Component, AfterViewInit, ViewChild, OnDestroy } from '@angular/core';
 import { Project } from '../../shared/models/project.model';
 import {
   Feature,
@@ -33,7 +27,7 @@ import {
 } from '../../services/drawing-tools/drawing-tools.service';
 import { ProjectService } from '../../services/project/project.service';
 import { FeatureService } from '../../services/feature/feature.service';
-import { Observable, Subscription } from 'rxjs';
+import { combineLatest, Observable, Subscription } from 'rxjs';
 import { List } from 'immutable';
 import { getPinImageSource } from './ground-pin';
 import { NavigationService } from '../../services/router/router.service';
@@ -42,14 +36,16 @@ import { GoogleMap } from '@angular/google-maps';
 // To make ESLint happy:
 /*global google*/
 
+const normalIconScale = 30;
+const enlargedIconScale = 50;
+
 @Component({
   selector: 'ground-map',
   templateUrl: './map.component.html',
   styleUrls: ['./map.component.css'],
 })
-export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
+export class MapComponent implements AfterViewInit, OnDestroy {
   private subscription: Subscription = new Subscription();
-  focusedFeatureId: string | null = null;
   features$: Observable<List<Feature>>;
   activeProject$: Observable<Project>;
   private initialMapOptions: google.maps.MapOptions = {
@@ -60,6 +56,8 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
     streetViewControl: false,
     mapTypeId: google.maps.MapTypeId.HYBRID,
   };
+  private focusedMarker: google.maps.Marker | null = null;
+  private markers: google.maps.Marker[] = [];
   private crosshairCursorMapOptions: google.maps.MapOptions = {
     draggableCursor: 'crosshair',
   };
@@ -80,23 +78,16 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
     this.activeProject$ = this.projectService.getActiveProject$();
   }
 
-  ngOnInit() {
-    this.subscription.add(
-      this.navigationService.getFeatureId$().subscribe(id => {
-        this.focusedFeatureId = id;
-      })
-    );
-  }
-
   ngAfterViewInit() {
     this.subscription.add(
-      this.features$.subscribe(features => this.onFeaturesUpdate(features))
-    );
-    this.subscription.add(
-      this.activeProject$.subscribe(project =>
-        this.onProjectActivation(project)
+      combineLatest([
+        this.activeProject$,
+        this.features$,
+      ]).subscribe(([project, features]) =>
+        this.onProjectAndFeaturesUpdate(project, features)
       )
     );
+
     this.subscription.add(
       this.drawingToolsService
         .getEditMode$()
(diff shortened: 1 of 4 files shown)


web-ng/src/app/components/map/map.component.ts
```diff
@@ -60,6 +57,8 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
     streetViewControl: false,
     mapTypeId: google.maps.MapTypeId.HYBRID,
   };
+  private focusedMarker: google.maps.Marker | null = null;
```
gino-m, line 60: Can we use undefined here instead? Then it would be:
`private focusedMarker?: google.maps.Marker;`

web-ng/src/app/components/map/map.component.ts
```diff
@@ -163,42 +138,109 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
     }
   }
 
-  onFeatureClick(featureId: string | null) {
-    this.navigationService.setFeatureId(featureId);
+  private deleteMarkers() {
```
gino-m, line 141: It seems this is only ever called together with `clearGoogleMapDataLayer()`. Can they be combined into one method? The combined method name might then be something like `clearMap()` to make it clearer what this is doing.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -163,42 +138,109 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        new firestore.GeoPoint(event.latLng.lat(), event.latLng.lng())
+      );
+      this.featureService.updatePoint(newFeature);
+    });
+    this.markers.push(marker);
   }
 
-  private clearGoogleMapDataLayer() {
-    this.map.data.forEach(f => this.map.data.remove(f));
+  private focusMarker(marker: google.maps.Marker | null) {
```
gino-m, line 200: Iirc we've been using "selected" and "deselected", can we use the same terms here for consistency?

Also, consider splitting this into two methods, "selectMarker" and "deselectMarker".

web-ng/src/app/components/map/map.component.ts
```diff
@@ -163,42 +138,109 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    if (marker === this.focusedMarker) {
+      return;
+    }
+    if (marker !== null) {
```
gino-m, line 204: I think `if (marker)` would work here and would be more idiomatic ts.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -163,42 +138,109 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      this.setIconSize(marker, enlargedIconScale);
+      marker.setDraggable(true);
+    }
+    if (this.focusedMarker !== null) {
```
gino-m, line 208: Same here.

web-ng/src/app/services/feature/feature.service.ts
```diff
@@ -93,8 +103,14 @@ export class FeatureService {
       layerId,
       new firestore.GeoPoint(lat, lng)
     );
-    return this.dataStore
-      .updateFeature(project.id, newFeature)
-      .then(() => this.selectFeature(newFeature.id));
+    await this.dataStore.updateFeature(project.id, newFeature);
+    return this.selectFeature(newFeature.id);
```
gino-m, line 107: `selectFeature` doesn't return a value, so this return isn't necessary.

web-ng/src/app/services/feature/feature.service.ts
```diff
@@ -93,8 +103,14 @@ export class FeatureService {
...
+  }
+
+  private updatePointInternal(project: Project, feature: Feature) {
```
gino-m, line 110: Let's make this method `async` for consistency with other new methods.

gino-m (changes requested):
Excellent! Just a few humble suggestions, hope they're helpful.

web-ng/src/app/components/map/map.component.ts
DaoyuT: I think null makes sense, why undefined?

web-ng/src/app/services/feature/feature.service.ts
DaoyuT: Oops these comes from IDE suggestion, I don't quite understand the best practices, so let me revert it for now.

web-ng/src/app/services/feature/feature.service.ts
DaoyuT: Same as above, reverted.

web-ng/src/app/components/map/map.component.ts
DaoyuT: Renamed. If we split into "selectMarker" and "deselectMarker". Then we have to 
1. remember to deselect previous selected marker when selecting marker
2. make sure we deselect before select because select will assign selectedMarker, then we won't know who was previously selected.
So these two are basically still coupled. So I'll keep it in one method.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -60,6 +57,8 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
     streetViewControl: false,
     mapTypeId: google.maps.MapTypeId.HYBRID,
   };
+  private selectedMarker: google.maps.Marker | null = null;
```
gino-m, line 60: If we have the option, let's use `selectedMarker?` instead. Then `| null = null` isn't required.

web-ng/src/app/services/feature/feature.service.ts
gino-m: I couldn't find the comment you're referring to here.. for my own edification, why can't this be async?

gino-m (changes requested):
A few questions/suggestions, should be ready to merge after those!

web-ng/src/app/components/map/map.component.ts
```diff
@@ -163,42 +137,114 @@ export class MapComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        width: size,
+        height: size,
+      },
+    } as google.maps.Icon;
+    marker.setIcon(newIcon);
+  }
+
+  private addGeoJsonToMap(feature: GeoJsonFeature) {
+    const addedFeatures = this.map.data.addGeoJson(
+      (feature as GeoJsonFeature).geoJson
```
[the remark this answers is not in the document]
DaoyuT: No it does not need, good catch

web-ng/src/app/services/feature/feature.service.ts
DaoyuT: Oh, nvm, this is not async because there is no `await` in this method.

web-ng/src/app/components/map/map.component.ts
DaoyuT: It does look cleaner, changed.

web-ng/src/app/services/feature/feature.service.ts
gino-m: Does this work?
```ts
  private async updatePointInternal(project: Project, feature: Feature) {
    if (project.layers.isEmpty()) {
      return;
    }
    await this.dataStore.updateFeature(project.id, feature);
  }
```

gino-m (changes requested):

gino-m (approved):
LGTM! Great work! 📍

PR #460, opened 2020-10-11
[Code health] Extract fields from Map inside FeaturePanelComponent instead of template

gino-m:
Also resolves snake-case named "field_entry`.


parulraheja98 (approved):
LGTM 🚢

PR #465, opened 2020-10-31
[Observation list] Show form fields in order

### LINKED ISSUE #461: [Observation list] Show form fields in order
The order of fields shown in the side panel appears to be random. They should be based on the `index` property of the field.
image

parulraheja98:
Also fixed the order of layer.

Closes #461

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -49,7 +49,7 @@ export class LayerDialogComponent implements OnDestroy {
   lang: string;
   layer?: Layer;
   layerName!: string;
-  layersSize!: number;
+  layerCount!: number;
   projectId?: string;
   subscription: Subscription = new Subscription();
   fieldTypes = FieldType;
@@ -63,7 +63,7 @@ export class LayerDialogComponent implements OnDestroy {
       projectId: string;
       layer?: Layer;
       createLayer: boolean;
-      layersSize: number;
+      layerCount: number;
     },
     private dialogRef: MatDialogRef<LayerDialogComponent>,
     private dataStoreService: DataStoreService,
@@ -76,7 +76,7 @@ export class LayerDialogComponent implements OnDestroy {
     // Disable closing on clicks outside of dialog.
     dialogRef.disableClose = true;
     this.fields = List<Field>();
-    this.init(data.projectId, data.createLayer, data.layersSize, data.layer);
+    this.init(data.projectId, data.createLayer, data.layerCount, data.layer);
   }
 
   addQuestion() {
@@ -122,7 +122,7 @@ export class LayerDialogComponent implements OnDestroy {
   init(
     projectId: string,
     createLayer: boolean,
-    layersSize: number,
+    layerCount: number,
     layer?: Layer
   ) {
     if (!createLayer && !layer) {
@@ -130,7 +130,7 @@ export class LayerDialogComponent implements OnDestroy {
     }
     this.projectId = projectId;
     this.layer = layer;
-    this.layersSize = layersSize;
+    this.layerCount = layerCount;
     this.layerName = this.layer?.name?.get(this.lang) || '';
     this.color = this.layer?.color || DEFAULT_LAYER_COLOR;
     if (!layer) {
@@ -170,7 +170,7 @@ export class LayerDialogComponent implements OnDestroy {
     const forms = this.layerService.createForm(formId, fields);
     const layer = new Layer(
       this.layer?.id || '',
-      /* index */ this.layersSize,
+      /* index */ this.layerCount,
       this.color,
       // TODO: Make layerName Map
       StringMap({ [this.lang]: this.layerName }),
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Component, Inject, NgZone, OnDestroy } from '@angular/core';
+import { Component, Inject, OnDestroy } from '@angular/core';
 import {
   MatDialogRef,
   MAT_DIALOG_DATA,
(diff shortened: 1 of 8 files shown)


web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -49,6 +49,7 @@ export class LayerDialogComponent implements OnDestroy {
   lang: string;
   layer?: Layer;
   layerName!: string;
+  layersSize!: number;
```
gino-m, line 52: Does this refer to the number of layers? If so, `layerCount` would be more idiomatic.

gino-m (changes requested):
Hi @parulraheja98! Sorry if I misunderstood, but this PR seems to be ensuring new layers are added to the end of the layers list; it doesn't appear to do anything with the fields shown in the edit observation form. It looks like we needed to do this anyway - if so, could you update the PR description and remove the reference GitHub issue?

gino-m (changes requested):
Please disregard previous comment; the change needed was so small I missed it. Nice job!

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: This comment is still valid. Thanks!

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -170,7 +170,7 @@ export class LayerDialogComponent implements OnDestroy {
     const forms = this.layerService.createForm(formId, fields);
     const layer = new Layer(
       this.layer?.id || '',
-      /* index */ this.layersSize,
+      /* index */ this.layerCount,
```
gino-m, line 173: Sorry I just thought of this now - what if number of layers has changed while the layer dialog is opened (e.g., a layer was added in another window/tab)? In that case this index will be out of date. Can we instead assign this when the layer is added back to the project? Sorry about for late comment!

parulraheja98: Good Point 💯

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -199,8 +201,10 @@ export class LayerDialogComponent implements OnDestroy {
 
   onClose() {
     this.dialogRef.close();
-    // TODO: refactor this path into a custom router wrapper
-    return this.router.navigate([`p/${this.projectId}`]);
+    this.ngZone.run(() => {
```
gino-m, line 204: Why are Angular zones needed here?

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -165,12 +166,13 @@ export class LayerDialogComponent implements OnDestroy {
...
       return;
     }
+    const layerCount = await this.getLayerCount();
     const fields = this.layerService.convertFieldsListToMap(this.fields);
     const formId = this.form?.id;
     const forms = this.layerService.createForm(formId, fields);
     const layer = new Layer(
       this.layer?.id || '',
-      /* index */ this.layerCount,
+      /* index */ layerCount,
```
gino-m, line 175: I actually meant that we should set this value outside the component, in this case in the LayerService. That way there's no need to add a dependency on the active project.

Side nit: Please update the name `updateLayer` to `addOrUpdateLayer` if appropriate.

gino-m (changes requested):
Almost there! Thanks for the edits!

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
parulraheja98: I was having a warning here in the console Navigation triggered outside angular zone, that's why used zone here.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
gino-m: Thanks!

web-ng/src/app/services/layer/layer.service.ts
```diff
@@ -76,10 +81,10 @@ export class LayerService {
   }
 
   /**
-   * Updates layer of a project with a given layer value.
+   * Adds/Updates the layer of a project with a given layer value.
    */
-  updateLayer(projectId: string, layer: Layer): Promise<void> {
-    return this.dataStoreService.updateLayer(projectId, layer);
+  addOrUpdateLayer(projectId: string, layer: Layer): Promise<void> {
+    return this.dataStoreService.addOrUpdateLayer(projectId, layer);
```
gino-m, line 87: I would suggest making `getLayerCount()` private and updating the index here if the layer is a new one. We can tell if the layer is new by using a special value like '-1' until this method is called. This will abstract the details of how the layer is added and ensure the index is set right before saving.

gino-m (changes requested):

web-ng/src/app/services/layer/layer.service.ts
parulraheja98: Thanks , I really like this point 👍

web-ng/src/app/services/layer/layer.service.ts
```diff
@@ -83,7 +83,11 @@ export class LayerService {
   /**
    * Adds/Updates the layer of a project with a given layer value.
    */
-  addOrUpdateLayer(projectId: string, layer: Layer): Promise<void> {
+  async addOrUpdateLayer(projectId: string, layer: Layer): Promise<void> {
+    if (layer.index === -1) {
+      const layerCount = await this.getLayerCount();
+      (layer.index as number) = layerCount;
```
parulraheja98, line 89: I know you might ask why I am casting here, the reason is index is readonly property and I cannot assign it directly. Another approach could be creating new Layer instance which I don't prefer.

gino-m: Unfortunately this breaks immutability of the object; we rely on the assumption that our model objects are immutable, and this can lead to undesired consequences elsewhere. Instead, I would define a new method in Layer, `withIndex(number)`, that would return a copy of the Layer with the new index set. Thanks!!

gino-m (changes requested):
Thanks for the edits!

web-ng/src/app/services/layer/layer.service.ts
parulraheja98: appreciate it 😁

gino-m (approved):
LGTM!! 🎯

PR #483, opened 2020-11-04
[Code health] Simplify access to active project 

### LINKED ISSUE #296: [Project] Hide project page until loaded
- [ ] Consider using Resolvers to load project before route enables components.
- [ ] Replace async reference to project, ACLs, layer config, etc. with simpler sync reference

gino-m:
- [x] Add "project loading" interstitial
- [x] Only load project page once project is loaded
- [x] Replace async uses of project with sync ones, simplifying usage

Fixes #296.

### DIFF
--- a/web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
+++ b/web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
@@ -28,7 +28,7 @@ import { getPinImageSource } from '../map/ground-pin';
 import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
 
 @Component({
-  selector: 'drawing-tools',
+  selector: 'ground-drawing-tools',
   templateUrl: './drawing-tools.component.html',
   styleUrls: ['./drawing-tools.component.css'],
   changeDetection: ChangeDetectionStrategy.OnPush,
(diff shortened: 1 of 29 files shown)


web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
```diff
@@ -14,48 +14,53 @@
...
-    this.layers$ = projectService.getActiveProject$().pipe(
-      map(project => {
-        this.selectedLayerId = project.layers.keySeq().first();
-        this.drawingToolsService.setSelectedLayerId(this.selectedLayerId);
-        return List(project.layers.valueSeq().toArray()).sortBy(l => l.index);
-      })
-    );
+  }
+
+  ngOnChanges(): void {
```
gino-m, line 58: @DaoyuT @parulraheja98 @gauravchalana1 @gyarill I tried using `@Input` and `ngOnChanges` to simplifying making the application reactive to changes in the project config, but this approach makes me sad. The reason it makes me sad is that on some cases we still need the activeProject$ stream (e.g, when we're merging with feature or observation streams), and in other we don't want to react to changes in the project config (e.g., when editing an observation). This means we get up with at least three different ways of accessing the active project.

I personally prefer consistency and not needing to decide each time which approach to use, so considering rolling back to pure RxJS (no `@Input` properties, no cached `activeProject` in `projectService`).

Before I do this, do you have any thoughts/ideas/opinions about this?

gino-m:
Closed in favor of #487 and #488.

PR #498, opened 2020-11-10
Update readme to include offline base map sources

parulraheja98:


gino-m (approved):
Thanks for adding this!

PR #500, opened 2020-11-10
Deselect marker and set all markers unclickable when entering edit mode.

### LINKED ISSUE #468: [Add polygon] Clicking add polygon button has no effect
Since this isn't implemented we can hide the button for now.

@DaoyuT Would you kindly take a look at this one?

### LINKED ISSUE #482: [Add point] Points on map are clickable while in "add" mode
Points and other features on map should be disabled while in "add point" mode to prevent UI from getting into an undefined state where side panel is opened/active while adding points.

DaoyuT:
Fixes #482, fixes #468.

Behavior now:
Click a marker, it gets bigger, routed to that feature.
Click `add point`, the selected marker returns to normal size, routed to project url with no fragment parameter.
Click `add point` again, exits edit mode. All markers back to clickable.

There is only `add point` button on the map, no `add polygon`.

### DIFF
--- a/web-ng/src/app/components/drawing-tools/drawing-tools.component.html
+++ b/web-ng/src/app/components/drawing-tools/drawing-tools.component.html
@@ -20,7 +20,7 @@ class="drawing-buttons"
   <mat-button-toggle value = "{{pointValue}}" (click)="onButtonClick()">
     add point
   </mat-button-toggle>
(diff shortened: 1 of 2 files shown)


gino-m (approved):
Nice! Thanks for fixing this!

PR #516, opened 2020-11-20
Zoom and Pan the map when marker is clicked

### LINKED ISSUE #490: [Map] Zoom and pan when point is clicked
Center viewport on point, and if map is at less than zoom level N, zoom map to level N.

DakshSaini:
**Description**

- Zoom and Pan the map when marker is clicked

Closes #490


web-ng/src/app/components/map/map.component.ts
```diff
@@ -178,6 +178,21 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
 
+  private centerViewportOnMarker(options: google.maps.MarkerOptions): void {
+    if (options.position) {
+      this.map.panTo(options.position);
+    }
+  }
+
+  private setMapZoomDefault(): void {
+    if (
+      this.initialMapOptions.zoom &&
```
gino-m, line 189: Iiuc this will zoom out the map to the original zoom level, whereas we would like to zoom in closer to the point. Can we add a separate constant or param for the preferred "zoomed in" level? Default to 13?

gino-m (changes requested):
Really excited to received this change! One nit/request. Thanks!

web-ng/src/app/components/map/map.component.ts
DakshSaini: Sorry for the late reply. This would only zoom in if the current zoom level is less than 3, it will never zoom out because if current zoom is greater than 3, then the condition will be false. If we need zoom to be 13 on click, then this if statement would not be useful as we can just set the zoom value. Please let me know if that makes sense. Thanks!

gino-m: Hi @DakshSaini ! Zoom level 3 is at the scale of an entire country. It's the default zoom level shown when the map is opened. When a point is clicked, we'd like the map to zoom in to approx. the scale of a house or building, which is much greater (approx. 13). That's why I was suggesting using a different constant to specify the "zoomed in" zoom level vs. the default "zoomed out" level (`this.initialMapOptions.zoom`). Does that make sense?

DakshSaini: @gino-m, thank you for the feedback and explanation, and I totally agree! I have now made the necessary changes.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -217,7 +215,7 @@ export class MapComponent implements AfterViewInit, OnDestroy {
       this.selectMarker(marker);
       this.navigationService.setFeatureId(feature.id);
       this.centerViewportOnMarker(options);
-      this.setMapZoomDefault();
+      this.setMapZoom(zoomedInLevel);
```
gino-m, line 218: Nit: There's not hint in this line that zoomedInLevel is a constant, and that setMapZoom will only have an effect if zoomed out (the naming "setMapZoom" makes it sound like a plain-old mutator method).

Also, "centerViewportOnMarker" should only be provided with what it needs to work (min. surface area) rather than the whole `options` object.

To make the code more self-describing I'd suggest merging centerViewportOnMarker and setMapZoom into a single class method called "panAndZoom(lat, lng)". That method can access the `zoomedInLevel` constant directly. In future PRs we should probably rename static constant to all UPPER CASE.

gino-m (changes requested):
One more suggestion.  Thanks for the edit and for your patience!

web-ng/src/app/components/map/map.component.ts
gino-m: Hi @DakshSaini! Would you be able to incorporate this small refactor into this PR? If not please let me know and we can do it as a fast-follow after merge. Thanks

DakshSaini: Hi @gino-m, my sincere apologies for the delay. I have made the required changes now.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -179,6 +180,18 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     });
   }
 
+  private panAndZoom(
+    position?: google.maps.LatLng | google.maps.LatLngLiteral
```
gino-m, line 184: I assume you did this because this is the type defined in MapOptions? For this method, can we instead accept required google.maps.LatLng?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -202,6 +215,7 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     marker.addListener('click', () => {
       this.selectMarker(marker);
       this.navigationService.setFeatureId(feature.id);
+      this.panAndZoom(options.position);
```
gino-m, line 218: ...and then here use `options.position! as google.maps.LatLng` instead? The code would fail with a runtime error if position were missing or a different type, but since we are defining the value just above  in the same function, this can never happen.

gino-m (changes requested):
One more nit, otherwise ready to 🚢 !

web-ng/src/app/components/map/map.component.ts
gino-m: No worries at all, happy to have your contribution!

web-ng/src/app/components/map/map.component.ts
gino-m: Thanks! Gentle reminder to add the cast and to change the method to accept only `google.maps.LatLng`!

gino-m (approved):
Nice work! 🚢 !!

web-ng/src/app/components/map/map.component.ts
DakshSaini: Oops!, just noticed it, sorry pushed wrong stash earlier. Fixed now.

PR #521, opened 2020-11-23
Improve "add a point" workflow.

DaoyuT:
New "Add a point" button looks like:
Screen Shot 2020-11-23 at 2 18 17 PM
Hover on the button:
Screen Shot 2020-11-23 at 2 18 30 PM
The button is checked:
Screen Shot 2020-11-23 at 2 18 42 PM
The cancel button:
Screen Shot 2020-11-23 at 2 19 44 PM

Both clicking on the cancel button and clicking on the map to add a point exits the edit mode now.


web-ng/src/app/components/drawing-tools/drawing-tools.component.css
```diff
@@ -27,3 +27,15 @@
   background-color: white;
   transform: translate(-50%, 0);
 }
+
+.add-point-button {
+  background-color: #202225;
```
gino-m, line 32: Can these colors be refactored into styles/_variables.scss instead of being declared here?

web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
```diff
@@ -33,16 +39,27 @@ import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
...
+export class DrawingToolsComponent implements OnInit, OnDestroy {
+  private subscription: Subscription = new Subscription();
   pointValue = 'point';
   polygonValue = 'polygon';
   selectedValue = '';
   private lastSelectedValue = '';
   selectedLayerId = '';
   readonly layers$: Observable<List<Layer>>;
   readonly lang: string;
+  private black = '#202225';
```
gino-m, line 51: If we can't get this from styles, perhaps define it as at file level as `readonly const BLACK = ..;`. Same for GREEN.

gino-m (changes requested):
Nicely done! 🗺️ Small things, otherwise LGTM.

web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
DaoyuT: Done.

web-ng/src/app/components/drawing-tools/drawing-tools.component.css
DaoyuT: Good idea, extracted as css variables.

gino-m (approved):
🚀 LGTM! 🚢

PR #524, opened 2020-11-28
[Layer dialog] Suppress validation of form editor when user does not interact with it

### LINKED ISSUE #496: [Layer dialog] No fields should be required in question if nothing is filled in
Otherwise we can't add layers without forms

parulraheja98:
- This will allow layer to be saved if label is empty and field count is 1.
- It will show validation error if label is empty and field count is > 1.

Closes #496


gino-m (changes requested):
Hi @parulraheja98! Do I understand correctly that this PR makes it so that the first question is not required if a second question is not added? If so that's not the desired behavior; the behavior we're looking for is that the "template question"(the empty question presented when the layer dialog is first opened) is required only if the user hasn't interacted with it by filling in the question name, and options, etc.

Rather than turning off validation at the form level, is it possible to just ignore validation on save if the template question is not modified (i.e. empty)?

parulraheja98:
@gino-m, Can I get a re-review?

Screen Shot 2020-12-01 at 3 28 37 PM

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -196,6 +196,18 @@ export class LayerDialogComponent implements OnDestroy {
     }
   }
 
+  private isFieldOptionsValid(formFieldEditor: FormFieldEditorComponent) {
+    if (!formFieldEditor.optionEditors) {
+      return;
```
gino-m, line 201: Does `undefined` have special meaning? If not this should return true or false.

gino-m (changes requested):
A few nits, otherwise LG. Confirmed WAI locally. 🚀

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -100,7 +107,12 @@ export class FormFieldEditorComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   private validateLabel() {
-    if (this.label || (!this.label && this.fieldCount === 1)) {
+    if (
+      this.label ||
+      (!this.label &&
```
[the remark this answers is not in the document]
parulraheja98: Done 👍

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -196,6 +196,18 @@ export class LayerDialogComponent implements OnDestroy {
     }
   }
 
+  private isFieldOptionsValid(formFieldEditor: FormFieldEditorComponent) {
```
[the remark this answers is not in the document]
parulraheja98: this function is needed as control is not part of form group and exists as separated component.

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
parulraheja98: Done

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -106,16 +111,16 @@ export class FormFieldEditorComponent implements OnInit, OnChanges, OnDestroy {
...
-    ) {
-      return null;
-    }
-    return { labelInvalid: true };
+  private validateLabel(control: FormControl) {
+    return this.isFormEmpty() ? null : Validators.required(control);
+  }
+
+  private isFormEmpty(): boolean {
+    return (
```
gino-m, line 119: "()" are not necessary here, please remove.

gino-m (changes requested):
LGTM! One nit, otherwise ready to 🚢 !

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
parulraheja98: Unfortunately I tried this but this will throw error on npm run check and I didn't do this intentionally it is done by prettier.

gino-m: Interesting, I've never seen that before. Thanks for letting me know, I'll look into it!

gino-m (approved):
Great work, rock 'n' roll! 🚀

PR #530, opened 2020-11-30
Clean up and fix unit test failures and warnings

gino-m:
This fixes tests failing locally when run with `ng test`. Several unit tests were trying to use actual Angularfire libs instead of mocks; replaced with mocks where appropriate.

@gyarill PTAL?


gino-m:
@parulraheja98 @gauravchalana1 @DaoyuT FYI.

gyarill (dismissed):

gyarill (approved):

PR #533, opened 2020-12-03
[Import CSV] Implement importing features from CSV file

### LINKED ISSUE #252: [Top FR] Import point features from CSV
Allow features to be added to a layer from CSV. The simplest way to do this would be do add an overflow menu option next to layers in the side panel, "Import points". The CSV would then just need to contain two columns, lat and lng.

gino-m:
Click "Import CSV" from the layer overflow menu (three dots next to layer names in the right side panel):

image

@gyarill PTAL? Adding @Tiyarah who worked on the backend component as FYI.

Closes #252.


gyarill (approved):
Sorry for the delay Gino. I was just trying too hard to find something to comment on...

Good job:)

Issue #320, opened 2020-06-05
[Top FR] Export points and observations to CSV


gino-m:
For import into Earth Engine, TensorFlow, and other platforms. Involves updating existing script to work with new db format.

PR #537, opened 2020-12-14
[Export CSV] Add export CSV UI

gino-m:
Includes related change to Observation model and services.

Contributes to #320. 

Depends on #536.

### DIFF
--- a/web-ng/src/app/services/observation/observation.service.ts
+++ b/web-ng/src/app/services/observation/observation.service.ts
@@ -87,7 +87,7 @@ export class ObservationService {
   }
 
   createNewObservation(
-    user: User | null | undefined,
+    user: User,
     project: Project,
     feature: Feature
   ): Observation | LoadingState {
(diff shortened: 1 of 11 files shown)


web-ng/src/app/services/observation/observation.service.ts
```diff
@@ -98,9 +98,7 @@ export class ObservationService {
       .getLayer(feature.layerId)!
```
gyarill, line 98: don't we need a check now user != ANONYMOUS_USER or are you planning to allow ANONYMOUS_USER to leave observation?

gyarill (changes requested):

web-ng/src/app/services/observation/observation.service.ts
gino-m: As we currently have it configured, Firebase will refuse unauthenticated users, so the only time data can be submitted anonymously is to the local emulator setup, which is WAI. Please cmiiw!

gyarill: well i just wondered what the user would see it if happens?:P

gyarill (approved):

web-ng/src/app/services/observation/observation.service.ts
gino-m: If the user is able to access this UI at all then there's something wrong, since we currently require authentication to see the basic map UI. But even if they can access the map UI without logging in, Firebase won't allow them to save data to the db, and an error will be written to the console. In general, we should do better than fail silently, but in this case reaching that point would be impossible.

gino-m:
Thanks for reviewing!

Issue #519, opened 2020-11-22
[Map] All points flicker and there is a delay when a feature is updated remotely


gino-m:
Whenever a feature is updated in the remote db (including due to a user action in the UI), all points on the map flicker momentarily. This is likely because we remove all map features and add them back on each change. Instead, the UI component should diff the latest state with the current one and only update state that has changed.

@DaoyuT could you PTAL?

PR #539, opened 2020-12-16
[Map] Only apply the diff when there are features added/removed

DaoyuT:
Only apply the diff when there are features added/removed, instead of clear all and then add all.
This should solve the issue #519 (All points flicker and there is a delay when a feature is updated remotely).


web-ng/src/app/components/map/map.component.ts
```diff
@@ -81,13 +83,39 @@ export class MapComponent implements AfterViewInit, OnDestroy {
   }
 
   ngAfterViewInit() {
+    const dummyProject: Project = new Project(
+      '',
+      StringMap({ en: 'title' }),
+      StringMap({ en: 'description' }),
+      Map({}),
+      Map({})
+    );
```
gino-m, line 92: Instead of creating placeholder project and features here for initial state, can we use `switchMap` to chain the streams as we do elsewhere? That might look something like:
```js
this.activeProject$
  .pipe(switchMap(project => this.features$.subscribe(features => doStuff...)));
```

web-ng/src/app/components/map/map.component.ts
```diff
@@ -81,13 +83,39 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
-        this.features$,
-      ]).subscribe(([project, features]) =>
-        this.onProjectAndFeaturesUpdate(project, features)
-      )
+        this.activeProject$.pipe(startWith(dummyProject)),
+        this.features$.pipe(startWith(List<Feature>())),
+      ])
+        .pipe(pairwise())
+        .pipe(
+          map(([[oldProject, oldFeatures], [newProject, newFeatures]]) => {
```
gino-m, line 100: Rather than look at the previous value received in the stream, reactive UI elements should  compare their current UI state with the state from the incoming event. The idea is that events in the stream are idempotent and independent, which affords us some guarantees that make the code foolproof and debuggable. Would it be possible to compare against the features in the map instead of the previous value in the stream?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -101,9 +129,13 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
 
-  onProjectAndFeaturesUpdate(project: Project, features: List<Feature>) {
-    this.clearMap();
-    this.addMarkersAndGeoJsonsToMap(project, features);
+  onProjectAndFeaturesUpdate(
+    project: Project,
+    featuresToAdd: List<Feature>,
+    feautresToRemove: List<Feature>
+  ) {
+    this.removeMarkersAndGeoJsonsOnMap(feautresToRemove);
```
gino-m, line 137: Considering the above, instead of receiving the features to add and remove, we'd receive the latest List of features, and do a pairwise comparison, adding, removing, and updating as we go.

gino-m (changes requested):
Thanks for the change, @DaoyuT ! A few suggestions on design, impl looks great!

web-ng/src/app/components/map/map.component.ts
DaoyuT: Good point, now comparing based on the UI state(this.markers for location feature & this.map.data for geo json feature) and values from the stream.

web-ng/src/app/components/map/map.component.ts
DaoyuT: Done.

web-ng/src/app/components/map/map.component.ts
DaoyuT: I did this because I was using `pairwise`, removed since now not using `pairwise` anymore.

gino-m (approved):
🚀  Nice work @DaoyuT, 🚢 !

PR #544, opened 2020-12-20
[Auth] Make route auth protected

parulraheja98:


web-ng/src/app/services/auth/auth.guard.service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Injectable } from '@angular/core';
+import { Router } from '@angular/router';
+import { AuthService } from './auth.service';
+import { Observable, of } from 'rxjs';
+import { catchError, map } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AuthGuardService {
```
gino-m, line 25: Please rename class and file to AuthGuard and auth.guard.ts, respectively, to avoid confusion with AuthService.

web-ng/src/app/shared/models/user.model.ts
```diff
@@ -19,4 +19,5 @@ export interface User {
   email: string;
   displayName?: string;
   photoURL?: string;
+  uid?: string;
```
gino-m, line 22: Rather than add another identifier, you can add a boolean "isAuthenticated" which will be set to `true` when creating real `Users` and `false` in the `ANONYMOUS_USER` constant.

gino-m (changes requested):
A few initial comments. Thanks!

web-ng/src/app/services/auth/auth.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(private router: Router, private authService: AuthService) {}
+
+  canActivate(): Observable<boolean> {
+    return this.authService.getUser$().pipe(
+      map(response => {
+        if (response.isAuthenticated) {
+          return true;
+        }
+        return false;
+        // TODO: this.router.navigate(['/signin']);
```
gino-m, line 35: Please finish these TODOs. Be sure to assign `/signin` to a constant. Thanks!

web-ng/src/app/services/auth/auth.service.ts
```diff
@@ -49,7 +50,9 @@ export class AuthService {
           return of(null);
         }
       }),
-      map(user => user || ANONYMOUS_USER),
+      map(
+        user => (user && { ...user, isAuthenticated: true }) || ANONYMOUS_USER
```
gino-m, line 54: Rather than rebuild the user here, just set `isAuthenticated` when building the user in `FirebaseDataConverter`.

web-ng/src/app/shared/models/user.model.ts
```diff
@@ -19,4 +19,5 @@ export interface User {
   email: string;
   displayName?: string;
   photoURL?: string;
+  isAuthenticated?: boolean;
```
gino-m, line 22: Please remove `?` - this should never be `undefined`.

gino-m (changes requested):
Almost there!

gino-m (approved):
Nice work! Sorry i didn't realize we weren't calling the converter when deserializing Users, good catch!

PR #558, opened 2020-12-22
[Import CSV] Authenticate calls to import CSV Cloud Function

gino-m:


gino-m:
@parulraheja98 @gauravchalana1 @DaoyuT @DakshSaini @gyarill  This one up for grabs for review! Please reassign to yourself or ping this thread if you can review. Thanks!

parulraheja98:
@gino-m , LGTM

gino-m:
> @gino-m , LGTM

If it looks good please approve to unblock merge. Thanks!

parulraheja98 (approved):
SHIP IT 🚀

web-ng/src/app/services/auth/auth.service.ts
```diff
@@ -76,4 +76,24 @@ export class AuthService {
...
     return this.router.navigate(['/']);
   }
+
+  /**
+   * Returns a fresh JSON Web Token (JWT) used to identify the user to a
+   * Firebase service.
+   */
+  private async getIdToken(): Promise<string> {
+    const user = await this.afAuth.currentUser;
+    return await user!.getIdToken(/* forceRefresh */ true);
```
gino-m, line 86: Just tested on hosted Firebase and this is returning `error="invalid_token" error_description="The access token could not be verified"`. Trying to debug this but running out of leads. Please hold off on merging for now.

gino-m: Correction: the server is replying with this error in the "Authorization" response header.

gino-m: This also appears relevant:
https://firebase.google.com/docs/auth/admin/manage-cookies#verify_session_cookie_and_check_permissions

gino-m: And here:
https://firebase.google.com/docs/auth/web/auth-state-persistence#modifying_the_auth_state_persistence

gino-m:
Update:
My understanding is that if we want export CSV to be able to stream downloads (#586), we'll need to access the Cloud Function via browser hyperlink. Since there's no way to set auth headers in hyperlinks, the only way we can authenticate would via cookies. Getting session cookies working ([instructions](https://firebase.google.com/docs/auth/admin/manage-cookies)) should get us there. If we start using session cookies for auth, this PR becomes obsolete, so closing it for now.

PR #574, opened 2020-12-24
[Sign in] Prevent sign in page from showing if already signed in

### LINKED ISSUE #547: [Sign in] Block access to sign in page if already signed in
We should not allow authenticated users to access `/signin`. Instead, we can create an auth guard (`UnauthGuard`?) that will redirect to `/p/:new` for now if the user is already logged in.

parulraheja98:
Closes #547


web-ng/src/app/services/auth/unauth.guard.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Observable, of } from 'rxjs';
+import { catchError, map } from 'rxjs/operators';
+import { environment } from '../../../environments/environment';
+
+const DEFAULT_ROUTE = ['p', ':new'];
+
+@Injectable({
+  providedIn: 'root',
+})
+export class UnauthGuard {
```
gino-m, line 28: I did some searching and found that `canActivate()` can take an  [`ActivatedRouteSnapshot`](https://angular.io/api/router/ActivatedRouteSnapshot). In that case, instead of creating a separate guard class as per my original brainstorm (sorry!), can we just instead just update AuthGuard to react differently when trying to activate `/signin`? 

In the future if we have more routes we can make these behaviors configurable by passing `data` as part of the Routers in `RoutingModule`; for now hard-coding `/signin` in the AuthGuard should be ok.

gino-m (approved):
One suggestion, thanks!

Please be sure to name PRs for what they do, for ex:
"[Sign in] Prevent sign in page from showing if already signed in"

gino-m (changes requested):
Undo accidental approval

web-ng/src/app/services/auth/auth.guard.ts
```diff
@@ -14,36 +14,55 @@
...
     return this.authService.getUser$().pipe(
-      map(response => {
-        if (environment.useEmulators) {
-          return true;
+      map(user => {
+        if (state.url === '/signin') {
+          return this.handleSignin(user);
         }
-        if (response.isAuthenticated) {
+        if (this.isAuthenticated(user)) {
```
gino-m, line 44: I think this block can be made more self-describing; ideally the code would read more like the rules it implements. 

In practice, this might look something like this:
```ts
  canActivate(
    route: ActivatedRouteSnapshot,
    state: RouterStateSnapshot
  ): Observable<boolean> {
    return this.authService.getUser$().pipe(
      map(user => this.canUserActivate(user, route)),
      catchError(() => of(false))
    ).do(ok => if (!ok) this.router.navigate([AuthService.SIGN_IN_URL]); );
  }

  canUserActivate(user, route): boolean {
   {
      // User can access all routes when running local emulator.
      if (environment.useEmulators) {
         return true;
      }
      // Special rule for sign in URL - only allow access if *not* authenticated.
      if (route.something === AuthService.SIGN_IN_URL) {
        return !user.isAuthenticated;
      }
      // All other routers require sign in.
      return user.isAuthenticated;
  }

File: web-ng/src/app/services/auth/auth.guard.ts [hunk 3]
```diff
@@ -14,36 +14,55 @@
...
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ): Observable<boolean> {
     return this.authService.getUser$().pipe(
-      map(response => {
-        if (environment.useEmulators) {
-          return true;
+      map(user => {
+        if (state.url === '/signin') {
```
gino-m, line 41: Will `url` also return any URL fragment (#anchor) or query (?a=123) params? If so this won't work if the user adds them to the URL. Not sure, but I suspect you may need to use `ActivatedRouteSnapshot` instead.

web-ng/src/app/services/auth/auth.guard.ts
```diff
@@ -14,36 +14,55 @@
...
+          return this.handleSignin(user);
```
gino-m, line 42: Good effort on refactoring this logic into multiple functions; in this case, the added indirection makes the logic harder to follow because a) the method names `handleSigin` and `isAuthenticated` don't accurately reflect what the methods do and b) the logic being implemented is simpler than the refactor would lead on.

gino-m (changes requested):
Thanks for the fixes - a few more suggestions to make the code even more readable. HTH!

web-ng/src/app/services/auth/auth.guard.ts
parulraheja98: thank you 🥇

web-ng/src/app/services/auth/auth.guard.ts
parulraheja98: I am using string includes which will check for substring now.

gino-m (approved):
@parulraheja98 I realized early this morning that the example I sent you wasn't correct since it didn't consider redirect from sign-in page, but I see you already fixed it. Great work! A+++!

PR #575, opened 2020-12-24
[Auth] UI breaks when user signs out in another tab

### LINKED ISSUE #568: [Auth] UI breaks when user signs out in another tab
For example, clicking a point doesn't show the feature panel.

To fix this, we should subscribe to `authService.isAuthenticated$()` in `MainPageContainerComponent` and redirect if false.

@parulraheja98 could you PTAL?

parulraheja98:
Closes #568

### DIFF
--- a/web-ng/src/app/components/main-page/main-page.component.ts
+++ b/web-ng/src/app/components/main-page/main-page.component.ts
@@ -24,6 +24,9 @@ import { ProjectService } from '../../services/project/project.service';
 import { ObservationService } from '../../services/observation/observation.service';
 import { take } from 'rxjs/operators';
 import { NavigationService } from '../../services/router/router.service';
+import { AuthService } from '../../services/auth/auth.service';
+import { AuthGuard } from '../../services/auth/auth.guard';
+import { Router } from '@angular/router';
 
 /**
  * Root component for main application page showing map, layers list, and
@@ -44,7 +47,9 @@ export class MainPageComponent implements OnInit {
     private projectService: ProjectService,
     private featureService: FeatureService,
     private observationService: ObservationService,
-    private dialog: MatDialog
+    private authService: AuthService,
+    private dialog: MatDialog,
+    private router: Router
   ) {
     // TODO: Make dynamic to support i18n.
     this.sideNavOpened = true;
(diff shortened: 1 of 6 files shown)


web-ng/src/app/components/main-page/main-page.component.ts
```diff
@@ -70,6 +75,14 @@ export class MainPageComponent implements OnInit {
         .getObservationId$()
         .subscribe(id => this.editObservation(id))
     );
+    // Redirect to sign in page if user is not authenticated.
+    this.subscription.add(
+      this.authService.isAuthenticated$().subscribe(isAuthenticated => {
+        if (!isAuthenticated) {
+          this.router.navigate([AuthGuard.SIGN_IN_URL]);
```
gino-m, line 82: Would it be possible to move the this constant into AuthService so we don't need to introduce a dep on AuthGuard in component? Ideally components will be entirely independent of lower level details and global effects like auth guards.

web-ng/src/app/services/auth/auth.service.ts
```diff
@@ -62,7 +63,9 @@ export class AuthService {
   }
 
   isAuthenticated$(): Observable<boolean> {
-    return this.getUser$().pipe(map(user => user.isAuthenticated));
+    return this.getUser$().pipe(
```
gino-m, line 66: As mentioned in a previous PR, burying the workaround override the expected behavior of a method like we're doing here will lead to "surprises" when debugging - it would be clearer of `isAuthenticated$()` always returned whether the user is actually authenticated. Callers can always disregard this value if `environment.useEmulators` is true.

gino-m (changes requested):
A few suggestions. Thanks!

parulraheja98:
@gino-m , Can you please let me know what test of build is failing here? My npm run check and ng test show no errors locally.

gino-m:
> @gino-m , Can you please let me know what test of build is failing here? My npm run check and ng test show no errors locally.

Hi @parulraheja98, it seems to be due to a failing in lint check:
```
/workspace/web-ng/src/app/components/observation-form/observation-form.component.ts
  250:56  error  Replace `⏎······Option⏎····>)?.first()·as·Option)` with `Option>)?.first()·as·Option)⏎······`  prettier/prettier
```

parulraheja98:
@gino-m  Ready for re-review

gino-m (approved):
Great fix! 🥳

PR #584, opened 2020-12-27
[Layer dialog] Fix default layer pin color

### LINKED ISSUE #551: [Layer dialog] Default layer color appears black instead of orange
When creating a new layer, the pin color shown in the style selector is black, but the layer is saved, it becomes orange. The pin shown on the dropdown should display the default color when the dialog is opened.

DakshSaini:
Closes #551

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
@@ -24,7 +24,7 @@
 <div class="default-style">
   <div class="default-style-tag">Default style</div>
   <app-edit-style-button
-    [markerColor]="layer!.color || '#ff9131'"
+    [markerColor]="layer!.color || defaultColor"
     (markerColorChange)="onMarkerColorChange($event)"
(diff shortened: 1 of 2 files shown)


web-ng/src/app/components/layer-dialog/layer-dialog.component.html
```diff
@@ -24,7 +24,7 @@
 <div class="default-style">
   <div class="default-style-tag">Default style</div>
   <app-edit-style-button
-    [markerColor]="layer!.color || '#000'"
+    [markerColor]="layer!.color || '#ff9131'"
```
gino-m, line 27: Would it be possible to read this from this constant instead?
https://github.com/google/ground-platform/blob/4e8c99e193b8f6c1679113bab41e72c6dca97565/web-ng/src/app/components/layer-dialog/layer-dialog.component.ts#L47

gino-m (changes requested):
Thanks for the fix! One nit.

gino-m (approved):
Nice work!

PR #597, opened 2021-01-06
[Create project] Stop modal from closing when error occurs creating project

DakshSaini:
Fixes onCreateProject function.

### DIFF
--- a/web-ng/src/app/components/title-dialog/title-dialog.component.ts
+++ b/web-ng/src/app/components/title-dialog/title-dialog.component.ts
@@ -37,14 +37,17 @@ export class TitleDialogComponent {
   }
 
   async onCreateProject() {
-    const projectId = await this.projectService.createProject(
-      this.projectTitleForm.get('title')?.value
-    );
+    let projectId = '';
     try {
-      this.navigationService.setProjectId(projectId);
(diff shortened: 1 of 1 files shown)


web-ng/src/app/components/title-dialog/title-dialog.component.ts
```diff
@@ -37,14 +37,17 @@ export class TitleDialogComponent {
...
     try {
-      this.navigationService.setProjectId(projectId);
+      projectId = await this.projectService.createProject(
+        this.projectTitleForm.get('title')?.value
+      );
     } catch (e) {
       console.warn('Project creation failed', e);
     }
-    this.matDialogRef.close();
+    if (projectId) {
```
gino-m, line 48: Can `createProject` return a falesy value? If not, can you remove the conditional and move this up inside the try block?

gino-m (changes requested):
Thanks for the fix!

web-ng/src/app/components/title-dialog/title-dialog.component.ts
DakshSaini: Fixed now!

gino-m (approved):

PR #600, opened 2021-01-11
[Project] Replace project URLs with more user friendly URLs

### LINKED ISSUE #590: [Project] Replace project URLs with more user friendly URLs
`/p` -> `/project`
`/p/:new` -> `/project/new`

DakshSaini:
Closes #590

### DIFF
--- a/web-ng/src/app/services/router/router.service.ts
+++ b/web-ng/src/app/services/router/router.service.ts
@@ -35,8 +35,8 @@ export class NavigationService {
   static readonly OBSERVATION_ID_NEW = 'new';
   static readonly PROJECT_ID_NEW = 'new';
   static readonly PROJECT_ID = 'projectId';
-  static readonly DEFAULT_PROJECT_ROUTE = 'project';
-  static readonly SIGN_IN_ROUTE = 'signin';
+  static readonly PROJECT_SEGMENT = 'project';
+  static readonly SIGN_IN_SEGMENT = 'signin';
 
   private static fragmentParamsToSideNavMode(params: HttpParams): SideNavMode {
     if (params.get(NavigationService.OBSERVATION_ID_FRAGMENT_PARAM)) {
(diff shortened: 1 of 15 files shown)


gino-m (changes requested):
Nice work! Would it be possible to roll a small refactor into this PR? Namely, it seems "project" and "new" are sprinkled across our codebase. As a quick fix, can we always refer to constants in the router instead? In a follow-up PR we can also replace calls to `this.router.navigate()` with methods in our `NavigationService`. Wdyt?

DakshSaini:
> Nice work! Would it be possible to roll a small refactor into this PR? Namely, it seems "project" and "new" are sprinkled across our codebase. As a quick fix, can we always refer to constants in the router instead? In a follow-up PR we can also replace calls to `this.router.navigate()` with methods in our `NavigationService`. Wdyt?

Sounds good, changes made. I'll be sure to create a follow-up PR.

web-ng/src/app/services/router/router.service.ts
```diff
@@ -33,6 +33,10 @@ export class NavigationService {
   private static readonly OBSERVATION_ID_FRAGMENT_PARAM = 'o';
   static readonly LAYER_ID_NEW = 'new';
   static readonly OBSERVATION_ID_NEW = 'new';
+  static readonly PROJECT_ID_NEW = 'new';
+  static readonly PROJECT_ID = 'projectId';
+  static readonly DEFAULT_PROJECT_ROUTE = 'project';
```
gino-m, line 38: Why is this the "default" project route? Shouldn't it be something like the "project URL base path segment"? (`PROJECT_SEGMENT`?)

web-ng/src/app/services/router/router.service.ts
```diff
@@ -33,6 +33,10 @@ export class NavigationService {
...
+  static readonly SIGN_IN_ROUTE = 'signin';
```
gino-m, line 39: Same here - this isn't a complete route, just a segment. Please rename accordingly.

gino-m (approved):
LGTM! Filed #605 for followup refactor. Thanks!!!

PR #601, opened 2021-01-11
[Layer list] Replace "Delete" menu option with "Delete layer"

### LINKED ISSUE #595: [Layer list] Replace "Delete" menu option with "Delete layer"

DakshSaini:
Closes #595


gino-m (approved):
Thanks for the fix!!!

PR #603, opened 2021-01-13
[Layer dialog] Add per-layer permissions to layer dialog

### LINKED ISSUE #474: [Layer dialog] Add per-layer permissions to layer dialog
For now users can only add points, so a simple checkbox  would suffice, selected/checked by default. Roughly:

```
Permissions                                     Default style
[x] Contributors may add points                 [Pin icon/button]
```

To represent this, we can add an `contributorsCanAdd: ['points']` to the layer object when selected.

This is needed to allow us to lock imported reference layers .

parulraheja98:
image

Closes #474


gino-m (approved):
💯 !!! LGTM!

PR #608, opened 2021-01-15
Add redirects for old short URLs

gino-m:
Follow up to #600.

@DakshSaini PTAL?

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -5,6 +5,18 @@
   },
   "hosting": {
     "public": "web-ng/dist/web-ng",
+    "redirects": [
+      {
+        "source": "/p/[:]new",
+        "destination": "/project/new",
+        "type": 301
+      },
+      {
+        "source": "/p/:id",
+        "destination": "/project/:id",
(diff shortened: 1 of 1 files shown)


firebase.json
```diff
@@ -5,6 +5,18 @@
   },
   "hosting": {
     "public": "web-ng/dist/web-ng",
+    "redirects": [
+      {
+        "source": "/p/[:]new",
```
DakshSaini, line 10: I believe we should use `"source": "/p{,/[:]new}"`. It worked for me. Reference - [redirects](https://firebase.google.com/docs/hosting/full-config#redirects)

DakshSaini (changes requested):

firebase.json
gino-m: Just tested it the original change again on our hosted instance and it works. `/p{,/[:]new}` would match both `/p` and `/p/:new`, which is not what we want here.

DakshSaini (approved):

PR #616, opened 2021-01-16
[ACLs] Hide "Add layer" and "Customize layer" based on ACLs 

### LINKED ISSUE #610: [ACLs] Hide "Add layer" and "Customize layer" based on ACLs
Hide "Add layer" button and "Customize layer" menu item if user is not a manager or the project owner.

A new method should be added to `AuthManager`, `canManageProject()` which returns `true` if the user is `owner` or `manager` of the project.

@parulraheja98 PTAL?

parulraheja98:
image
Closes #610


gino-m:
Awesome to see this! I'm out and about right now, but one initial comment - have you considered adding auth functions into Auth service instead of creating a new class for them? They seem to be within scope of that class, no?

parulraheja98:
@gino-m , moved the logic to auth service, Can I get re-review?

web-ng/src/app/services/auth/auth.service.ts
```diff
@@ -74,4 +76,21 @@ export class AuthService {
     await this.afAuth.signOut();
     return this.router.navigate(['/']);
   }
+
+  private async getUserEmail() {
+    const user = await this.user$.pipe(take(1)).toPromise();
```
gino-m, line 81: Since the UI is not reactive (i.e., it only checks the user permissions once when created)m would to make this code more readable can we instead keep the `currentUser` as a member of this class? We could subscribe to `user$` in the constructor, for example: `this.user$.subscribe(user => this.currentUser = user)`.

Then all of the auth methods can be sync instead of async.

web-ng/src/app/services/auth/auth.service.ts
```diff
@@ -74,4 +76,21 @@ export class AuthService {
...
+    return user?.email;
+  }
+
+  async canManageProject(acl: AclEntry[]) {
+    const userEmail = await this.getUserEmail();
+    const entry = acl.find(val => val.email === userEmail);
```
gino-m, line 87: I think this can be simplified as:
```ts
return !!acl.find(entry => entry.email === userEmail && entry.isManager());
``

File: web-ng/src/app/services/auth/auth.service.ts [hunk 3]
```diff
@@ -74,4 +76,21 @@ export class AuthService {
     await this.afAuth.signOut();
     return this.router.navigate(['/']);
   }
+
+  private async getUserEmail() {
+    const user = await this.user$.pipe(take(1)).toPromise();
+    return user?.email;
+  }
+
+  async canManageProject(acl: AclEntry[]) {
```
gino-m, line 85: Please add return types here and throughout

web-ng/src/app/services/project/project.service.ts
```diff
@@ -92,4 +93,13 @@ export class ProjectService {
     );
     return Promise.resolve(projectId);
   }
+
+  getProjectAcl(project: Project) {
```
gino-m, line 97: Please add JsDoc and define return type.

web-ng/src/app/shared/models/acl-entry.model.ts
```diff
@@ -0,0 +1,5 @@
+import { Role } from '../../shared/models/role.model';
+
+export class AclEntry {
+  constructor(readonly email: string, readonly role: Role) {}
```
gino-m, line 4: In this class you could define this:
```ts
static readonly MANAGER_ROLES = [Role.MANAGER, Role.OWNER];

public isManager(): boolean {
  return MANAGER_ROLES.includes(this.role);
}

**Review by gino-m — CHANGES_REQUESTED** [review]
Just a few small style/readability suggestions, otherwise looks great!

File: web-ng/src/app/shared/models/acl-entry.model.ts [hunk 6]
```diff
@@ -0,0 +1,11 @@
+import { Role } from '../../shared/models/role.model';
+
+export class AclEntry {
+  MANAGER_ROLES = [Role.MANAGER, Role.OWNER];
```
gino-m, line 4: Please mark `private static readonly` and access as AclEntry.MANAGER_ROLES.

web-ng/src/app/components/layer-list-item/layer-list-item.component.html
```diff
@@ -32,7 +31,11 @@ <h3 matLine>
         <mat-icon>more_vert</mat-icon>
       </button>
       <mat-menu #layerMenu="matMenu">
-        <button mat-menu-item (click)="onCustomizeLayer()">
+        <button
+          mat-menu-item
+          (click)="onCustomizeLayer()"
+          *ngIf="canViewCustomizeLayer"
```
gino-m, line 37: Sorry I forgot to mention in previous review - please `canViewCustomize... ` might be a bit confusing - perhaps simply `canCustomizeLayer` would be clearer here.

web-ng/src/app/components/layer-list/layer-list.component.html
```diff
@@ -23,7 +23,7 @@ <h3 matSubheader>Layers</h3>
       [layer]="entry"
     ></ground-layer-list-item>
   </div>
-  <div class="add-layer">
+  <div class="add-layer" *ngIf="canViewAddLayer">
```
gino-m, line 26: Same here - `canAddLayer`?

web-ng/src/app/services/auth/auth.service.ts
```diff
@@ -37,6 +38,8 @@ const ANONYMOUS_USER: User = {
 })
 export class AuthService {
   user$: Observable<User>;
+  currentUser?: User;
```
gino-m, line 41: Please make `currentUser` and `user$` private.

web-ng/src/app/components/layer-list/layer-list.component.ts
```diff
@@ -44,6 +48,16 @@ export class LayerListComponent {
           List(project.layers.valueSeq().toArray()).sortBy(l => l.index)
         )
       );
+    this.initLayerListPermission();
+  }
+
+  async initLayerListPermission(): Promise<void> {
```
gino-m, line 54: Since layer list component can never be shown without a project being loaded, can we do what we did with `currentUser` and expose `getCurrentProject()` from the `ProjectService`? Then this logic can all be synchronous. We `take(1)` here, so the current impl isn't reactive to remote changes to Project anyway.

For the edge case where permissions are changed while the user has the page open: We enforce ACLs server side, so when if they are in a UI which they have now been denied access to, we can show an error when the save will fails [TBD later].

Does that make sense?

gino-m (changes requested):
Thanks @parulraheja98 for all the changes, looks great! Hope you don't mind if we roll one more small refactor into this PR? Thanks again! 🥇

gino-m (approved):
Great work, Mr. PR!!!! 🥳 🚀 

🚢 it!

PR #618, opened 2021-01-22
[Router] Refactor calls to Router navigate() into NavigationService

### LINKED ISSUE #605: [Router] Refactor calls to Router navigate() into NavigationService
Classes should never call router.navigate() directly; instead, we should encapsulate these calls in NavigationService. For example:

```ts
  this.navigationService.signIn();
  this.navigationService.newProject();
  this.navigationService.selectProject(projectId);
  // etc..
```

@DakshSaini FYI

DakshSaini:
Closes #605

### DIFF
--- a/web-ng/src/app/services/router/router.service.ts
+++ b/web-ng/src/app/services/router/router.service.ts
@@ -156,7 +156,7 @@ export class NavigationService {
    * Navigate to the current URL, updating the observation id in the URL
    * fragment.
    */
-  setObservationId(id: string) {
+  setObservationId(id: string | null) {
     this.setFragmentParam(NavigationService.OBSERVATION_ID_FRAGMENT_PARAM, id);
   }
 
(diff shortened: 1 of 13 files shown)


web-ng/src/app/services/router/router.service.ts
```diff
@@ -169,11 +169,35 @@ export class NavigationService {
   }
 
   /**
-   * Navigate to the URL with new project id
+   * Navigate to the URL with new project id.
```
gino-m, line 172: Can we rename this to `selectProject` while we're at it, since it doesn't just set the id, I also removes other params in the process.

web-ng/src/app/services/router/router.service.ts
```diff
@@ -156,7 +156,7 @@ export class NavigationService {
    * Navigate to the current URL, updating the observation id in the URL
    * fragment.
    */
-  setObservationId(id: string) {
+  setObservationId(id: string | null) {
```
gino-m, line 159: Can we rename this `editObservation()` for clarity?

gino-m (changes requested):
This is great, a much needed improvement. A small request or two. Thanks!

gino-m (approved):
Nicely done! 🚀 🎸

PR #621, opened 2021-01-25
[ACLs] Hide "import CSV" & Share Button if user is not an owner or manager

### LINKED ISSUE #611: [ACLs] Disable "Share" button if not manager
Disable "Share" button if user is not a manager or the project owner.

A new method should be added to `AuthManager`, `boolean canManageProject()` which returns `true` if the user is `owner` or `manager` of the project.

@parulraheja98 One more?

### LINKED ISSUE #613: [ACLs] Hide "import CSV" on layers if user is not an owner or manager
Hide link if `AuthManager.canManageProject()` is false.

One more!

parulraheja98:
image
closes #613 , closes #611

### DIFF
--- a/web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
+++ b/web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
@@ -44,7 +44,6 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
   readonly lang: string;
   readonly layerListItemActionsType = LayerListItemActionsType;
   subscription: Subscription = new Subscription();
-  canCustomizeLayer = false;
 
   constructor(
     private sanitizer: DomSanitizer,
@@ -59,7 +58,6 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
     // TODO: Make dynamic to support i18n.
     this.lang = 'en';
     this.layerPinUrl = sanitizer.bypassSecurityTrustUrl(getPinImageSource());
-    this.initLayerItemPermission();
   }
 
   ngOnInit() {
@@ -143,13 +141,9 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
     );
   }
 
-  initLayerItemPermission(): void {
-    const project = this.projectService.getCurrentProject();
-    if (!project) {
-      return;
-    }
-    const acl = this.projectService.getProjectAcl(project);
-    this.canCustomizeLayer = this.authService.canManageProject(acl);
+  canShare(): boolean {
+    const acl = this.projectService.getCurrentProjectAcl();
+    return acl ? this.authService.canManageProject(acl) : false;
   }
 
   ngOnDestroy(): void {
--- a/web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
+++ b/web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
@@ -50,8 +50,7 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
     private importDialog: MatDialog,
     private dataStoreService: DataStoreService,
     private navigationService: NavigationService,
-    private projectService: ProjectService,
-    private authService: AuthService
+    readonly projectService: ProjectService
   ) {
     // TODO: Make dynamic to support i18n.
     this.lang = 'en';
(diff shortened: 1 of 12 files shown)


web-ng/src/app/components/project-header/project-header.component.ts
```diff
@@ -94,6 +97,15 @@ export class ProjectHeaderComponent implements OnInit, OnDestroy {
     });
   }
 
+  initPermission(): void {
+    const project = this.projectService.getCurrentProject();
+    if (!project) {
+      return;
+    }
+    const acl = this.projectService.getProjectAcl(project);
```
gino-m, line 105: Why not just have `getProjectAcl` return the ACL of the current project? It can return an empty ACL if no project is loaded.

web-ng/src/app/components/project-header/project-header.component.ts
```diff
@@ -94,6 +97,15 @@ export class ProjectHeaderComponent implements OnInit, OnDestroy {
...
+    this.canShare = this.authService.canManageProject(acl);
```
gino-m, line 106: We don't need to cache `canShare` as an internal state; just add a method `canShare()` that calls the auth service on the ACL. Then we won't need `initPermissions` either.

gino-m (changes requested):
Sorry for the delay! Great progress, just a few more suggestions, LG so far!

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
```diff
@@ -139,13 +137,9 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
...
 
-  initLayerItemPermission(): void {
-    const project = this.projectService.getCurrentProject();
-    if (!project) {
-      return;
-    }
-    const acl = this.projectService.getProjectAcl(project);
-    this.canCustomizeLayer = this.authService.canManageProject(acl);
+  canShare(): boolean {
+    const acl = this.projectService.getCurrentProjectAcl();
```
gino-m, line 141: Could you kindly let `getCurrentProjectAcl()` return `[]` instead of `undefined`? 

Also, we'll never need to call `canManageProject()` on an arbitrary ACL; it will always be on the current project, so you can move the call to `getCurrentProjectAcl()` into `canManageProject()`.

Then `canShare` will be just:

```
  return this.authService.canManageProject();
```

Wdyt?

web-ng/src/app/components/layer-list/layer-list.component.ts
```diff
@@ -47,16 +46,11 @@ export class LayerListComponent {
...
   }
 
-  initLayerListPermission(): void {
-    const project = this.projectService.getCurrentProject();
-    if (!project) {
-      return;
-    }
-    const acl = this.projectService.getProjectAcl(project);
-    this.canAddLayer = this.authService.canManageProject(acl);
+  canShare(): boolean {
```
gino-m, line 51: Since is this is shared logic (no pun intended), we can also move this method into NavigationService (eg as `canShareProject()`). See previous comment.

gino-m (changes requested):
A few more suggestions. Thanks for your hard work!

web-ng/src/app/components/layer-list/layer-list.component.ts
parulraheja98: Unfortunately, this results in circular dependency: 
Screen Shot 2021-02-02 at 10 42 53 PM

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
parulraheja98: I moved the logic of canManageProject to the project service as it will result in circular dependency if I keep the logic in auth service.

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
```diff
@@ -139,13 +137,9 @@ export class LayerListItemComponent implements OnInit, OnDestroy {
...
+    return acl ? this.authService.canManageProject(acl) : false;
```
gino-m: I think this can be simplified as:
```
return acl && this.authService.canManageProject(acl);
```

gino-m (changes requested):
I see it's not obvious how to refactor `canShare()` so let's leave it as-is for now. One comment, otherwise LGTM. Sorry again for the delay!

web-ng/src/app/components/layer-list-item/layer-list-item.component.ts
parulraheja98: @gino-m , in the latest revision canShare() method doesn't exist. Thank you :)

gino-m (approved):
LGTM. Thanks for the changes!

PR #629, opened 2021-02-22
[Export CSV] Give file meaningful name

### LINKED ISSUE #560: [Export CSV] Give file meaningful name
It currently downloads as "exportCsv.csv". Instead, we can use the name of the layer.

DakshSaini:
Closes #560 

Screen Shot 2021-02-22 at 8 37 01 PM


functions/export-csv.js
```diff
@@ -51,6 +51,10 @@ async function exportCsv(req, res) {
   });
 
   res.type("text/csv");
+  res.setHeader(
+    "Content-Disposition",
+    "attachment; filename=" + layer.name.en + ".csv"
```
gino-m, line 56: What happens if there are slashes or other special characters in the filename? Some ideas here:
https://stackoverflow.com/questions/8485027/javascript-url-safe-filename-safe-string

gino-m (changes requested):
Cool!! One comment, PTAL 🙏

functions/export-csv.js
DakshSaini: > What happens if there are slashes or other special characters in the filename? Some ideas here:
> https://stackoverflow.com/questions/8485027/javascript-url-safe-filename-safe-string

/ changes to _. Would we like to replace any special character with _?

gino-m: I think "-" would be a more modern approach. Wdyt?

DakshSaini: > I think "-" would be a more modern approach. Wdyt?

Sounds good, made the change, please take a look again :)

functions/export-csv.js
```diff
@@ -51,6 +52,12 @@ async function exportCsv(req, res) {
   });
 
   res.type("text/csv");
+  res.setHeader(
+    "Content-Disposition",
+    "attachment; filename=" +
+      layerName.replace(/[^a-z0-9]/gi, "-").toLowerCase() +
```
gino-m, line 58: LG! Please assign this "filenamization" operation to it's own function for better readability and testability. Thanks!

gino-m (changes requested):

functions/export-csv.js
```diff
@@ -32,6 +32,7 @@ async function exportCsv(req, res) {
 
   const layers = project.get("layers") || {};
   const layer = layers[layerId] || {};
+  const layerName = layer.name.en;
```
gino-m, line 35: This will fail if `name` is missing. Consider `layer.name && layer.name['en']`.

functions/export-csv.js
```diff
@@ -124,4 +129,11 @@ function getMultipleChoiceValues(id, element) {
   return option.code || option.label["en"] || "";
 }
 
+/**
+ * Returns the file name in lowercase (replacing any special characters with '-') for csv export
+ */
+function getFileName(layerName) {
+  return layerName.replace(/[^a-z0-9]/gi, "-").toLowerCase() + ".csv";
```
gino-m, line 136: It appears you're converting to lower case _after_ slugifying, which will replace all uppercase characters with `-`.

Also, we should probably specify a default in case layerName is empty. How about:
```js
layerName = layerName || 'ground-export';
fileBase = layerName.toLowerCase().replace(/[^a-z0-9]/gi, "-");
return `${layerName}.csv`;
```

gino-m (changes requested):
Sorry for the really long delay! Let's try to get LGTM from other team members to speed things up in the future!

functions/export-csv.js
DakshSaini: > It appears you're converting to lower case _after_ slugifying, which will replace all uppercase characters with `-`.
> 
> Also, we should probably specify a default in case layerName is empty. How about:
> 
> ```js
> layerName = layerName || 'ground-export';
> fileBase = layerName.toLowerCase().replace(/[^a-z0-9]/gi, "-");
> return `${layerName}.csv`;
> ```

Example seem to work fine as it will first convert everything to lowercase and then replace special chars with '-':

Screen Shot 2021-03-31 at 11 21 54 PM

functions/export-csv.js
```diff
@@ -133,7 +133,9 @@ function getMultipleChoiceValues(id, element) {
  * Returns the file name in lowercase (replacing any special characters with '-') for csv export
  */
 function getFileName(layerName) {
-  return layerName.replace(/[^a-z0-9]/gi, "-").toLowerCase() + ".csv";
+  layerName =
```
gino-m, line 136: This will fail with an error if `layerName` is null or undefined. Please see previous suggestion to fix :)

gino-m (changes requested):

DakshSaini:
@gino-m, can I get a re-review?

gino-m (approved):
🚀 !!

PR #634, opened 2021-03-29
[Reposition point] Update UI for confirming/cancelling move operations

DaoyuT:
Also fixed a bug where the close button in feature panel does not deselect the selected marker.


gino-m:
@DaoyuT We spoke, since this is stilL WIP can you please ping when ready for another look?

DaoyuT:
> @DaoyuT We spoke, since this is still WIP can you please ping when ready for another look?

Added a change to subscribe to selected feature/observation change and cancel the reposition.

web-ng/src/app/components/map/map.component.css
```diff
@@ -13,4 +13,48 @@
...
+
+.reposition-confirm-dialog {
+  position: absolute;
+  bottom: 2em;
+  left: 50%;
+  transform: translate(-50%, 0);
+  width: 407px;
+  height: 66px;
+  border-radius: 8px;
+  background: #188038;
```
gino-m, line 25: Can we use style vars and theme for colors in this sheet? See other scss for examples

web-ng/src/app/components/map/map.component.ts
```diff
@@ -207,16 +229,34 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     };
     const marker = new google.maps.Marker(options);
     marker.addListener('click', () => {
-      this.selectMarker(marker);
+      if (this.disableMapClicks) {
+        return;
+      }
       this.navigationService.selectFeature(feature.id);
     });
+    marker.addListener('dragstart', (event: google.maps.MouseEvent) => {
```
gino-m, line 237: Can we break these listeners out into their own functions for readability?

web-ng/src/app/components/map/map.component.css
DaoyuT: It seems this color is specific to this confirm dialog, extracted as a variable in the scss file.

web-ng/src/app/components/map/map.component.ts
DaoyuT: Done.

gino-m (approved):
LGTM!

PR #637, opened 2021-04-02
Clearer error messages when features are invalid

### LINKED ISSUE #628: Clearer error messages when features are invalid

parulraheja98:
Closes #628


web-ng/src/app/shared/converters/firebase-data-converter.ts
```diff
@@ -366,9 +366,10 @@ export class FirebaseDataConverter {
    * @param id the uuid of the project instance.
    * @param data the source data in a dictionary keyed by string.
    */
-  static toFeature(id: string, data: DocumentData): Feature | undefined {
+  static toFeature(id: string, data: DocumentData): Feature | void {
```
gino-m, line 369: I've never seen a method return `void` on an error state. Perhaps we should just throw Error instead with a meaningful message?

web-ng/src/app/shared/converters/firebase-data-converter.ts
```diff
@@ -366,9 +366,10 @@ export class FirebaseDataConverter {
...
     if (!this.isFeatureValid(data)) {
-      console.warn(`Invalid feature ${id} in remote data store ignored`);
+      const errors = this.getFeatureErrors(id, data);
```
gino-m, line 371: Why split error state out from conversion logic? I'd think we'd do them at the same time so that we don't need to manage drift between conversion logic and `getFeatureErrors`.

gino-m (changes requested):
Some initial design suggestions. Please DM me if you'd like to discuss in more detail.

parulraheja98:
@gino-m , PTAL :D

web-ng/src/app/shared/converters/firebase-data-converter.ts
```diff
@@ -367,8 +367,9 @@ export class FirebaseDataConverter {
    * @param data the source data in a dictionary keyed by string.
    */
   static toFeature(id: string, data: DocumentData): Feature | undefined {
-    if (!this.isFeatureValid(data)) {
-      console.warn(`Invalid feature ${id} in remote data store ignored`);
+    const errors = this.validateFeature(data);
```
gino-m, line 370: Hi @parulraheja98, I think extracting errors first into a data structure is probably overkill for this case. I was actually suggesting something different: just check the values as you extract them, and throw an exception when the value can't be converted. Would this work?

gino-m (changes requested):

web-ng/src/app/shared/converters/firebase-data-converter.ts
parulraheja98: I thought it would be much cleaner having an object with key value pair depicting the error :), but I will implement your approach :)

parulraheja98:
@gino-m , RFAL

parulraheja98:
@gino-m , Gentle ping

web-ng/src/app/shared/converters/firebase-data-converter.ts
gino-m: That assumes we want to keep track of and report multiple issues with the local db. Since they should be a rare edge case, the extra complexity of keeping track of error might not be warranted. The key recommendation is to keep validation and conversion logic together so that they don't drift or so we don't need to look in two places to update conversion logic.

gino-m (approved):
Nice work, thanks for the changes! 💯

PR #639, opened 2021-04-07
[Form designer] Add support for date and time fields

### LINKED ISSUE #638: [Edit layer] Add date and time as field types
When creating layer within project adding date and time as Form Input as well along Text, Multiple Choice and Photo.

HemanParbhakar:
In this branch date and time input in form is implemented.

Fixes #638 

@atalekar, @DaoyuT  PTAL?


DaoyuT (approved):
LGTM, can you fix the broken test? It seems you have some lint issues.

web-ng/src/app/shared/models/form/field.model.ts
```diff
@@ -39,6 +39,8 @@ export enum FieldType {
   TEXT = 1,
   MULTIPLE_CHOICE = 2,
   PHOTO = 3,
+  DATE = 4,
```
atalekar, line 42: Please update the documentation above.

atalekar, line 42: Please also ensure that both of these datatypes contain timezone info OR the underlying transport format is epoch time.

atalekar (approved):

web-ng/src/app/shared/models/form/field.model.ts
HemanParbhakar: ok

HemanParbhakar: for this need to make sure that the input coming from android form should be in epoch format not in string format will look into it.

atalekar: > for this need to make sure that the input coming from android form should be in epoch format not in string format will look into it.

if you use epoch format - you need to validate it is an uint64 - and the enum should have units specified (seconds, milliseconds etc as _S, _MS, etc)

HemanParbhakar: would be storing date and time in epoch format. for example (1617972920290)  https://developer.android.com/reference/java/lang/System#currentTimeMillis()

HemanParbhakar: will need to make some changes in web as well so that when the user views it is should converted to human readable format and when import csv is called it should be in epoch format.

gino-m: epoch is usually used for date + time in ms. In the case of `DATE`, we're only storing the date, not the time. 

Perhaps ISO date format string is more appropriate (`YYYY-MM-DD`)?  We don't need to bind our data model on how things are persisted in local/remote db, but worth considering that for convenience our data model is influenced by Firestore, which stores data in a quasi-JSON-like structure. In these structures, ISO 8601 appears to be the de facto standard:
https://docs.jsonata.org/date-time#json-and-iso-8601

In those case `YYYY-MM-DD` is acceptable to date-only fields.

HemanParbhakar:
> @googlebot I signed it!

HemanParbhakar:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgoogle%2Fground-platform%2Fpull%2F639) for more info**.

@googlebot I signed it!

web-ng/src/app/shared/models/form/field.model.ts
atalekar: hmm, some comments, 
a)  downstream it is VERY important to have timezone info. Manually mucking with timezones and trying to match satellite passover times with sensor readings and ground observations etc is too complicated to deal with if done in a non-standard way. 

b) `DATE` I am ok being `YYYY-MM-DD Z`  but that doesn't conform with ISO 8601 format.  https://stackoverflow.com/questions/52215961/how-can-i-write-an-iso-8601-date-with-a-timezone-but-no-time-component#:~:text=Officially%2C%20the%20ISO%208601%20spec,intervals%20%2D%20mentioned%20further%20below).

c) `TIME` that we need for our project isn't time of day - but rather `DATETIME` (with timezone). So again having epoch time here makes life easy. (our usecase is time of irrigation start and stop - sometime it could start at 11 pm to 1 am - not normal, but somebody could it) 

My recommendation is - The UI representation should be decided on the Android side / web frontend side, but storage should be one of epoch time or timezone aware datetime. Keeping both `DATE` and `TIME` as one of those 2 conveys information to UI how to interpret it for UI but it stores using the same format for both. Keeping things as epoch time - means it is not human readable in storage, but we significantly reduce opportunities of making mistakes by potentially not assigning timestamp etc and doing some computation with it. (also storage space is reduce - n character string vs uint64)

gino-m:
@HemanParbhakar Could you please resolve merge conflicts and failing tests/lint?

@googlebot rescan

gino-m (approved):

PR #649, opened 2021-04-29
[Import CSV] Show upload progress and success/error message

### LINKED ISSUE #528: [Import CSV] Show upload progress and success/error message

parulraheja98:
Relavant screenshots:

**Error of CSV upload:**

Screen Shot 2021-04-28 at 8 42 42 PM

**Success of CSV upload:**

Screen Shot 2021-04-28 at 8 43 09 PM

**Loading state:**
spinner-1

@jacobmclaws FYI

Closes #528


jacobmclaws:
Hey! @parulraheja98 Sure thing. Can I ask, when the CSV successfully uploads, what would the person see on the map? Usually some points would be added, correct? Layers/observations would probably appear in the side panel? Just asking so I know how much feedback we need to show the person when the upload is successful (if the screen noticeable changes then just a snackbar should work fine). 

Also, do we have any idea what some of the causes might be so that we could suggest any tips in the text when an upload fails?

jacobmclaws:
Here are my ideas so far. Let me know if this makes sense. 

**Loading state:** use the material [linear progress bar](https://kstatic.googleusercontent.com/files/b14128fc1118df8911bb9d5405a075c013d1f060e2fee2d6d78154f0d1143b41cbf1dc31d09de993bec3b18e078e8bcd4b226b245a394e70852d20d10a45f0e3)

Greens: primary 1EA362 ;  secondary 95CBB1
image

**Successful upload:** use the material snackbar
image

**Failed upload:** use the material banner with "Try again" and "Dismiss" buttons (I'd like to add more info about what might have gone wrong if we have any general tips we could add)
image

gino-m:
> Hey! @parulraheja98 Sure thing. Can I ask, when the CSV successfully uploads, what would the person see on the map? Usually some points would be added, correct? Layers/observations would probably appear in the side panel? Just asking so I know how much feedback we need to show the person when the upload is successful (if the screen noticeable changes then just a snackbar should work fine).

Yes, that's correct. When upload succeed, new points should appear on the map. There's currently a delay between the upload completing and points appearing, we have a bug out to fix that (can't find it now).

> Also, do we have any idea what some of the causes might be so that we could suggest any tips in the text when an upload fails?

The most likely cause for the upload to fail would be the file being in the incorrect format. There are many things that can go wrong, and it may be useful to leave space for the exact error message returned from the server.

Thanks for the mocks, this is a big help! @parulraheja98, wdyt?

gino-m:
Hi @jacobmclaws, @parulraheja98 and I just spoke. A few questions/comments:

Regarding the progress line:

* Will the user be able to interact with the app while uploading? If so, what if they upload a second file while an upload is still in progress?
* If the upload hangs or simply takes too long, should there be a way for the user to cancel the upload?

Regarding the error banner:

* For security reasons, I'm fairly sure we can't cause the local file to be uploaded again, so "Try again" might now be possible.
* There's no Material banner component in Angular, so we'd have to create it ourselves. Since "Try again" action won't be there anyway, would a snackbar or similar be appropriate here?

Thanks again for your support!

web-ng/src/app/components/import-dialog/import-dialog.component.html
```diff
@@ -13,21 +13,28 @@
...
-  <button mat-button mat-dialog-close>Cancel</button>
-  <button mat-button [mat-dialog-close]="true" color="primary">Upload</button>
-</mat-dialog-actions>
+
+<div *ngIf="isUploadingCSV; else defaultView">
+  <h2 mat-dialog-title>Uploading CSV</h2>
+  <mat-spinner mode="indeterminate" class="spinner"></mat-spinner>
+</div>
+<ng-template #defaultView>
+  <h2 mat-dialog-title>Import CSV</h2>
```
gino-m, line 22: "CSV"->"features"

web-ng/src/app/components/import-dialog/import-dialog.component.html
```diff
@@ -13,21 +13,28 @@
...
-      ></file-upload>
-    </div>
-  </form>
-</mat-dialog-content>
-<mat-dialog-actions align="end">
-  <button mat-button mat-dialog-close>Cancel</button>
-  <button mat-button [mat-dialog-close]="true" color="primary">Upload</button>
-</mat-dialog-actions>
+
+<div *ngIf="isUploadingCSV; else defaultView">
```
gino-m, line 17: `isUploadingCSV` -> `isImporting`

web-ng/src/app/components/import-dialog/import-dialog.component.ts
```diff
@@ -30,37 +31,44 @@ export class ImportDialogComponent {
...
-      files[0] as File
-    );
+    try {
+      this.isUploadingCSV = true;
+      await this.dataImportService.importCsv(
+        this.projectId,
+        this.layerId,
+        files[0] as File
+      );
+      this.notificationService.success('Csv import completed');
```
gino-m, line 65: "Import complete" and "Importing features failed"

gino-m (changes requested):
A few small changes, PTAL!

parulraheja98:
RFAL!

gino-m (approved):
💯 %! Thanks for the fix! This will provide a good base for more detailed progress indicators, and for my work on #696. 🥳

web-ng/src/styles.scss
```diff
@@ -42,6 +42,19 @@ button .mat-icon + label {
...
+snack-bar-container.success {
+  background-color: mat-color($primary);
+}
+
+snack-bar-container.error {
+  background-color: mat-color($warn);
+}
+
+snack-bar-container.notification {
+  color: white;
```
gino-m, line 54: @parulraheja98 Is there a `mat-color` we can use here instead? Perhaps submitted in a followup PR? :)

PR #653, opened 2021-05-05
Hide layers in add point dropdown if not allowed.

DaoyuT:
Hide add point button if no layers allowed.

### DIFF
--- a/web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
+++ b/web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
@@ -29,11 +29,14 @@ import { ProjectService } from '../../services/project/project.service';
 import { Observable, Subscription } from 'rxjs';
 import { tap } from 'rxjs/operators';
 import { Layer } from '../../shared/models/layer.model';
+import { Role } from '../../shared/models/role.model';
 import { List } from 'immutable';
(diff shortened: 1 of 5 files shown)


web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
```diff
@@ -81,11 +85,38 @@ export class DrawingToolsComponent implements OnInit, OnDestroy {
...
+        this.addableLayers(
+          project,
+          authService.getCurrentUser().email,
+          'points'
+        )
       )
     );
   }
 
+  private addableLayers(
```
gino-m, line 97: Can we move this into the `ProjectService`, as in #621, and recast as`canUserAddPointsToLayer(Layer)`? Then the `pipe` can `filter` instead of `map`,  the method name will be more self documenting, and won't rely on the string representation. Wdyt?

gino-m (changes requested):
Thanks for the change! Humble suggestion inline.

web-ng/src/app/components/drawing-tools/drawing-tools.component.ts
DaoyuT: Sounds good, changed.

gino-m (approved):
💯  🚀  👍

PR #660, opened 2021-05-10
Allow user to add Number fields.

DaoyuT:
New type Number in form editor.
Screen Shot 2021-05-10 at 12 15 32 PM

Displays number response in each observation card.
Screen Shot 2021-05-10 at 12 16 12 PM

Able to edit a number field when editing an observation.
Screen Shot 2021-05-10 at 12 16 25 PM


gino-m (approved):
Nice work, thanks for the change!

PR #691, opened 2021-05-18
[Form editor] Form editor allows questions with label containing only whitespace, e.g. " "

### LINKED ISSUE #674: [Form editor] Form editor doesn't trim whitespace from labels
**To Reproduce**
1. Add layer, give it a name
2. Enter a space as the first question label
3. Save

**Expected behavior**
The layer is saved with no form questions. As a general rule, user-defined labels entered into the form editor should always have leading and trailing whitespace `trim()`ed.

**Actual behavior**
A form question is created with label " ".

@parulraheja98 Could you PTAL?

parulraheja98:
Closes #674


gino-m (changes requested):
Where is the whitespace being trimmed from text fields? I'd assume it would be done in the layer dialog component, through which the user might enter such text (e.g. "   Some question with leading and trailing white space   ", not in the service. PTAL?

web-ng/src/app/services/layer/layer.service.ts
```diff
@@ -125,10 +125,25 @@ export class LayerService {
    */
   private static createFormMap(id: string, form: Form): Map<string, Form> {
     let forms = Map<string, Form>();
+    if (LayerService.isFormFieldEmpty(form)) {
```
gino-m, line 128: Logic that deals with forms doesn't belong in the service; this should be in the UI/form component.

gino-m: Also, we don't want to create a form if the field is empty, but looking at this code it here it appears we do create the form {} in the db.

gino-m (changes requested):

parulraheja98:
@gino-m , This PR handles delete the form value from firestore if user saves with empty label or delete all the fields.
Let's handle trimming of label in question and Multiple choice in separate PR :)

web-ng/src/app/components/layer-dialog/layer-dialog.component.ts
```diff
@@ -130,16 +130,7 @@ export class LayerDialogComponent implements OnDestroy {
...
-        FieldType.TEXT,
-        /* label= */
-        '',
-        /* required= */
-        false,
-        /* index= */
-        this.fields.size
-      );
-      this.fields = this.fields.push(newField);
+      this.createDefaultField();
```
gino-m, line 133: What's the difference between `createDefaultField()` and `addQuestion()`?

parulraheja98: Ah, I didn't see there was a method already!

gino-m (changes requested):
I tested this and unfortunately the trimming happens _as_ the user types, not on blur. The `trim()` probably needs to happen later.

gino-m:
@parulraheja98 If it's easier, we can also trim the leading/training spaces on save. Either would be fine.

parulraheja98:
@gino-m , This change adds trim on blur

gino-m (approved):
Thanks @parulraheja98 for this fix, and thank you for adding return types where they were missing. 💯 !

PR #692, opened 2021-05-18
[Form editor] Set focus to "Option" label field after clicking "Add option"

### LINKED ISSUE #663: [Form editor] Set focus to "Option" label field after clicking "Add option"
1. Add/Customize Layer
2. Click "Add option"

Focus should be set to "Option" field:

image

@parulraheja98, could you PTAL?

parulraheja98:
Closes #663


web-ng/src/app/components/option-editor/option-editor.component.ts
```diff
@@ -56,6 +59,9 @@ export class OptionEditorComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges() {
+    if (this.optionInput) {
```
gino-m, line 62: This causes the focus to be placed into the option label _on every change_ to anything on the option form (including Code). Instead, we just want to focus here when a new option is added.

gino-m (changes requested):
PTAL!

web-ng/src/app/components/option-editor/option-editor.component.ts
```diff
@@ -53,6 +56,7 @@ export class OptionEditorComponent implements OnInit, OnChanges {
         label: value.label,
       });
     });
+    this.focusOptionInput();
```
gino-m, line 59: Can we move this call to where new options are added?

gino-m: @parulraheja98 Gentle ping!

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -320,4 +321,14 @@ export class FormFieldEditorComponent implements OnInit, OnChanges, OnDestroy {
       editor.optionGroup.markAllAsTouched();
     });
   }
+
+  private focusNewOption(index: number): void {
+    // setTimeout is to added to wait for the option to load.
+    setTimeout(() => {
+      if (this.optionEditors) {
+        const option = this.optionEditors.find((_val, ind) => ind === index);
```
gino-m, line 329: Rather than passing an index and calling "find" can we just use `this.optionEditors.length + 1`?

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
```diff
@@ -320,4 +321,14 @@ export class FormFieldEditorComponent implements OnInit, OnChanges, OnDestroy {
       editor.optionGroup.markAllAsTouched();
     });
   }
+
+  private focusNewOption(index: number): void {
+    // setTimeout is to added to wait for the option to load.
```
gino-m, line 326: How does this work? I would think this would result in a race condition, but may just be my own ignorance.

gino-m (changes requested):
Tested locally and LGTM, just some questions first.

web-ng/src/app/components/form-field-editor/form-field-editor.component.ts
DaoyuT: If I am understanding your code correctly @Parul#0929, the reason you added the `setTimeout` is that if you don't set it, then it will execute too soon, so that it will focus on the second last item because angular thought that that was the last.  So you could force angular to re-render it before doing the focus. Here are some example how you can do it. https://stackoverflow.com/questions/35105374/how-to-force-a-components-re-rendering-in-angular-2 & https://stackoverflow.com/questions/43457149/angular-4-what-is-the-right-way-to-wait-for-operation.

DaoyuT (approved):
LGTM

gino-m (approved):
Sorry for slow follow-up. LGTM!

PR #700, opened 2021-05-18
form editor: dragging behaviour

### LINKED ISSUE #686: [Form editor] Dragging questions up or down causes other questions to move
**To Reproduce**
1. Create a form with questions numbered 1-10
2. Drag 10 up through the ranks

**Expected behavior**
Other questions (besides the one being dragged) remain where they are

**Actual behavior**
Questions sometimes invert position with the previous question

gyarill:
when dragging element insert it on the drop stop instead of switching position of it and element that it is being dropped on

closes #686


gino-m (approved):
LGTM, tested locally, works 💯 !

PR #701, opened 2021-05-18
[Layer dialog] Improve dimensions of layer dialog

### LINKED ISSUE #688: [Layer dialog] Improve dimensions of layer dialog
- [x] Add 24px margin so that dialog never bumps up against or overlaps edges of screen
- [x] Set max width to 1024px.
- [x] Ensure dialog header (title) and footer (Cancel/Save) do not scroll. The area with the default style, permissions , and form designer can scroll freely.

@DakshSaini PTAL?

DakshSaini:
Closes #688

### DIFF
--- a/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
+++ b/web-ng/src/app/components/layer-dialog/layer-dialog.component.html
@@ -21,76 +21,75 @@
   (focusOut)="setLayerName($event)"
 ></inline-editor>
 <hr class="horizontal-line" />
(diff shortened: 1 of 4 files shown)


gino-m (changes requested):
Thanks for this @DakshSaini ! It doesn't look like the dialog width is changing when I increase the screen size or the value in the stylesheet. Does the question field maybe need to be made to expand on resize (flex fill)?

gino-m (approved):
Nicely done. Tested locally, this is a big improvement! 🥇

PR #703, opened 2021-05-19
observation-form: make sure that we do not pop up with warning of unsaved changes right after saving them.

### LINKED ISSUE #675: [Add/Edit Observation] "You have unsaved changes" shows when saving observation
"You have unsaved changes in observation form, do you want to save them?" appears even when clicking "Save".

Hi @DaoyuT! Do you have a cycle to address this one?

gyarill:
We achieve it by marking form pristine if saving observation happens. That way we can leave check alone that loads warning when form is dirty because when we mark form pristine it is not dirty anymore

related code introduced in https://github.com/google/ground-platform/commit/6519a1f5b0750b07518a034ba098cb5a0e9588ae (btw. super cool idea!)

I tested by:
1. clicking "save" which does not generate confirm after this change
2. navigating back, clicking different feature, clicking "cancel"  which all still generates confirm popup

closes #675

### DIFF
--- a/web-ng/src/app/components/observation-form/observation-form.component.ts
+++ b/web-ng/src/app/components/observation-form/observation-form.component.ts
@@ -115,7 +115,10 @@ export class ObservationFormComponent {
         );
         this.dataStoreService
           .updateObservation(this.projectId!, updatedObservation)
-          .then(() => this.navigateToFeature())
+          .then(() => {
+            this.observationForm?.markAsPristine();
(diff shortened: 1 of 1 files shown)


gino-m (approved):
Nice work! Was the form was getting marked dirty because its values are overwritten asynchronously in `updateObservation`? If so this might be worth mentioning in a comment in a follow-up. Merging this PR, thanks! 🚀

Issue #705, opened 2021-05-19
[Layer list] Not all features deleted when deleting layer


gino-m:
The features are deleted by the client one at a time, so if the browser is closed or reloaded before complete, the operation will be interrupted. We likely will want move this operation to server-side an implement in a single transaction. In the interim, we can try to execute the deletion and project config update into a single atomic transaction.

See https://firebase.google.com/docs/firestore/manage-data/transactions#web-namespaced-api

PR #706, opened 2021-05-19
[Map] Hide features w/no layer

### LINKED ISSUE #695: [Map] Some imported GeoJSON geometries appear black instead of selected color
These don’t appear in Android. Were they features that were deleted?

gino-m:
Fixes #695, workaround for #705. @DaoyuT PTAL?


DaoyuT (approved):
LGTM

Issue #696, opened 2021-05-18
[Import features] Show number of features imported on completion


gino-m:
The server returns JSON containing a property "count" that can be used to determine how many features were successfully imported.

Issue #564, opened 2020-12-22
[Import CSV] Imported features takes a long time to appear


gino-m:
OTOO several minutes for 50 points. According to the [docs](https://firebase.google.com/docs/firestore/manage-data/transactions#batched-writes):

> For bulk data entry, use a server client library with parallelized individual writes. Batched writes perform better than serialized writes but not better than parallel writes.

Note that when performing the import, the Cloud Function returns immediately, but imported features don't appear in the db for several minutes.

PR #724, opened 2021-05-23
[Import features] Wait for inserts to finish before returning

gino-m:
@DaoyuT I ended up doing this because it was blocking me from being able to import GeoJSON so that I could test a fix on the Android client.

I also added messages to the response in `{error: '...'}` and count int `{count: ..}` while I was there.

May address #564 and #694. Contributes to #696. 

PTAL?

### DIFF
--- a/functions/import-csv.js
+++ b/functions/import-csv.js
@@ -65,7 +65,7 @@ async function importCsv(req, res) {
         inserts.push(insertRow(projectId, layerId, row));
       } catch (err) {
         console.error(err);
-        re.unpipe(busboy);
+        res.unpipe(busboy);
         await res
           .status(HttpStatus.BAD_REQUEST)
           .end(JSON.stringify({ error: err.message }));
--- a/functions/import-csv.js
+++ b/functions/import-csv.js
@@ -43,6 +43,10 @@ async function importCsv(req, res) {
   // Dictionary used to accumulate form field values, keyed by field name.
   const params = {};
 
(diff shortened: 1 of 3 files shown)


functions/import-csv.js
```diff
@@ -64,19 +62,23 @@ async function importCsv(req, res) {
 
     file.pipe(csvParser()).on("data", async (row) => {
       try {
-        await insertRow(projectId, layerId, row);
+        inserts.push(insertRow(projectId, layerId, row));
       } catch (err) {
         console.error(err);
-        await res.status(HttpStatus.BAD_REQUEST).end("{}");
-        // TODO(#525): Abort stream on error. How?
+        re.unpipe(busboy);
```
DaoyuT, line 68: Is this supposed to be `res.unpipe()` ? Same question for below

DaoyuT (dismissed):
LGTM

functions/import-csv.js
gino-m: Yikes, last minute typo! Fixed.

> Same question for below

What were you referring to? Don't see another instance of `re.`..

DaoyuT: Oh maybe I was looking at the same one and did not notice, LGTM

DaoyuT (approved):

functions/import-csv.js
gino-m: Thanks for reviewing!

PR #729, opened 2021-05-26
[Form editor] Question and field type not aligned

### LINKED ISSUE #708: [Form editor] Question and field type not aligned

DakshSaini:
Fixes #708

### DIFF
--- a/web-ng/src/app/components/form-field-editor/form-field-editor.component.scss
+++ b/web-ng/src/app/components/form-field-editor/form-field-editor.component.scss
@@ -19,6 +19,7 @@
 .field-type {
   margin-left: 30px;
   font-size: 12px;
+  margin-top: -3px;
 }
 
 .icon-field-type {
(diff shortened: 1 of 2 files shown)


web-ng/src/app/components/form-field-editor/form-field-editor.component.scss
```diff
@@ -19,6 +19,7 @@
 .field-type {
   margin-left: 30px;
   font-size: 12px;
+  margin-top: -3px;
```
gino-m, line 22: Negative margins are valid CSS, but they may make the layout harder to reason about. Is it possible to reduce the margin of one the question label divs instead?

gino-m (changes requested):
Comment/suggestion. Thanks for the fix!

gino-m:
> Comment/suggestion. Thanks for the fix!

@DakshSaini Gentle ping! 😃

gino-m (approved):
Thanks for the fix! 🔨

PR #693, opened 2021-05-18
[Form editor] Form questions should collapse on blur

DakshSaini:
Closes #681, closes #707.

PR #730, opened 2021-05-26
[Form editor] Only show drag handle on hover over question or option and on focus

### LINKED ISSUE #707: [Form editor] Only show drag handle on hover over question or option and on focus

DakshSaini:
Contributes to #707


gino-m (approved):
Cool, thanks! Could you also handle the case where the handle shows because the question is selected/expanded, perhaps in #693? Keeping #707 open to track that part of the change.

PR #763, opened 2021-06-16
[Map] Render polygons on map

HemanParbhakar:
- [x] Rendering polygon from remote db.
- [x] Make polygons clickable.
- [x]  On click opening observation window.

Screenshot 2021-06-17 at 1 41 29 AM


web-ng/src/app/components/map/map.component.ts
```diff
@@ -165,11 +175,21 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     });
   }
 
+  private removePolygonOnMap(newFeatureIds: List<string>) {
+    this.map.data.forEach(f => {
```
DaoyuT, line 179: I don't think you should use `this.map.data` here, unlike geojson, the polygons are not added to the `this.map.data`. They are more like the Marker objects, they are added to/removed from the map using `.setMap(this.map)` or `.setMap(null)`.

You can try adding/removing polygons in firebase and see if it reflects on the map.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -165,11 +175,21 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+    });
+  }
+
   private addFeaturesToMap(project: Project, features: List<Feature>) {
     const locationFeatureIds = this.markers.map(m => m.getTitle());
     const geoJsonFeatureIds: String[] = [];
+    const polygonFeatureIds: String[] = [];
     this.map.data.forEach(f => {
       geoJsonFeatureIds.push(f.getProperty('featureId'));
+      polygonFeatureIds.push(f.getProperty('featureId'));
```
DaoyuT, line 192: The `f` here is each geojson object that was imported by `this.map.data.addGeoJson(...)`, it has nothing to do with the polygons.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -294,6 +320,30 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
 
+  private addPolygonToMap(feature: PolygonFeature) {
+    const vertices: google.maps.LatLng[] = [];
+    feature.polygonVertices.map(vertex => {
+      vertices.push(new google.maps.LatLng(vertex.latitude, vertex.longitude));
+    });
+    // Construct the polygon.
+    const polygon = new google.maps.Polygon({
+      paths: vertices,
+      strokeColor: '#ff9131',
```
DaoyuT, line 331: The color should refer to the color of the layer right?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -294,6 +320,30 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+      strokeOpacity: 1,
+      strokeWeight: 4,
```
DaoyuT, line 333: Should we make it thicker when selected?

DaoyuT (changes requested):

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: replaced with project.layers.get(feature.layerId)?.color

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: updated in function
`polygon.addListener('click', () => {
      polygon.setOptions({ strokeWeight: 6 });
      this.panAndZoom(vertices[0]);
      this.zone.run(() => {
        this.navigationService.selectFeature(feature.id);
      });
    });
`

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: removed.

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: updated.

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: but in marker title is used as identification and in geoJson featureId is used as identification. But in polygon we don't have such thing to setup we only have paths.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -294,6 +326,34 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+    const polygon = Polygon(feature.id , new google.maps.Polygon({
+      paths: vertices,
+      clickable: true,
+      strokeColor: color,
+      strokeOpacity: 1,
+      strokeWeight: 4,
+      fillOpacity: 0,
+    }));
+    polygon.addListener('click', () => {
+      polygon.setOptions({ strokeWeight: 6 });
```
DaoyuT, line 345: We should extract this to an constant. 
Do a pull from upstream first. I merged a PR which introduced these two consts in this component:
```
const normalPolygonStrokeWeight = 3;
const enlargedPolygonStrokeWeight = 6;
```
You can reuse them.

DaoyuT (changes requested):
LGTM overall, a few small suggests.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -42,6 +49,11 @@ const normalIconScale = 30;
 const enlargedIconScale = 50;
 const zoomedInLevel = 13;
 
+function Polygon(id: string, polygon: google.maps.Polygon): google.maps.Polygon & { id: string } {
+  return Object.assign(polygon, { id });
+}
+type Polygon = google.maps.Polygon & { id: string };
```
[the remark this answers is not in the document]
HemanParbhakar: done

web-ng/src/app/components/map/map.component.ts
```diff
@@ -165,9 +180,20 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+        this.polygons.splice(i, 1);
+      }
+    }
+  }
+
   private addFeaturesToMap(project: Project, features: List<Feature>) {
     const locationFeatureIds = this.markers.map(m => m.getTitle());
     const geoJsonFeatureIds: String[] = [];
+    const polygonFeatureIds = this.polygons.map(m => m.id);
+    console.log(this.map.data);
```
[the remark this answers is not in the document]
HemanParbhakar: removed console.log(this.map.data);

web-ng/src/app/components/map/map.component.ts
```diff
@@ -294,6 +326,34 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+      clickable: true,
+      strokeColor: color,
+      strokeOpacity: 1,
+      strokeWeight: 4,
+      fillOpacity: 0,
+    }));
+    polygon.addListener('click', () => {
+      polygon.setOptions({ strokeWeight: 6 });
+      this.panAndZoom(vertices[0]);
+      this.zone.run(() => {
```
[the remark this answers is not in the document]
HemanParbhakar: removed this.zone.run

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: added 
`
const normalPolygonStrokeWeight = 3;
const enlargedPolygonStrokeWeight = 6;
`

gino-m:
@HemanParbhakar Please resolve conflicts when you get a chance.

DaoyuT (dismissed):
LGTM

web-ng/src/app/components/map/map.component.ts
```diff
@@ -46,9 +47,15 @@ import firebase from 'firebase/app';
...
 const normalIconScale = 30;
 const enlargedIconScale = 50;
+const zoomedInLevel = 13;
 const normalPolygonStrokeWeight = 3;
 const enlargedPolygonStrokeWeight = 6;
-const zoomedInLevel = 13;
+
+type Polygon = google.maps.Polygon & { id: string };
+
+function Polygon(id: string, polygon: google.maps.Polygon): Polygon {
```
gino-m, line 56: Can you just use `google.maps.Polygon` directly and call `.set('id', id)` and `get('id')` to retain the id?

web-ng/src/app/components/map/map.component.ts
```diff
@@ -69,6 +76,7 @@ export class MapComponent implements AfterViewInit, OnDestroy {
   };
   private selectedMarker?: google.maps.Marker;
   private markers: google.maps.Marker[] = [];
+  private polygons: Polygon[] = [];
```
gino-m, line 79: If you make this `Map<string, Polygon>` you won't need to iterate to find a polygon by id.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -370,6 +395,35 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     });
   }
 
+  private addPolygonToMap(project: Project, feature: PolygonFeature) {
+    const color = project.layers.get(feature.layerId)?.color;
+    const vertices: google.maps.LatLng[] = [];
+    feature.polygonVertices.map(vertex => {
+      vertices.push(new google.maps.LatLng(vertex.latitude, vertex.longitude));
+    });
+    // Construct the polygon.
```
gino-m, line 404: This comment doesn't add any clarity over what's in the code so it can be removed.

web-ng/src/app/components/map/map.component.ts
```diff
@@ -370,6 +395,35 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+        strokeWeight: normalPolygonStrokeWeight,
+        fillOpacity: 0,
+      })
+    );
+    polygon.addListener('click', () => {
+      polygon.setOptions({ strokeWeight: enlargedPolygonStrokeWeight });
+      this.panAndZoom(vertices[0]);
+      this.navigationService.selectFeature(feature.id);
+    });
+    if (this.map.googleMap !== undefined) {
```
gino-m, line 421: Is there a chance that this is called before the map is initialized? If so we should prevent this from happening, or polygons will just get dropped.

web-ng/src/app/shared/converters/firebase-data-converter.ts
```diff
@@ -571,11 +582,23 @@ export class FirebaseDataConverter {
     return Role[role].toLowerCase();
   }
 
+  private static isFeatureValid(data: DocumentData): boolean {
```
gino-m, line 585: Looks like this isn't used and can be deleted?

gino-m (changes requested):
A few comments.. thanks!

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: it does not have a option of set and get id https://developers.google.com/maps/documentation/javascript/reference/polygon#PolylineOptions

gino-m: [google.maps.Polygon](https://developers.google.com/maps/documentation/javascript/reference/polygon#Polygon) which you're using here extends [MVCObject](https://developers.google.com/maps/documentation/javascript/reference/event#MVCObject), which has `get()` and `set()` methods. Why can't these method be used?

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: this.map.googleMap is optional that why have added check it always called afterwards.

web-ng/src/app/shared/converters/firebase-data-converter.ts
HemanParbhakar: removed

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: removed

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: right implemented was following some other link.

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: Polygon changed to google.maps.Polygon

web-ng/src/app/components/map/map.component.ts
DaoyuT: Oh I gave the suggestion of extending Polygon with `id` field, didn't realize that MVCObject has `get()` & `set()` for storing the ID.  Sorry for the detour @HemanParbhakar.  This should look much cleaner, thanks @gino-m.

web-ng/src/app/components/map/map.component.ts
gino-m: > If you make this Map<string, Polygon> you won't need to iterate to find a polygon by id.

👆 

Could you also do this for `markers`?

web-ng/src/app/components/map/map.component.ts
gino-m: This method should never be called before the map is initialized, so `this.map.googleMap!` should be used instead with no check.

gino-m (changes requested):

web-ng/src/app/components/map/map.component.ts
```diff
@@ -343,9 +369,10 @@ export class MapComponent implements AfterViewInit, OnDestroy {
   }
 
   private selectMarkerWithFeatureId(selectedFeatureId: string | null) {
-    const markerToSelect = this.markers.find(
-      marker => marker.getTitle() === selectedFeatureId
-    );
+    const markerToSelect = this.markers.get(selectedFeatureId as string);
+    // const markerToSelect = this.markers.find(
+    //   marker => marker.getTitle() === selectedFeatureId
+    // );
```
gino-m, line 375: Delete commented out code

gino-m (changes requested):
Please also fix lint errors

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: removed.

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: updated

web-ng/src/app/components/map/map.component.ts
HemanParbhakar: updated

gino-m (approved):
LGTM! Thanks for the fixes.

PR #765, opened 2021-06-21
Combine back and delete button into dropdown menu in feature panel header.

### LINKED ISSUE #711: [Delete feature] Delete feature option gone
Introduced in #451. Can we add this back to the overflow menu? 

@jacobmclaws Did you have mocks for this?

DaoyuT:
Screenshot:
Screen Shot 2021-06-21 at 3 09 30 PM

Fixes #711.

### DIFF
--- a/web-ng/src/app/components/feature-panel-header/feature-panel-header.component.html
+++ b/web-ng/src/app/components/feature-panel-header/feature-panel-header.component.html
@@ -19,22 +19,17 @@ limitations under the License.
   <h3 matLine>
     {{ layer?.name?.get(lang) }}
   </h3>
-  <div [ngSwitch]="actionsType">
-    <div *ngSwitchCase="featureHeaderActionType.MENU">
-      <button mat-icon-button [matMenuTriggerFor]="featureMenu">
-        <mat-icon>more_vert</mat-icon>
-      </button>
-      <mat-menu #featureMenu="matMenu">
-        <button mat-menu-item (click)="onDeleteFeatureClick()">
-          <mat-icon>delete_forever</mat-icon>
-          <span>Delete</span>
-        </button>
-      </mat-menu>
-    </div>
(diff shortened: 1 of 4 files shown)


gino-m:
> Screenshot:
> Screen Shot 2021-06-21 at 3 09 30 PM

Would you like to put the "x" next to the overflow menu for now like in the example you showed me in Contacts? It seems odd to have the back/close option in an overflow. Wdyt?

gino-m:
> Screenshot:
> Screen Shot 2021-06-21 at 3 09 30 PM
> 
> Fixes #711.

@jacobmclaws Do you have any suggestions on how to present a header (eventually multiline) with an overflow and back/close CTA?

jacobmclaws:
I agree it's quite strange to have the button that takes you back in the dropdown. I would suggest a button above with some text. Something like this:
image

Not sure that "layers list" is the appropriate name for that side-panel view. Just "Back" probably also works in this context. I would suggest using our Ground green color for it to indicate it's part of the navigational system UI. Let me know if that makes sense. Thanks!

DaoyuT:
Done. Thanks @jacobmclaws for the suggestion.
Screen Shot 2021-07-05 at 6 16 42 PM

web-ng/src/app/components/feature-panel-header/feature-panel-header.component.html
```diff
@@ -14,27 +14,25 @@
 limitations under the License.
 -->
 
+<button mat-button
+        color="primary"
+        aria-label="back"
+        (click)="onCloseClick()">
+  <mat-icon>arrow_back</mat-icon>
+  <span>Back to layers list</span>
```
gino-m, line 22: Please replace with "Back" as suggested by @jacobmclaws

gino-m (approved):
One comment, otherwise LGTM.

gino-m (approved):
Made the change directly. LGTM!

Issue #769, opened 2021-06-23
[Export CSV] Allow column names to be customized on export


gino-m:
On "Export observations":

1. Scan all features being exported for keys in `properties`.
2. Show dialog with table of all property names and form questions on the left, with generated column names on the right (e.g., "Total cost (USD)" -> "total_cost_usd").
3. Duplicate column names should show in red.
4. Allow user to edit column names.
5. Add checkbox for "Set at default", selected by default on the first run. Save mapping to layer when ticked.
6. Pass mapping to export function, use in results.

@kdyson Wdyt? This would be a longer-term solution.

kdyson:
That sounds reasonable.

gino-m:
@jo-spek FYI

jo-spek:
Great feature that would enable GeoJSON fields to be automatically ingested by the EU Information System, that accepts "Area", "ProducerCountry", "ProducerName" and "ProductionPlace".

Issue #716, opened 2021-05-20
[Export data] Include imported LOI properties in exported data


gino-m:
@kdyson FYI

gino-m:
Some options discussed with @kdyson:

1. Add one column per property, but a) they might collide with question names used for column headings and b) we'd need to scan the entire database twice, once to collect all the column names from GeoJSON properties, and one to actually export the data.
2. Add a column called "properties" which contains a JSON blob. I'm not a big fan of this approach since it makes analysis harder (e.g. won't just work in Sheets).
3. Surround properties with brackets or pre/append special symbols (e.g., "[id_prod]").
4. Allow the user to define their own column names at export time (w/option to set as default).
5. Export as a separate table, user would need to join by uuid.

PR #772, opened 2021-06-23
[Export observations] Export feature id and label properties

gino-m:
This is a temporary solution until we can implement #716 and/or #769.

@DaoyuT PTAL?


DaoyuT (approved):
LGTM

PR #781, opened 2021-07-06
Fix drag dialog

DaoyuT:
Force angular to refresh when dragging marker starts.

This should fix the bug when dragging a newly created marker does not display the confirmation dialog for repositioning.

Diffbase=#765


gino-m (approved):
LGTM! 🚀

PR #789, opened 2021-07-21
[Layer dialog] Pressing the ESC key should have the same effect as clicking "Cancel"

### LINKED ISSUE #787: [Layer dialog] Pressing the ESC key should have the same effect as clicking "Cancel"
@parulraheja98 PTAL?

parulraheja98:
Closes #787


gino-m (approved):
🚢  IT!

PR #791, opened 2021-07-21
[Layer dialog] Drag affordance for multiple choice options should only show on hover and focus

### LINKED ISSUE #785: [Layer dialog] Drag affordance for multiple choice options should only show on hover
image

@DakshSaini Looks like a regression since this was fixed in #730. PTAL?

DakshSaini:
Fixes #785


gino-m (approved):
Nice! ⏸️ 🚫 🕵️  Thanks!

PR #792, opened 2021-07-28
[Feature details] Update label of header based on selected feature

### LINKED ISSUE #344: [Feature details] Update header based on selected feature
Current design:
image

Some users seeing the demo have been confused when we just show the layer name here. What can we show to let the user know they're seeing details about the selected point/line/polygon? Note that @Tiyarah is adding an import feature that will allow a custom caption to be specified for each feature.

So ideally the heading would reflect that:
* The panel is showing details for the selected point/line/polygon
* The selected feature's caption, if present
* In which layer the feature resides.

For example, the heading could show the custom caption if present, the feature type (Point / Line / Polygon), the feature's Plus Code, and the name of the layer.

@sergeydolgov1 @jacobmclaws for guidance.

parulraheja98:
Closes #344

### DIFF
--- a/web-ng/src/app/components/feature-panel-header/feature-panel-header.component.ts
+++ b/web-ng/src/app/components/feature-panel-header/feature-panel-header.component.ts
@@ -34,7 +34,7 @@ import {
   LocationFeature,
 } from '../../shared/models/feature.model';
 import { FeatureService } from '../../services/feature/feature.service';
-
+import { Map } from 'immutable';
 @Component({
   selector: 'ground-feature-panel-header',
   templateUrl: './feature-panel-header.component.html',
@@ -50,6 +50,9 @@ export class FeaturePanelHeaderComponent
   readonly featureType = FeatureType;
   subscription: Subscription = new Subscription();
   featureTypeValue?: FeatureType;
+  private readonly CAPTION_PROPERTIES = ['caption', 'label', 'name'];
+  private readonly ID_PROPERTIES = ['id', 'identifier', 'id_prod'];
+  private featureProperties?: Map<string, string | number>;
 
   constructor(
     private sanitizer: DomSanitizer,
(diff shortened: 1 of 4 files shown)


web-ng/src/app/shared/models/feature.model.ts
```diff
@@ -15,33 +15,38 @@
...
+  readonly properties?: Map<string, string | number>;
 }
 
 export class LocationFeature implements Feature {
   constructor(
     readonly id: string,
     readonly layerId: string,
     // TODO: User custom type instead of exposing types from data layer.
-    readonly location: firebase.firestore.GeoPoint
+    readonly location: firebase.firestore.GeoPoint,
```
gino-m, line 31: Leaky abstraction: GeoPoint is specific to the remote datastore layer. In our domain model it abstracted as `Point`

gino-m (changes requested):
LGTM overall, suggested minor changes. @DaoyuT could you also PTAL?

gino-m:
Hi @parulraheja98! Would you be able to circle back to this one?

web-ng/src/app/shared/models/feature.model.ts
gino-m: We chatted; sorry I hadn't realized we don't have `Point` class defined in the model yet. Please disregard for this PR, we'll circle back.

gino-m (approved):

web-ng/src/app/components/feature-panel-header/feature-panel-header.component.ts
```diff
@@ -130,6 +134,34 @@ export class FeaturePanelHeaderComponent
     });
   }
 
+  getFeatureName(): string | number {
```
gino-m, line 137: @DaoyuT This is the method we were talking about that may need to be refactored (and `| number` removed`).

gino-m: `FeatureService` could be a good candidate.

PR #808, opened 2021-08-18
Add unit tests to map component.

DaoyuT:
Add unit tests to map component.


gino-m:
@DaoyuT Something seems to be failing with:
```
ERROR: 'InvalidValueError: setMap: not an instance of Map; and not an instance of StreetViewPanorama'
```

web-ng/src/app/components/map/map.component.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+import { GoogleMapsModule } from '@angular/google-maps';
+
+// import { initialize } from "@googlemaps/jest-mocks";
+// import createGoogleMapsMock from 'jest-google-maps-mock';
+
+describe('MapComponent', () => {
+  let component: MapComponent;
+  let fixture: ComponentFixture<MapComponent>;
+  const dialog: Partial<MatDialog> = {};
+  const map: jasmine.SpyObj<google.maps.Map> = jasmine.createSpyObj(
```
gino-m, line 39: IIuc, this uses as real Map instance but stubs out selected methods. This still makes the test dependent on the impl of Map. Can we use a fake here instead, declaring stubs only for the methods we use?

web-ng/karma.conf.js
```diff
@@ -54,5 +54,15 @@ module.exports = function (config) {
         flags: ['--no-sandbox'],
       },
     },
+    // files: ['src/app/shared/mocks/google-maps-api.js'],
```
gino-m, line 57: Remove commented lines

web-ng/src/app/app.component.ts
```diff
@@ -29,8 +29,11 @@ import { environment } from '../environments/environment';
 })
 export class AppComponent {
   public constructor(@Inject(DOCUMENT) private doc: Document) {
-    this.initGoogleMap();
+    if (environment.googleMapsApiKey !== '') {
```
gino-m, line 32: Please add a comment indicating this is for testing

gino-m (changes requested):
Nice work, pretty good coverage of the map component!

@os-micmec Could you also PTAL?

web-ng/src/app/app.component.ts
DaoyuT: Added an `env` entry in environment variables, and used that.

gino-m (dismissed):

PR #831, opened 2021-11-10
Added the "Building and Deploying Ground from Source" guide to /docs

mmisim:
I added the guide to the docs/ directory so that it will be accessible as a standalone page via https://groundplatform.org/build-and-deploy-ground-from-source.html, as well as via the rendered markdown file from the Github repo. Going forward, I think we should consider whether we want to repurpose the public-facing https://groundplatform.org site for all of the Ground documentation. If we do, we should chose a more docs-friendly Jekyll theme with better navigation. We could then fold in the pages from the wikis. I'm not sure if we want to keep something resembling the current landing page, though, and as far as I can tell Github pages only supports one site per repo.

Source: https://docs.google.com/document/d/12wDlYhTktsCvZuYVudOchpq_pykoNONxID01JL35elM/edit


gino-m (approved):
Excellent work! Looking forward to sharing this with interested partners!

gino-m:
> Going forward, I think we should consider whether we want to repurpose the public-facing https://groundplatform.org site for all of the Ground documentation.

Agreed! Do you think it would make sense to start on that transition now for developers, or once we have something non-technical users can spin up?

mmisim:
> Agreed! Do you think it would make sense to start on that transition now for developers, or once we have something non-technical users can spin up?

I think it depends on whether we want a non-technical public site in the meantime. If we transition this site to a developer docs portal now, it could make Ground as a whole seem less accessible for less-technical users. Also, developers are going to gravitate towards Github anyway, where they can view the rendered markdown files, so there may not be much value in spinning up a separate docs site now. Maybe once we have a non-technical implementation, we can reevaluate what developer-focused content is still relevant, and then rebuild the site to address the needs of developers and non-developers.

WDYT?

gino-m:
@mmisim That makes sense. Let's make the external site more user-focused, we can also decide whether it makes sense to keep developer docs there or not once we get to that point. Thanks!

PR #834, opened 2021-11-11
Add tests to DialogService

os-micmec:


web-ng/src/app/services/dialog/dialog.service.spec.ts
```diff
@@ -17,17 +17,49 @@
...
 
   it('should be created', () => {
     expect(service).toBeTruthy();
   });
+
+  it('should return correct dialog without discard actions', () => {
+    service.openConfirmationDialog('testTitle', 'testMessage');
+
+    expect(dialogSpy).toHaveBeenCalledWith(ConfirmationDialogComponent, {
+      maxWidth: '500px',
```
gino-m, line 42: Only check non-constant values  here and throughout (https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). You can use a partial matcher you showed me today.

web-ng/src/app/services/dialog/dialog.service.spec.ts
```diff
@@ -17,17 +17,49 @@
...
+      autoFocus: false,
+      data: {
+        title: 'testTitle',
+        message: 'testMessage',
+        showDiscardActions: false,
+      },
+    });
+  });
+
+  it('should return correct dialog with discard actions', () => {
```
gino-m, line 52: If the public entry point to the class being tested is a single method I think it would be clearer to group tests by the method name `describe(..)`. In cases where multiple methods are tests at once we can adapt the description accordingly

web-ng/src/app/services/dialog/dialog.service.spec.ts
```diff
@@ -17,17 +17,49 @@
...
   });
+
+  it('should return correct dialog without discard actions', () => {
+    service.openConfirmationDialog('testTitle', 'testMessage');
+
+    expect(dialogSpy).toHaveBeenCalledWith(ConfirmationDialogComponent, {
+      maxWidth: '500px',
+      autoFocus: false,
+      data: {
+        title: 'testTitle',
```
gino-m, line 45: Consider assigning repeated constants to a local constant.

web-ng/src/app/services/dialog/dialog.service.spec.ts
```diff
@@ -17,17 +17,49 @@
...
 
 describe('DialogService', () => {
   let service: DialogService;
+  let dialogSpy: jasmine.Spy;
 
   beforeEach(() => {
     TestBed.configureTestingModule({ imports: [MatDialogModule] });
     service = TestBed.inject(DialogService);
+
+    dialogSpy = spyOn(TestBed.inject(MatDialog), 'open');
```
gino-m, line 31: Do we need to spy on a real `MatDialog`, or can we just used `createSpyObj` to create a dialog with the methods we need to spy on"?

gino-m (changes requested):
A few comments, sorry for the delay!

crypt096 (dismissed):

web-ng/src/app/services/dialog/dialog.service.spec.ts
gino-m: Hi @os-micmec Gentle ping!

web-ng/src/app/services/dialog/dialog.service.spec.ts
os-micmec: Okay, I changed the describe name! Let me know if this works now.

web-ng/src/app/services/dialog/dialog.service.spec.ts
os-micmec: Done!

web-ng/src/app/services/dialog/dialog.service.spec.ts
os-micmec: Done!

gino-m (approved):

gino-m (approved):

Issue #835, opened 2021-11-12
Creating date or time fields in web client is failing


rachaprince:
**Describe the bug**
Date / Time inputs cannot be added to the layer form 

**To Reproduce**
1. Add a date or time field to a layer form and try to save the layer.

**Expected behavior**
1.  The layer saves and updates

**Actual behavior**
1. An exception is thrown
```
main-es2018.195642eb39c14cdc3fdf.js:1 Error: Unsupported field type 5
    at Function.fieldTypeToString (main-es2018.195642eb39c14cdc3fdf.js:1)
    at Function.fieldToJS (main-es2018.195642eb39c14cdc3fdf.js:1)
    at main-es2018.195642eb39c14cdc3fdf.js:1
    at main-es2018.195642eb39c14cdc3fdf.js:1
    at main-es2018.195642eb39c14cdc3fdf.js:1
    at Ce.Mpt7.Ce.iterate.Ie.iterate (main-es2018.195642eb39c14cdc3fdf.js:1)
    at e.__iterate (main-es2018.195642eb39c14cdc3fdf.js:1)
    at Pn (main-es2018.195642eb39c14cdc3fdf.js:1)
    at e.reduce (main-es2018.195642eb39c14cdc3fdf.js:1)
    at Function.formToJS (main-es2018.195642eb39c14cdc3fdf.js:1)
```

gino-m:
@os-micmec @DaoyuT Would either of you have cycles today to take a look at this and #836?

PR #836, opened 2021-11-12
Fix date time fields

rachaprince:
Fixes issue #835  by adding date and time fields to the firebase data converter.

**How was this tested?**
- On a locally running Firebase emulator/ web app
- Unit tests


gino-m (approved):
Thanks for fixing this so quickly!

Issue #821, opened 2021-10-15
[Code health] Implement unit test coverage of all non-trivial business logic


gino-m:

gino-m:
@gino-m To survey code and outline components/services in need of tests.

gino-m:
To start:

* @os-micmec DialogService
* @gino-m DataImport
* @DaoyuT DrawingTools

os-micmec:
For next week let's do:

- @os-micmec AuthService
- @DaoyuT FeatureService
- @gino-m NotificationService + ObservationService

PR #843, opened 2021-11-18
Add tests for NotificationService

gino-m:
Contributes to #821.

@os-micmec PTAL?

### DIFF
--- a/web-ng/src/app/services/notification/notification.service.spec.ts
+++ b/web-ng/src/app/services/notification/notification.service.spec.ts
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


web-ng/src/app/services/notification/notification.service.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      providers: [{ provide: MatSnackBar, useValue: matSnackBar }],
+    });
+    service = TestBed.inject(NotificationService);
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  describe('success', () => {
```
os-micmec, line 37: This is probably a stupid q, but what's the advantage of having this sub-describes over just having multiple it()'s?

gino-m: For methods with multiple cases it groups tests together in code and in outputs without us needed to repeat the method name in each test description.

crypt096 (approved):

os-micmec (approved):

PR #844, opened 2021-11-18
Add tests for ObservationService

gino-m:
Contributes to #821.

Also adds `TestData` builder class with reasonable default values.

### DIFF
--- a/web-ng/src/testing/test-data.ts
+++ b/web-ng/src/testing/test-data.ts
@@ -1,0 +1,80 @@
+/**
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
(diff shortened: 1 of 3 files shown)


web-ng/src/testing/test-data.ts
```diff
@@ -0,0 +1,80 @@
...
+import { User } from './../app/shared/models/user.model';
+import firebase from 'firebase/app';
+import { Form } from '../app/shared/models/form/form.model';
+import { Role } from './../app/shared/models/role.model';
+
+/**
+ * Shorthand builders with reasonable defaults for use by tests. Tests should
+ * override values that are relevant even if they're the same as the default
+ * values. Defaults values should only be used for states which are not under
+ * test.
```
gino-m, line 31: @os-micmec @DaoyuT FYI, how do you feel about this approach? See `newProject()` below for example.

DaoyuT: This is helping if I need 1 project or 1 layer, what if I need multiple? How can we help that use case?

gino-m: You can call `newProject()` multiple times with different args, incl different layers.

gino-m: Note that I haven't added default args to methods besides `newProject()` and `newLayer`() yet but I plan to.

gino-m:
Obsolete.M erged in #902.

PR #847, opened 2021-11-30
[Add/edit observation] Add affordance for required fields

os-micmec:
Ensure that users can visualize which fields are required in their forms.

Screen Shot 2021-11-30 at 10 59 12

Added tests to check this is respected.


web-ng/src/app/components/observation-form/observation-form.component.spec.ts
```diff
@@ -212,6 +222,46 @@ describe('ObservationFormComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('should create text fields with right "required" option', () => {
+    for (const el of fixture.debugElement.queryAll(
+      By.css('.field-response div mat-form-field input')
+    )) {
+      if (component.observationFields) {
```
gino-m, line 230: Add break; here (and below).

os-micmec: Done!

gino-m (approved):

web-ng/src/app/components/observation-form/observation-form.component.css
```diff
@@ -42,3 +42,16 @@
...
+  margin-bottom: 1em;
+}
+/* The spacing between radio buttons and checkboxes should be larger */
+.mat-radio-button,
+.mat-checkbox {
+  margin-right: 1em;
+}
+.asterix--after:after {
+  content: '*';
+  color: red;
```
gino-m, line 56: Please use Material themes for colors and fonts, in this case `mat-color($warn)` is probably the right one to use. Example in https://github.com/google/ground-platform/blob/44f31e2d115edd899f88579372d336964fa38c95/web-ng/src/styles.scss

gino-m (changes requested):

web-ng/src/app/components/observation-form/observation-form.component.css
os-micmec: Done! Changed stylesheet to SCSS.

gino-m (approved):

PR #848, opened 2022-01-07
[Persistence] Fix conversion of multiple choice responses

gino-m:
@atalekar @rachaprince Could you kindly test against your db?

### DIFF
--- a/web-ng/src/app/services/observation/observation.service.spec.ts
+++ b/web-ng/src/app/services/observation/observation.service.spec.ts
@@ -15,6 +15,7 @@
  */
 
 import { TestBed } from '@angular/core/testing';
+import { TestData } from '../../../testing/test-data';
 import { ObservationService } from './observation.service';
 import { DataStoreService } from '../data-store/data-store.service';
 import { ProjectService } from '../project/project.service';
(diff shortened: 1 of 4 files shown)


rachaprince (dismissed):

gino-m:
@rachaprince Sorry I didn't update this before; upon further inspection, I'm not sure what value is being considered invalid in your instance. I've updated this PR to add logging so that we can see what the value and its type are when the error occurs.

rachaprince:
I believe the values being considered invalid in our instance are the Timestamp and multiple choice values, based on the logs I collected. Let me know if I can do more investigating on our end.

when I test on https://github.com/google/ground-platform/pull/849, I no longer see errors in our db

gino-m:
> I believe the values being considered invalid in our instance are the Timestamp and multiple choice values, based on the logs I collected. Let me know if I can do more investigating on our end.
> 
> when I test on #849, I no longer see errors in our db

Sg. Would you like to add the failsafe added here into #848 instead so we can catch this if it happens again (i.e., log an error and return a default value)?

gino-m:
@rachaprince Reviving this fix - could you please restamp?

gino-m:
@rachaprince Correction: Tests are failing; is this PR still needed? If so I'll fix the tests, if not I'll revert.

gino-m:
Abandoning this PR as it wasn't used.

PR #849, opened 2022-01-13
Add support for Array and Timestamp dataconversion

rachaprince:
Add support for Timestamp conversions, and fix error in multiple choice conversion


gino-m (approved):
LGTM! Have you tested this with both Date and Time field types?

rachaprince:
Hmm, actually it looks like more work needs to be done to support Date and Time fields in the UI side-panel. I'll take a look and see

PR #853, opened 2022-01-28
Revert "Add support for Array and Timestamp dataconversion"

gino-m:
@rachaprince I merged before seeing your comment in google/ground-platform#849 , should I revert?


rachaprince:
I think that this is still an improvement on what was in the code before. Unfortunately I don't think I have bandwidth this week to add frontend support for date and time fields. Should I make a github issue for it?

gino-m:
> I think that PR #849 is still an improvement on what was in the code before, so we probably don't need to revert it. Unfortunately I don't think I have bandwidth this week to add frontend support for date and time fields. Should I make a github issue for add UI support for Date and Time fields?

Sure, sgtm. Thanks for filing.

PR #885, opened 2022-06-23
Add comment to test CD

os-micmec:
I will remove this later!


gino-m (approved):
LGTM. I assume this is meant to test CD, not CI as the PR descriptions suggests? If so you may want to update it for posterity.

Issue #873, opened 2022-05-18
[Migration] Rename all the things


gino-m:
Classes, variables, Firebase collection/project/field keys:

- [x] Project -> Survey [@gino-m]
- [x] Layers -> Job [@os-micmec]
- [x] Forms -> Task [@shobhitagarwal1612]
- [x] Feature -> LocationOfInterest  [@gino-m]
- [x] Field (aka Question/Element) -> Step [@DaoyuT]
- [x] Observation -> Submission [@gino-m]
- [x] Response -> Result [@gino-m]

User role enum and Firebase strings:

- [x] manager -> survey-organizer [@DaoyuT]
- [x] contributor -> data-collector [@DaoyuT]

PR #902, opened 2022-07-19
Rename feature to location of interest / loi

gino-m:
Work towards #873.

@DaoyuT  PTAL?

New tests for ObservationService accidentally leaked into this PR. For expediency, can we merge them together with the rename?


gino-m:
@DaoyuT Note that I use both `loi` and `LocationOfInterest`, trying to use good judgement where each is more useful (clarify vs brevity).

DaoyuT (approved):

PR #905, opened 2022-07-20
Use consistent prefix (`ground`)

### LINKED ISSUE #888: [Code health] Use consistent app component prefix
As per:
https://angular.io/guide/styleguide#component-custom-prefix

Replace symbol `gnd` and `app` with `ground` wherever it occurs.

gino-m:
Fixed #888.

### DIFF
--- a/web/src/index.html
+++ b/web/src/index.html
@@ -27,6 +27,6 @@ limitations under the License.
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
 </head>
 <body>
-  <app-root></app-root>
+  <ground-root></app-root>
 </body>
 </html>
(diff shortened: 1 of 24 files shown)


web/src/index.html
```diff
@@ -14,19 +14,25 @@
 limitations under the License.
 -->
 
-<!doctype html>
```
os-micmec, line 17: Is this needed in this PR?

gino-m: Prettier did it.. Should we allow Prettier to reformat as we go, or should we do a global reformat separately?

os-micmec: We can keep this reformatting in this PR for simplicity, these comment add enough context. I think it's a good idea to run a general reformatting in a separate PR.

os-micmec (approved):

Issue #913, opened 2022-07-22
[Model] Disambiguate "task" by merging with "step"


gino-m:
Suggest proceeding the following steps:

- [x] Rename `element` to `step`
- [x] Move `steps` up to `job` and delete `tasks`
- [ ] Rename `steps` to `tasks`

See also https://github.com/google/ground-android/issues/1164

PR #915, opened 2022-07-25
Pull `Step` up into `Job` and remove `Task`

gino-m:
This PR is getting large; I'll do renames and other cleanup in a separate PR.

Towards #913.

@DaoyuT / @os-micmec PTAL?


gino-m:
@DaoyuT / @os-micmec Gentle ping!

DaoyuT (approved):
The rename make sense to me. But why are we removing Task which was Forms before? I think it'd be helpful for me to understand if we can sync for 10 mins tmr, WDYT?

gino-m:
> The rename make sense to me. But why are we removing Task which was Forms before? I think it'd be helpful for me to understand if we can sync for 10 mins tmr, WDYT?

Good q - this is an intermittent step before https://github.com/gino-m/ground-platform/pull/1, as described in #913. The end state will be a set of `Tasks` (formerly known as `Steps`) directly under `Job`. We've removed the extra collection formerly known as `Forms` altogether; it was never used since one `Job` can only have one set of `Tasks` (ex. `Steps`). Does that make sense?

PR #923, opened 2022-07-28
[Import] Update Cloud Functions to use latest schema

gino-m:
Including new Firestore keys and GeoJSON-like geometry schema.

@os-micmec PTAL?


gino-m:
Note that I fix the lint warnings present in this PR in #924.

os-micmec (approved):

Issue #925, opened 2022-07-31
Can't deploy prod build to to Firebase hosting from local


gino-m:
Deploy works, but getting the following error after deploying `prod` built on local PC using `./build.sh`:

![image](https://user-images.githubusercontent.com/228050/182063344-b9a0705e-ff26-431a-ad2e-10d7189b94a4.png)

Looking in build artifact, it seems `process.env.*` are not getting replaced with actual API keys.

To fix this we could create alternate build environments, one using the keys in the local `.*` files, and one using the `env` values set in CD.

@os-micmec Would you like to give this one a shot?

gino-m:
@os-micmec Is local deploy working again? If so let's close this one.

os-micmec:
This is working now!

PR #927, opened 2022-08-01
Fix CD by injecting env variables

os-micmec:
It looks like Angular is not able to access the provided env variables. Instead, node.js is able to access them, so we create a small script that injects the variables into the `environments.prod.ts` file.

Towards #925 and #880

### DIFF
--- a/.github/workflows/firebase-deploy.yml
+++ b/.github/workflows/firebase-deploy.yml
@@ -33,22 +33,22 @@ jobs:
         run: npm install -g @angular/cli > /dev/null
 
       - name: Create firebase-config.ts
-        run: touch src/environments/.firebase-config.ts && echo $FIREBASE_CONFIG > src/environments/.firebase-config.ts
-        shell: bash
+        uses: "finnp/create-file-action@master"
         env:
-          FIREBASE_CONFIG: ${{ secrets.FIREBASE_CONFIG }}
+          FILE_NAME: "src/environments/.firebase-config.ts"
+          FILE_DATA: ${{ secrets.FIREBASE_CONFIG }}
 
       - name: Create google-maps-config.ts
-        run: touch src/environments/.google-maps-config.ts && echo $GOOGLE_MAPS_CONFIG > src/environments/.google-maps-config.ts
-        shell: bash
+        uses: "finnp/create-file-action@master"
         env:
-          FIREBASE_CONFIG: ${{ secrets.GOOGLE_MAPS_CONFIG }}
+          FILE_NAME: "src/environments/.google-maps-config.ts"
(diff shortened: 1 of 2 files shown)


.github/workflows/ng-env-replace.js
```diff
@@ -0,0 +1,16 @@
...
+    .replace('FIREBASE_CONFIG_AUTH_DOMAIN', process.env.FIREBASE_CONFIG_AUTH_DOMAIN)
+    .replace('FIREBASE_CONFIG_DATABASE_URL', process.env.FIREBASE_CONFIG_DATABASE_URL)
+    .replace('FIREBASE_CONFIG_PROJECT_ID', process.env.FIREBASE_CONFIG_PROJECT_ID)
+    .replace('FIREBASE_CONFIG_STORAGE_BUCKET', process.env.FIREBASE_CONFIG_STORAGE_BUCKET)
+    .replace('FIREBASE_CONFIG_MESSAGING_SENDER_ID', process.env.FIREBASE_CONFIG_MESSAGING_SENDER_ID)
+    .replace('FIREBASE_CONFIG_APP_ID', process.env.FIREBASE_CONFIG_APP_ID)
+    .replace('CLOUD_FUNCTIONS_URL', process.env.CLOUD_FUNCTIONS_URL)
+    .replace('OFFLINE_BASE_MAP_SOURCES_URL', process.env.OFFLINE_BASE_MAP_SOURCES_URL);
+
+fs.writeFileSync(targetPath, newFile);
```
gino-m, line 16: Won't this update the actual source file, not the build?

Perhaps it would it be simpler to create a separate `environment.cd.ts` that uses refers directly to `process.env` as `environment.prod.ts` did, and we can let `environment.prod.ts` refer to the .keyfiles for local builds?

gino-m (changes requested):

.github/workflows/firebase-deploy.yml
```diff
@@ -33,22 +33,22 @@ jobs:
         run: npm install -g @angular/cli > /dev/null
 
       - name: Create firebase-config.ts
-        run: touch src/environments/.firebase-config.ts && echo $FIREBASE_CONFIG > src/environments/.firebase-config.ts
-        shell: bash
+        uses: "finnp/create-file-action@master"
         env:
-          FIREBASE_CONFIG: ${{ secrets.FIREBASE_CONFIG }}
+          FILE_NAME: "src/environments/.firebase-config.ts"
+          FILE_DATA: ${{ secrets.FIREBASE_CONFIG }}
```
gino-m, line 39: Ah, so you can just define the file contents in the GitHub action? That's cool!

gino-m (approved):
Lgtm!

Issue #926, opened 2022-08-01
[Code health] Consolidate representation of points, polygons, and GeoJSON


gino-m:
See also https://github.com/google/ground-android/issues/929.

PR #928, opened 2022-08-07
Implement basic topology classes

gino-m:
Work towards #926. These will replace the various LOI implementations, and a reference to Geometry will be added to the main LOI implementation.

@DaoyuT PTAL?

### DIFF
--- a/web/src/app/shared/models/geometry/linear-ring.ts
+++ b/web/src/app/shared/models/geometry/linear-ring.ts
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


web/src/app/shared/models/geometry/linear-ring.ts
```diff
@@ -0,0 +1,33 @@
...
+ * must be equal, and the ring must not self-intersect. Either orientation of
+ * the ring is allowed.
+ *
+ * A ring is assumed to have either 0 or 3 or more points, and the first and
+ * last points are assumed to be equal (in 2D).
+ *
+ * Based on https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/LinearRing.html
+ */
+export class LinearRing extends LineString {
+  geometryType = GeometryType.LINEAR_RING;
```
DaoyuT, line 32: Should this class have a constructor with points?

gino-m: Wouldn't it inherit its superclass constructor? That should be sufficient?

gino-m: Done.

DaoyuT (approved):

PR #929, opened 2022-08-10
Load new geometry model from db

gino-m:


web/src/app/shared/converters/geometry-converter.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const map: { [key: string]: any } = {};
+  path.forEach((it, idx) => (map[String(idx)] = new GeoPoint(it[0], it[1])));
+  return map;
+}
+
+fdescribe('geometry-converter.ts', () => {
+  describe('toGeometry', () => {
+    // Valid / passing test cases.
+    [
```
gino-m, line 93: @os-micmec FYI, introducing parameterized testing here as you suggested. 👍

DaoyuT (approved):
LGTM

Issue #932, opened 2022-08-24
[Code health] Simplify and error handle in Firestore data converter


os-micmec:
There is a number of improvements we can make to our Firestore data conversion class.

- [ ] Split the different data conversions (LOI, Submission, etc) in their own separate classes
- [ ] Add/improve tests
- [ ] Return errors on the methods exposed by the data conversion classes

gino-m:
@os-micmec pasting in description from duplicate #937. Please feel free to merge the two if it helps:

- [ ] Fail gracefully when bad/malformed/missing data is found in Firestore
  - [ ]  Return `Foo | Error` from public datastore methods instead of throwing Errors. (Errors ok to use internally in private methods where it increases readability).
  - [ ] Handle errors by log the issue to console and either ignoring the entity, or displaying an error in the UI where appropriate
- [ ] Split firestore-data-converter.ts into per-class converters (e.g., JobConverter, TaskConverter, etc.)
  - [ ] In the process, add missing test coverage (no round trip). Parameterized testing example in [tests in #929](https://github.com/google/ground-platform/pull/929/files#diff-ce83bf3978ee20cb9e694f687980cf972b15b13842d42c62c2d6c3f1f2917ef0R99)

gino-m:
@DaoyuT Related to your current work on capturing auth errors

gino-m:
We downscoped #931, so this is now a duplicate.

PR #933, opened 2022-08-24
Extract LOI data converter from Firestore one

os-micmec:
Towards #932 

We will perform further extractions later.


gino-m (approved):
LGTM!

PR #945, opened 2022-09-01
Migrate cloud functions from js to ts.

DaoyuT:
WIP

@gino-m please review, I marked all the places I wasn't sure if I migrated correctly with `???`
 - [x] should we return response? (I don't think we have to return response, I think we send response by calling functions in response)
 - [x] find replacement for `next()` in busboy
 - [x] there is no `req.rawBody`, so I replaced it with `req.body`

### DIFF
--- a/.github/workflows/firebase-deploy.yml
+++ b/.github/workflows/firebase-deploy.yml
@@ -63,6 +63,6 @@ jobs:
       - name:  Deploy to Firebase functions
         uses: w9jds/firebase-action@master
         with:
-            args: deploy --only functions
+            args: deploy --only functions --project gnd-dev
         env:
             FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}


functions/src/common/datastore.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2018 Google LLC
+ * Copyright 2022 Google LLC
```
gino-m, line 3: We should retain year the file was created. We can optionally add a range (`2018-2022`), but we haven't been doing that so far.

functions/src/import-csv.ts
```diff
@@ -76,24 +79,26 @@ async function importCsv(req, res) {
...
   busboy.on("error", (err) => {
-    console.err("Busboy error", err);
-    next(err);
+    console.error("Busboy error", err);
+    // TODO bring next function back ????
+    // next(err);
   });
-  busboy.end(req.rawBody);
+  // rawBody ???
+  busboy.end(req.body);
```
gino-m, line 91: I think we can remove the call to `end` here.

functions/src/import-csv.ts
```diff
@@ -76,24 +79,26 @@ async function importCsv(req, res) {
...
-    await res.status(HttpStatus.OK).end(JSON.stringify({ count }));
+    res.status(HttpStatus.OK).end(JSON.stringify({ count }));
   });
 
   busboy.on("error", (err) => {
-    console.err("Busboy error", err);
-    next(err);
+    console.error("Busboy error", err);
+    // TODO bring next function back ????
+    // next(err);
```
gino-m, line 88: I don't think `next` is accessible here unless we pass an Express app to onHttpRequest. Instead, I think we can be just unpipe and return an error as we do above at line 69-70.

gino-m (changes requested):
Great work @DaoyuT! 

To your question, I agree we shouldn't be returning a response from the handlers. Other answers inline.

functions/src/common/datastore.ts
DaoyuT: I'll do the created year then.

But wait, how did this pick up the diff base? What I did was just create new ts file, copy content over and edit, and then delete js file. How did it know? 🧐

functions/src/import-csv.ts
DaoyuT: Here and every other occasion right?

functions/src/import-csv.ts
DaoyuT: like this?
```
busboy.on("error", (err) => {
        console.err("Busboy error", err);
        res.unpipe(busboy);
});
```

functions/src/import-csv.ts
gino-m: Only here. I think we hand it over to busboy to handle the stream and call the event handlers, which in turn may call end.

functions/src/common/datastore.ts
gino-m: It's clever!

functions/src/import-csv.ts
gino-m: Yes, and we may want to also set the response status to 500 and include the error message there as well.

gino-m (changes requested):
Some answers!

functions/src/import-csv.ts
DaoyuT: I meant removing the other `busboy.end(req.body);`, no?

gino-m: Yes, I see there are two instants, one here and one in import-geojson.ts. Those appear to be unnecessary.

gino-m (approved):
LGTM!

functions/package.json
```diff
@@ -1,33 +1,45 @@
...
   "description": "Cloud Functions for Firebase",
   "scripts": {
     "serve": "firebase serve --only functions",
     "shell": "firebase experimental:functions:shell",
     "emulators": "firebase emulators:start --config ../firebase.local.json --project fake-project-id",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
-    "test": "mocha --reporter list"
+    "test": "mocha --reporter list",
+    "build": "npm run lint && tsc"
```
gino-m, line 12: `npm run build` fails since `lint` isn't defined here as a script.

functions/src/import-csv.ts
```diff
@@ -76,24 +77,24 @@ async function importCsv(req, res) {
...
 /**
  * Transforms a dictionary of string in the form A->B[] into a dictionary of
  * strings in the form B->A. Values in B[] are assumed to appear at most once
  * in the array values in the provided dictionary.
  */
-function invertAndFlatten(obj) {
+function invertAndFlatten(obj: any) {
   return Object.keys(obj)
-    .flatMap((k) => obj[k].map((v) => ({ k, v })))
+    .flatMap((k) => obj[k].map((v: any) => ({ k, v })))
```
gino-m, line 97: Still getting compile errors:
```

src/common/datastore.ts:106:19 - error TS2550: Property 'fromEntries' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

106     return Object.fromEntries(
                      ~~~~~~~~~~~

src/common/datastore.ts:125:21 - error TS2550: Property 'fromEntries' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

125       return Object.fromEntries(
                        ~~~~~~~~~~~

src/import-csv.ts:97:6 - error TS2550: Property 'flatMap' does not exist on type 'string[]'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

97     .flatMap((k) => obj[k].map((v: any) => ({ k, v })))
        ~~~~~~~

src/import-csv.ts:97:15 - error TS7006: Parameter 'k' implicitly has an 'any' type.

97     .flatMap((k) => obj[k].map((v: any) => ({ k, v })))
                 ~

src/import-csv.ts:98:14 - error TS7006: Parameter 'o' implicitly has an 'any' type.

98     .reduce((o, { v, k }) => {
                ~

src/import-csv.ts:98:19 - error TS7031: Binding element 'v' implicitly has an 'any' type.

98     .reduce((o, { v, k }) => {
                     ~

src/import-csv.ts:98:22 - error TS7031: Binding element 'k' implicitly has an 'any' type.

98     .reduce((o, { v, k }) => {
                        ~
```

gino-m (changes requested):
A few suggested edits addressing some of the issues I mention in comments is in:
https://github.com/DaoyuT/ground-platform/pull/1

The PR also resolves merge conflicts with `master`.

Note that you should be able to test these changes locally, in two shells:

Session 1:

```
$ cd functions
$ npm run compile && npm run emulators
```

UI will be available at http://localhost:4000/

Then in shell 2:

```
$ cd web
$ npm i && ng serve -c local
```

Then:
1 . Go to http://localhost:4200/survey/new and create a survey with a simple Job.
2. In the side menu, click the overflow menu and click "Import lois"
3. Save this same to a .geojson file and try uploading:

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": { "type": "Point", "coordinates": [102.0, 0.5] },
      "properties": { "prop0": "value0" }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [100.0, 0.0],
            [101.0, 0.0],
            [101.0, 1.0],
            [100.0, 1.0],
            [100.0, 0.0]
          ]
        ]
      },
      "properties": {
        "prop0": "value0",
        "prop1": { "this": "that" }
      }
    }
  ]
}
```

You should see data appearing in the emulator UI, with no errors in the emulators log.

functions/package.json
DaoyuT: Got the eslint working. I don't know how to set up `gts` and I'm really bad at this. Let's stick to this for now, shall we?

DaoyuT:
> A few suggested edits addressing some of the issues I mention in comments is in [DaoyuT#1](https://github.com/DaoyuT/ground-platform/pull/1). The PR also resolves merge conflicts with `master`.
> 
> You should be able to test these changes locally with the following. In two shells:
> 
> Session 1:
> 
> ```
> $ cd functions
> $ npm run compile && npm run emulators
> ```
> 
> UI will be available at http://localhost:4000/
> 
> Leave that running, then in session 2:
> 
> ```
> $ cd web
> $ npm i && ng serve -c local
> ```
> 
> Then:
> 
> 1 . Go to http://localhost:4200/survey/new and create a survey with a simple Job. 2. In the side menu, click the overflow menu and click "Import lois" 3. Save this same to a .geojson file and try uploading:
> 
> ```
> {
>   "type": "FeatureCollection",
>   "features": [
>     {
>       "type": "Feature",
>       "geometry": { "type": "Point", "coordinates": [102.0, 0.5] },
>       "properties": { "prop0": "value0" }
>     },
>     {
>       "type": "Feature",
>       "geometry": {
>         "type": "Polygon",
>         "coordinates": [
>           [
>             [100.0, 0.0],
>             [101.0, 0.0],
>             [101.0, 1.0],
>             [100.0, 1.0],
>             [100.0, 0.0]
>           ]
>         ]
>       },
>       "properties": {
>         "prop0": "value0",
>         "prop1": { "this": "that" }
>       }
>     }
>   ]
> }
> ```
> 
> You should see data appearing in the emulator UI, with no errors in the emulators log.

Thanks for the instruction, `npm run build` works now and it compiles. But after triggering the `importGeoJson` function in web UI, it timed out with the following error message:
```
functions: Your function timed out after ~60s. To configure this timeout, see
      https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation.
>  /Users/daoyu/.nvm/versions/node/v16.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:635
>              throw new Error("Function timed out.");
>              ^
>
>  Error: Function timed out.
>      at Timeout._onTimeout (/Users/daoyu/.nvm/versions/node/v16.16.0/lib/node_modules/firebase-tools/lib/emulator/functionsEmulatorRuntime.js:635:19)
>      at listOnTimeout (node:internal/timers:559:17)
>      at processTimers (node:internal/timers:502:7)
```
I think we need to debug the cloud function and I'll need your help for that. I'll commit the fix for `npm run build`.

functions/package.json
gino-m: Sgtm, we just need to add `eslint` as a dependency for the build to pass; build fails for me, but I think it may you may have it install globally in your env.

functions/src/common/datastore.ts
```diff
@@ -15,110 +15,111 @@
...
+    );
+  };
+
+  static toFirestoreValue(value: any): any {
+    if (value === null) {
+      return null;
+    }
+    if (Array.isArray(value)) {
+      if (value.length === 2 && value.every((x) => typeof x === "number")) {
+        return new firestore.GeoPoint(value[0], value[1]);
```
gino-m, line 118: In a subsequent commit I reversed the order of x and y since GeoJSON coords are reversed. I'll send a PR with both of our commits shortly.

gino-m:
@DaoyuT Update: Based on output in https://console.cloud.google.com/log, I found that `@terraformer/wkt` needed to be available at runtime. Also, the npm alias used to include `jsonstream` wasn't working, so replaced it both `JSONStream` and its type lib with `jsonstream -ts`. The deploy works now, but both `importCsv` and `importGeoJson` are failing with a timeouts. I'll send a PR with both of our changes once I can get it working on the live dev Firebase.

PR #952, opened 2022-10-06
Converted to use new Point geometry class

nwkotto:
Worked with Daoyu to convert uses of PointOfInterest to GenericLocationOfInterest containing a Point geometry.

Will follow this up to convert geometries.

### DIFF
--- a/web/src/app/components/map/map.component.ts
+++ b/web/src/app/components/map/map.component.ts
@@ -23,7 +23,7 @@ import {
   ChangeDetectorRef,
 } from '@angular/core';
 import { Survey } from '../../shared/models/survey.model';
-import {Point} from '../../shared/models/geometry/point';
+import { Point } from '../../shared/models/geometry/point';
 import {
   LocationOfInterest,
   GeoJsonLocationOfInterest,
@@ -47,7 +47,7 @@ import {
   LocationOfInterestData,
   SelectLocationOfInterestDialogComponent,
 } from '../select-loi-dialog/select-loi-dialog.component';
-import {Coordinate} from '../../shared/models/geometry/coordinate';
+import { Coordinate } from '../../shared/models/geometry/coordinate';
 
 // To make ESLint happy:
 /*global google*/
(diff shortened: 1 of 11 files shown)


web/src/app/shared/converters/loi-converter/loi-data-converter.spec.ts
```diff
@@ -218,16 +213,16 @@ describe('loiToJS_Error', () => {
...
 
   const testData: {
     expectation: string;
     loi: LocationOfInterest;
     wantErrorMessage: string;
   }[] = [
     {
-      expectation: 'Cannot convert generic LOI',
-      loi: new GenericLocationOfInterest(
+      expectation: 'Cannot convert PointOfInterest',
```
DaoyuT, line 224: Not sure if we should keep this test case, maybe add a TODO to remind us deleting it after all the migration is done?

DaoyuT (approved):
LGTM, thanks for the PR

web/src/app/shared/converters/loi-converter/loi-data-converter.spec.ts
nwkotto: I think maybe we can simplify this once the migration is done, but it's still useful to test at least one error case.

web/src/app/components/map/map.component.ts
```diff
@@ -307,15 +319,21 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
-    this.newLocationOfInterestToReposition = new PointOfInterest(
-      loi.id,
-      loi.jobId,
-      new firebase.firestore.GeoPoint(event.latLng.lat(), event.latLng.lng())
+
+    this.newLocationOfInterestToReposition = new GenericLocationOfInterest(
+      id,
+      jobId,
+      new Point(new Coordinate(event.latLng.lat(), event.latLng.lng())),
+      ImmutableMap<string, string | number>()
```
DaoyuT, line 336: nit: is it possible to unify how we construct this field? There is `ImmutableMap<string, string | number>()`, `ImmutableMap()` `Map<string, string | number>()` & `Map()` in this PR. Maybe stick to `Map()` since it's the easiest?

DaoyuT (approved):
LGTM

PR #960, opened 2022-11-14
Update enum string to match GeoJson equivalent string representations.

DaoyuT:
Background:

#952 made change to delete `GEOJSON_GEOMETRY_TYPES` in `geometry-converter.ts`
image
and added string in `GeometryType` enum
image
but it does not match the equivalent string representations in GeoJson

Want LGTM from both @nwkotto & @gino-m

### DIFF
--- a/web/src/app/shared/converters/loi-converter/loi-data-converter.spec.ts
+++ b/web/src/app/shared/converters/loi-converter/loi-data-converter.spec.ts
@@ -21,7 +21,7 @@ import {
   LocationOfInterest,
   PointOfInterest,
 } from '../../models/loi.model';
-import { toGeometry, GEOJSON_GEOMETRY_TYPES } from '../geometry-converter';
+import { toGeometry, GEOMETRY_TYPES } from '../geometry-converter';
 import { LoiDataConverter } from './loi-data-converter';
 import { Map } from 'immutable';
 import { GeoPoint } from 'firebase/firestore';
@@ -31,7 +31,7 @@ import { Point } from '../../models/geometry/point';
 const x = -42.121;
 const y = 28.482;
 const geoPointData = {
-  type: GEOJSON_GEOMETRY_TYPES.get(GeometryType.POINT),
+  type: GEOMETRY_TYPES.get(GeometryType.POINT),
   coordinates: new GeoPoint(x, y),
 };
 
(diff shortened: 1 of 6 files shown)


gino-m:
@DaoyuT The updated geometry model is wire-format agnostic, so wherever possible we would avoid details about GeoJson constants in the shared classes. Can we instead move these to the data later, where the model is (de)serialized to/from GeoJson?

DaoyuT:
> @DaoyuT The updated geometry model is wire-format agnostic, so wherever possible we would avoid details about GeoJson constants in the shared classes. Can we instead move these to the data later, where the model is (de)serialized to/from GeoJson?

Make sense, done.

So far the `GEOJSON_GEOMETRY_TYPES` is used both in `geometry-converter.ts` & `loi-data-converter.ts`, once we are done with all the geometry migration, we can add `geometryToJS` converters in `geometry-converter.ts` and we would only use `GEOJSON_GEOMETRY_TYPES` in `geometry-converter.ts`.

web/src/app/shared/converters/loi-converter/loi-data-converter.ts
```diff
@@ -82,7 +82,7 @@ export class LoiDataConverter {
         jobId,
         geometry: {
           coordinates: new GeoPoint(geometry.coord.x, geometry.coord.y),
-          type: geometry.geometryType,
+          type: GEOJSON_GEOMETRY_TYPES.get(geometry.geometryType),
```
gino-m, line 85: Could you please add a TODO to refactor these geometry conversions into geometry-converter.ts?

web/src/app/shared/converters/loi-converter/loi-data-converter.spec.ts
```diff
@@ -21,7 +21,7 @@ import {
   LocationOfInterest,
   PointOfInterest,
 } from '../../models/loi.model';
-import { toGeometry } from '../geometry-converter';
+import { toGeometry, GEOJSON_GEOMETRY_TYPES } from '../geometry-converter';
```
gino-m, line 24: We don't use a strict GeoJson format in our db, so it might be more appropriate just call this map something like `GEOMETRY_TYPES`.

gino-m (changes requested):
Thanks for the fix. Some nits for ya :)

web/src/app/shared/converters/loi-converter/loi-data-converter.spec.ts
DaoyuT: The name `GEOJSON` is to indicate that these follow the GeoJson naming. Does that make sense to you?

gino-m (approved):

PR #975, opened 2022-12-01
[DX] Improve local and hosted development workflows

### LINKED ISSUE #949: [Devtools] Setup hosting in Firebase emulator config and simplify
Currently, local testing relies on running `ng serve` independently of the emulator, using a special env "local" to connect to the emulator.  Specifically, we run `npm i && npm run emulator` from functions/  , followed by `npm i && ng server -c local` from web/. Cloud Functions don't use redirects, but rather use local config to point to localhost. This can be simplified by using Firebase emulators for both hosting and functions. @DaoyuT @os-micmec  FYI.

gino-m:
This PR introduces the following DX improvements:
- [x] Allow Cloud Functions to be tested locally together with web app
- [x] Simplified npm commands to run
- [x] Use local npm installs instead of global ones
- [x] Upgrade deploy-on-merge action to use Node 16, latest scripts

Fixes #949. 

New instructions moved to #1028.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -92,6 +92,7 @@ Run `npm run build:dev` to build the project. The build artifacts will be stored
 ## Running unit tests
 
 Run `npm run test` to execute run tests locally in Chrome using
+
 [Karma](https://karma-runner.github.io) test runner.
 
 ## Running end-to-end tests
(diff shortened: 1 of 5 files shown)


gino-m:
@os-micmec for your review as discussed. @DaoyuT @nwkotto FYI!

os-micmec (approved):

PR #993, opened 2022-12-12
[CD] Fix broken imports on continuous deployment

### LINKED ISSUE #991: Fix broken continuous deployment
https://github.com/google/ground-platform/actions/runs/3678347146

Likely broken by merging #987.

@os-micmec FYI (for later)

See also #990.

gino-m:
Fixes #991.

Build is ok but deploy still fails with:

```
Error: Failed to load function definition from source: Failed to generate manifest from function source: Error: Cannot find module 'on-create-user'
Require stack:
- /Users/gmiceli/GitProjects/gino-m/ground-platform/functions/lib/index.js
- /Users/gmiceli/GitProjects/gino-m/ground-platform/functions/node_modules/firebase-functions/lib/runtime/loader.js
- /Users/gmiceli/GitProjects/gino-m/ground-platform/functions/node_modules/firebase-functions/lib/bin/firebase-functions.js
```

Note that upgrade to latest `firebase-functions@latest` doesn't fix this.


gino-m:
So... it appears NodeJS will look for non-relative  modules in`require()` in `node_modules`, and absolute ones (`require('/path'`) relative to the absolute root. The best workaround I could find so far is to set up a module alias `@`, and to use that to indicate paths starting form our deployment root. Suggesting we merge to get our CD working again until we can come up with a better solution.

gino-m:
@nwkotto Could you PTAL if you have a sec?

DaoyuT (approved):
LGTM, sorry for the delay.

Issue #983, opened 2022-12-09
[Code health] Reorganize component directories and modules


gino-m:
Based on https://medium.com/@dolanmiu/angular-folder-structure-follow-the-dom-8af5453d0808 and discussion with nwotto@, suggesting the following:

* One component = one directory = one module.
* Subcomponent's directories should be nested under their parent component's directory
* Components used in >1 place should live in shared/.

@DaoyuT @amysorto FYI

gino-m:
I already started this, need to verify all components were reorganized.

PR #1003, opened 2022-12-15
[Code health] Refactor header widget and subcomponents

### LINKED ISSUE #985: [Code health] Refactor header into single reusable template component
`survey-header` is currently used by both `survey-list` and `main-page`, but `header-layout` is used by `sign-in-page` and `survey-list`, as well as by `survey-header` itself. These both have several dependencies. Moving these to `shared/components` drags along all of these dependencies, which can be consolidated under a single shared component if there were refactored into a more rational structure.

gino-m:
Closes #985. Towards #983.

- [x] Add current user profile pic to shared header and use in survey list page
- [x] Next shared header components in directories to mirror hierarchy

@DaoyuT PTAL?


web/src/app/components/header/current-user-widget/current-user-widget.component.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(CurrentUserWidgetComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
DaoyuT, line 45: Can we add a test for `openProfileDialog` function?
(I realized this is not a new component, the function is extracted from other component, but you add a test, this component will reach 100% coverage :D)

DaoyuT (changes requested):
LGTM overall

web/src/app/components/header/current-user-widget/current-user-widget.component.spec.ts
gino-m: I plan to do an iteration over all components one I finish reshuffling them. Ok to submit the pure refactor first?

DaoyuT (approved):

PR #1018, opened 2022-12-27
[Code health] Fix lint --fix for web and disable `bracketSpacing` lint overrides

### LINKED ISSUE #1017: [Code health] Disable gts prettier overrides
Currently, this only includes `bracketSpacing: true`. I can't think of a justification for diverging from gts here but not elsewhere, so suggesting we just stick with the gts provided defaults to make formatting more predictable and to avoid bikeshedding over style.

gino-m:
Closes #1017.

@DaoyuT PTAL?


DaoyuT (approved):
LGTM

PR #1025, opened 2022-12-29
[Code health] Reorganize main page and subcomponents into nested directories under `src/app/pages`

### LINKED ISSUE #1023: [Code health] Allow and  use relative paths in imports for components under  current dir
Inspired by suggestion by @nwkotto.

gino-m:
Towards #983.

To simplify import paths, this PR relaxes path constraints to once again allow relative paths (closes #1023). These are intended to be used for `./` imports only. Importing from higher directory levels (`..`) should still always be done with absolute paths.

ground-pin.ts is also converted into a service so it can be shared more consistently by components.

@DaoyuT PTAL?


web/src/app/components/job-list-item/job-list-item.component.ts
```diff
@@ -18,7 +18,7 @@ import {Component, Input, OnInit, OnDestroy} from '@angular/core';
 import {DialogService} from 'app/services/dialog/dialog.service';
 import {ImportDialogComponent} from 'app/components/import-dialog/import-dialog.component';
 import {Job} from 'app/models/job.model';
-import {getPinImageSource} from 'app/components/map/ground-pin';
+import {getPinImageSource} from 'app/ground-pin';
```
DaoyuT, line 21: We need to find home for `ground-pin`, it can just sit right under `app/`, maybe `assets/` (it's not a image file, but it's providing a svg dynamically)? Or somewhere else?

DaoyuT (changes requested):
LGTM overall, one nit about `ground-pin`.

Also, was it a lot of work to move a component from `components/` to `pages/`? I assume not because you can use the IDE refactoring function. I am asking because I can see some of the components need to be moved back to `components/` for reusing purpose in the future.

web/src/app/components/job-list-item/job-list-item.component.ts
gino-m: Files in `assets` are actually where static images get served from. Maybe a service is more appropriate? Trying this now..

gino-m:
@DaoyuT PTAL?

DaoyuT (approved):
LGTM

PR #1026, opened 2023-01-06
Geometry model refactor: Added supported for polygons and multipolygons

nwkotto:
- Removed references to GeoJsonLocationOfInterest and AreaOfInterest in the map component
- Added support for Polygon and Multipolygon geometry types, refactoring the map component to support multiple polygons associated with an LOI instead of just one.
- Moved the logic for generating geometry type labels from loi-panel-header to geometry.ts

### DIFF
--- a/web/src/app/components/loi-panel-header/loi-panel-header.component.ts
+++ b/web/src/app/components/loi-panel-header/loi-panel-header.component.ts
@@ -29,10 +29,9 @@ import { DialogService } from 'app/services/dialog/dialog.service';
 import { DataStoreService } from 'app/services/data-store/data-store.service';
 import { NavigationService } from 'app/services/navigation/navigation.service';
 import { Subscription } from 'rxjs';
-import { GeoJsonLocationOfInterest } from 'app/models/loi.model';
 import { LocationOfInterestService } from 'app/services/loi/loi.service';
 import { Map } from 'immutable';
-import { Point } from 'app/models/geometry/point';
+import { GeometryType, geometryTypeLabel } from 'app/models/geometry/geometry';
 @Component({
   selector: 'ground-loi-panel-header',
   templateUrl: './loi-panel-header.component.html',
@@ -47,9 +46,9 @@ export class LocationOfInterestPanelHeaderComponent
   pinUrl: SafeUrl;
   readonly loiType = LocationOfInterestType;
   subscription: Subscription = new Subscription();
-  loiTypeValue?: LocationOfInterestType;
   private readonly CAPTION_PROPERTIES = ['caption', 'label', 'name'];
   private readonly ID_PROPERTIES = ['id', 'identifier', 'id_prod'];
+  private loiGeometryType?: GeometryType;
   private loiProperties?: Map<string, string | number>;
 
   constructor(
@@ -63,11 +62,7 @@ export class LocationOfInterestPanelHeaderComponent
     this.pinUrl = sanitizer.bypassSecurityTrustUrl(getPinImageSource());
     this.subscription.add(
       loiService.getSelectedLocationOfInterest$().subscribe(loi => {
-        if (loi instanceof GeoJsonLocationOfInterest) {
-          this.loiTypeValue = LocationOfInterestType.Polygon;
-        } else if (loi.geometry instanceof Point) {
-          this.loiTypeValue = LocationOfInterestType.Point;
-        }
+        this.loiGeometryType = loi.geometry?.geometryType;
         this.loiProperties = loi.properties;
       })
     );
@@ -138,8 +133,8 @@ export class LocationOfInterestPanelHeaderComponent
     if (caption) {
       return caption;
     }
-    const loiType =
-      this.loiTypeValue === LocationOfInterestType.Point ? 'Point' : 'Polygon';
+    const loiType = geometryTypeLabel(this.loiGeometryType);
+
     const id = this.findProperty(this.ID_PROPERTIES);
     if (id) {
       return loiType + ' ' + id;
@@ -169,4 +164,5 @@ export class LocationOfInterestPanelHeaderComponent
 enum LocationOfInterestType {
   Point = 'POINT',
   Polygon = 'POLYGON',
+  MultiPolygon = 'MULTIPOLYGON',
 }
--- a/web/src/app/components/map/map.component.spec.ts
+++ b/web/src/app/components/map/map.component.spec.ts
@@ -29,14 +29,11 @@ import { NavigationService } from 'app/services/navigation/navigation.service';
 import { Survey } from 'app/models/survey.model';
 import {
   LocationOfInterest,
-  GeoJsonLocationOfInterest,
-  AreaOfInterest,
   GenericLocationOfInterest,
 } from 'app/models/loi.model';
 import { Map, List } from 'immutable';
 import { Job } from 'app/models/job.model';
 import { BehaviorSubject, of } from 'rxjs';
-import { GeoPoint } from 'firebase/firestore';
 import { GoogleMapsModule } from '@angular/google-maps';
 import { urlPrefix } from 'app/components/map/ground-pin';
 import {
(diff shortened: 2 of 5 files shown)


web/src/app/models/geometry/geometry.ts
```diff
@@ -33,3 +33,14 @@ export enum GeometryType {
   POLYGON = 'POLYGON',
   MULTI_POLYGON = 'MULTI_POLYGON',
 }
+
+/** A label for a given geometry type. Defaults to 'Polygon'. */
+export function geometryTypeLabel(geometryType?: GeometryType): string {
```
gino-m, line 38: The label is presentation logic. Can we move this into the component or into a shared service? This will also make i18n easier later.

web/src/app/models/geometry/geometry.ts
```diff
@@ -33,3 +33,14 @@ export enum GeometryType {
...
+  if (geometryType === GeometryType.POINT) {
```
gino-m, line 39: Can we use a switch statement or an immutable Map here?

gino-m (changes requested):
Great work! A few small suggestions, otherwise LG.

web/src/app/components/map/map.component.ts
```diff
@@ -411,77 +429,40 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
-    jobName: string | undefined,
-    loi: AreaOfInterest
-  ) {
-    const vertices: google.maps.LatLng[] = loi.polygonVertices.map(
-      vertex => new google.maps.LatLng(vertex.latitude, vertex.longitude)
-    );
-
-    this.addPolygonToMap(loi.id, color, jobName, [vertices]);
+    const firstPolygon = polygons && polygons[0];
+    this.panAndZoom(firstPolygon?.getPaths().getAt(0).getAt(0));
```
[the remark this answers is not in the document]
nwkotto: Sure, that works for me! I can see the advantage of focusing in on the location, but I don't think it does too much at the moment, especially since we're not focusing in on the entire multipolygon.

web/src/app/models/geometry/geometry.ts
nwkotto: Moved it back to the loi panel header component for now.

web/src/app/models/geometry/geometry.ts
nwkotto: Yeah, swapped this for a switch statement

web/src/app/components/map/map.component.ts
```diff
@@ -497,17 +476,21 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
   private onPolygonClick(event: google.maps.PolyMouseEvent) {
     if (this.disableMapClicks) {
       return;
     }
     const candidatePolygons: google.maps.Polygon[] = [];
-    for (const polygon of this.polygons.values()) {
-      if (google.maps.geometry.poly.containsLocation(event.latLng!, polygon)) {
-        candidatePolygons.push(polygon);
+    for (const loiPolygons of this.polygons.values()) {
+      for (const polygon of loiPolygons) {
```
DaoyuT, line 488: I can see one very specific corner case here. When an LOI is Multipolygon which has overlapping polygons, and user clicks into the intersection, then we would have 2 polygon candidates, but it's the same LOI, this would mess up the assumption of 1 polygon == 1 LOI by the code later.

I don't think we have to address this now and it's a very rare case, but can you add a TODO or comment to describe this situation?

web/src/app/components/loi-panel-header/loi-panel-header.component.ts
```diff
@@ -133,13 +128,25 @@ export class LocationOfInterestPanelHeaderComponent
     });
   }
 
+  /** A label for a given geometry type. Defaults to 'Polygon'. */
+  private geometryTypeLabel(geometryType?: GeometryType): string {
```
DaoyuT, line 132: I saw you added a similar function in `geometry.ts`, do you want to refer to that instead?

DaoyuT (changes requested):
Thanks for the migration and the refactoring, looks pretty good in general!

web/src/app/components/loi-panel-header/loi-panel-header.component.ts
nwkotto: I actually moved that back in here, per a request from Gino.

web/src/app/components/map/map.component.ts
```diff
@@ -411,77 +429,38 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+          strokeWeight: normalPolygonStrokeWeight,
+        });
     }
-    if (this.selectedPolygon) {
-      this.selectedPolygon.setOptions({
-        strokeWeight: normalPolygonStrokeWeight,
-      });
-    }
-    this.selectedPolygon = polygon;
-    this.panAndZoom(polygon?.getPaths().getAt(0).getAt(0));
```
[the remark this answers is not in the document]
nwkotto: Gino suggested that we remove this, since if the user clicks on the polygon, it should already be in view anyway.

web/src/app/components/map/map.component.ts
```diff
@@ -74,15 +74,14 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     streetViewControl: false,
     mapTypeId: google.maps.MapTypeId.HYBRID,
   };
-  private selectedMarker?: google.maps.Marker;
-  private selectedPolygon?: google.maps.Polygon;
+  private selectedLocationOfInterestId: string | null = null;
```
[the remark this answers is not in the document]
nwkotto: :)

web/src/app/components/map/map.component.spec.ts
```diff
@@ -305,22 +327,11 @@ describe('MapComponent', () => {
     assertMarkerIcon(marker1, jobColor1, 50);
   }));
 
-  it('should select loi when polygon is clicked', () => {
```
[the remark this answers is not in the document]
nwkotto: Ah yeah, whoops, reflexively deleted this when I saw that it was using AreaOfInterest. Added it back.

web/src/app/components/loi-panel-header/loi-panel-header.component.ts
```diff
@@ -47,9 +46,9 @@ export class LocationOfInterestPanelHeaderComponent
   pinUrl: SafeUrl;
   readonly loiType = LocationOfInterestType;
   subscription: Subscription = new Subscription();
-  loiTypeValue?: LocationOfInterestType;
   private readonly CAPTION_PROPERTIES = ['caption', 'label', 'name'];
   private readonly ID_PROPERTIES = ['id', 'identifier', 'id_prod'];
+  private loiGeometryType?: GeometryType;
```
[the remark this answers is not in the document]
nwkotto: Good catch, I missed that. Added some test coverage to catch this kind of thing in the future.

web/src/app/components/loi-panel-header/loi-panel-header.component.ts
```diff
@@ -169,4 +176,5 @@ export class LocationOfInterestPanelHeaderComponent
 enum LocationOfInterestType {
```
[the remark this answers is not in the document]
nwkotto: Removed after some fixes and improved test coverage

web/src/app/components/map/map.component.ts
nwkotto: Ah yeah, true. Added a todo.

web/src/app/components/map/map.component.ts
```diff
@@ -366,36 +384,36 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
 
   /**
-   * Selecting LOI enlarges the marker or border of the polygon,
-   * pans and zooms to the marker/polygon. Selecting null is considered
+   * Selecting LOI enlarges the marker or border of the polygon(s),
+   * pans and zooms to the marker/polygon(s). Selecting null is considered
    * as deselecting which will change the selected back to normal size.
    */
   private selectLocationOfInterest(
     selectedLocationOfInterestId: string | null
```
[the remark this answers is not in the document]
nwkotto: haha sounds good.

DaoyuT (approved):
LGTM, thank you, Nick!

gino-m (approved):
LGTM! 🚀

DaoyuT (approved):
LGTM

Issue #887, opened 2022-06-23
[Create survey] New create survey flow


gino-m:
@vittorino Could you please paste the mocks for this flow here once ready?

gino-m:
@DaoyuT This is a bit broad, closing in favor of smaller issues

PR #1027, opened 2023-01-09
Add create survey page

DaoyuT:
Screenshot 2023-01-09 at 1 43 24 PM

Towards #887

### DIFF
--- a/web/src/app/pages/create-survey/create-survey.component.html
+++ b/web/src/app/pages/create-survey/create-survey.component.html
@@ -1,0 +1,35 @@
+<!--
+  Copyright 2023 Google LLC
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<div>
+  <h1>Start building your survey</h1>
+  <form [formGroup]="formGroup">
+    <mat-form-field class="input" appearance="fill">
+      <mat-label>
+        Title
+      </mat-label>
+      <input matInput [formControlName]="titleControlKey" placeholder="Survey title" />
+    </mat-form-field>
+
+    <mat-form-field class="input" appearance="fill">
+      <mat-label>
+        Description
+      </mat-label>
+      <input matInput [formControlName]="descriptionControlKey" placeholder="Survey description" />
+    </mat-form-field>
+  </form>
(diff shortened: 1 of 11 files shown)


web/src/app/pages/create-survey/create-survey.component.html
```diff
@@ -0,0 +1,35 @@
+<!--
+  Copyright 2022 Google LLC
```
gino-m, line 2: s/2022/2023/ throughout

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -0,0 +1,91 @@
...
+      [this.titleControlKey]: '',
+      [this.descriptionControlKey]: '',
+    });
+    navigationService.init(route);
+  }
+
+  ngOnInit(): void {
+    this.navigationService.getSurveyId$().subscribe(surveyId => {
+      if (surveyId) {
+        this.surveyService.activateSurvey(surveyId);
```
gino-m, line 50: Is the survey first created as a side-effect of this call? If we create the survey only after this first step is completed, we can avoid having a bunch of abandoned surveys in the db with no name/description. That would also make creation explicit.

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -0,0 +1,91 @@
...
+      }
+      this.currentSurveyId = surveyId;
+    });
+    this.surveyService
+      .getActiveSurvey$()
+      .subscribe(survey => this.loadExistingSurvey(survey));
+  }
+
+  loadExistingSurvey(survey: Survey): void {
+    this.formGroup.controls[this.titleControlKey].setValue(survey.title);
```
gino-m, line 60: Is there a way to set this using bindings rather than explicitly calling `setValue` (e.g. `[value]="survey.title"` in the template?

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    );
+  }
+
+  back(): void {
+    this.navigationService.navigateToSurveyList();
+  }
+
+  async continue(): Promise<void> {
+    const title = this.formGroup.controls[this.titleControlKey].value;
+    const descriptoin =
```
gino-m, line 72: Spelling nit: `description`

gino-m (changes requested):
Nice work! A few small suggestions inline.

web/src/app/pages/create-survey/create-survey.component.ts
DaoyuT: Discussed offline about the benefits of using angular form.

web/src/app/pages/create-survey/create-survey.component.html
DaoyuT: Done

web/src/app/pages/create-survey/create-survey.component.ts
DaoyuT: Done

gino-m (approved):
LGTM! 🥇

Issue #1043, opened 2023-02-02
Refactor create survey page


DaoyuT:
Refactor CreateSurvey page & CreateJob page into NameSurvey component & NameJob component under CreateSurvey page.

We want the entire onboarding wizard to only have one page, it would swap the components in it instead of navigating to different pages.

gino-m:
Duplicate of #1009.

PR #1051, opened 2023-02-16
Refactor CreateSurvey page & CreateJob page into components under one page

DaoyuT:
Refactor CreateSurvey page & CreateJob page into NameSurvey component & NameJob component under CreateSurvey page.

Now the entire onboarding wizard will only have one page, it would swap the components in it instead of navigating to different pages.

#1043


web/src/app/models/survey.model.ts
```diff
@@ -14,6 +14,15 @@
  * limitations under the License.
  */
 
+export enum OnboardingStatus {
```
gino-m, line 17: Nit: Rename to`SetupPhase`?

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -40,22 +44,52 @@ describe('CreateSurveyComponent', () => {
...
   const mockSurvey = new Survey(
     surveyId,
     title,
     description,
     /* jobs= */ Map(),
-    /* acl= */ Map()
+    /* acl= */ Map(),
+    OnboardingStatus.NAME_SURVEY
+  );
+  const mockSurveyNoJob = new Survey(
```
gino-m, line 63: Nit: These aren't mocks, they're real objects, so you can remove the "mock" prefix in var names.

web/src/app/components/survey-list/survey-list.component.spec.ts
```diff
@@ -55,7 +55,8 @@ describe('SurveyListComponent', () => {
         /* tasks= */ Map()
       ),
     }),
-    /* acl= */ Map()
+    /* acl= */ Map(),
+    OnboardingStatus.DONE
```
gino-m, line 59: Can we just allow `null` here and omit this value when no in the setup flow?

web/src/app/models/survey.model.ts
```diff
@@ -30,15 +39,17 @@ export class Survey {
...
   );
 
   constructor(
     readonly id: string,
     readonly title: string,
     readonly description: string,
     readonly jobs: Map<string, Job>,
-    readonly acl: Map<string, Role>
+    readonly acl: Map<string, Role>,
+    readonly onboardingStatus: OnboardingStatus
```
gino-m, line 52: If we make this nullable (null meaning "done") it'll be backwards compatible with existing surveys, and won't require DONE to always specified .

gino-m (changes requested):

web/src/app/pages/create-survey/create-survey.component.html
```diff
@@ -13,23 +13,22 @@
...
-    </mat-form-field>
-  </form>
-  <button id="back-button" mat-stroked-button (click)="back()">Back</button>
-  <button id="continue-button" mat-flat-button color="primary" (click)="continue()">Continue</button>
-</div>
\ No newline at end of file
+<div class="container">
+  <div class="card">
+    <name-survey
+      *ngIf="isCreateSurvey()"
```
gino-m, line 20: I'd strongly prefer a separate top-level component for create and edit. The layout, primarily the page chrome, will be different. We can avoid a lot of conditional branching by having separate top-level pages.

web/src/app/models/survey.model.ts
gino-m: Even better, can we remove this is favor of `State` := `INITIAL_SETUP` vs `null`? We can refer the state of the creation flow as follows:

1) Name survey: never - survey won't be saved until a name is specified.
2) Name job: when `jobs` is `null` or empty
3) Define tasks: when there is a job, but `tasks` is null or empty
4) "Who defines" LOI screen: new flag, for ex `dataCollectionType` - `PREDEFINED`, `AD_HOC`, `HYBRID`. Show this page when `dataCollectionType` is `null`
4) Define LOIs: when `dataCollectionType` is `STRUCTURED` or `HYBRID` but lois are empty.
5) Everything is set up, but `state == INITIAL_SETUP`.

DaoyuT: SG, made `SetupPhase` a enum for the component only and changed `OnboardingStatus` to `State`.

web/src/app/pages/create-survey/create-survey.component.spec.ts
DaoyuT: Done.

web/src/app/components/survey-list/survey-list.component.spec.ts
DaoyuT: Sg, done.

web/src/app/models/survey.model.ts
DaoyuT: Done.

web/src/app/pages/create-survey/create-survey.component.html
DaoyuT: Yes, this component is for create page only, there will be another one for edit, and when adding that page I'll redirect this page to edit page if survey is in `DONE` state.

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +50,105 @@ export class CreateSurveyComponent implements OnInit {
...
+  nameSurvey: NameSurveyComponent | undefined;
+
+  private async saveSurveyTitleAndDescription(): Promise<void> {
+    const titleAndDescription = this.nameSurvey!.toTitleAndDescription();
     if (this.currentSurveyId) {
       await this.surveyService.updateTitleAndDescription(
         this.currentSurveyId,
-        title,
-        description
+        titleAndDescription[0],
```
gino-m, line 115: You can use destructuring assignment on the preceding line to explicitly unpack this array:
`const [title, description] = titleAndDescription;`

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +50,105 @@ export class CreateSurveyComponent implements OnInit {
...
+        titleAndDescription[1]
+      );
+      this.navigationService.navigateToCreateSurvey(createdSurveyId);
+    }
+  }
+
+  @ViewChild('nameJob')
+  nameJob: NameJobComponent | undefined;
```
gino-m, line 128: n00b question: can we user `?` instead of `| undefined` here?

web/src/app/models/survey.model.ts
```diff
@@ -14,6 +14,14 @@
  * limitations under the License.
  */
 
+/**
+ * Represents the state of the survey, undefined should be considered as DONE.
+ */
+export enum State {
+  INITIAL_SETUP,
+  DONE,
```
gino-m, line 22: Can we use `null` instead of `DONE` for backwards (and forwards) compatibility as discussed?

gino-m (changes requested):
A few more that I missed yesterday. Thanks!

web/src/app/converters/firebase-data-converter.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import {DocumentData, Timestamp} from '@angular/fire/firestore';
-import {Survey} from 'app/models/survey.model';
+import {Survey, State} from 'app/models/survey.model';
```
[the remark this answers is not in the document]
DaoyuT: Sg, done.

web/src/app/pages/create-survey/create-survey.component.html
```diff
@@ -13,23 +13,22 @@
...
-      </mat-label>
-      <input matInput [formControlName]="descriptionControlKey" placeholder="Survey description" />
-    </mat-form-field>
-  </form>
-  <button id="back-button" mat-stroked-button (click)="back()">Back</button>
-  <button id="continue-button" mat-flat-button color="primary" (click)="continue()">Continue</button>
-</div>
\ No newline at end of file
+<div class="container">
+  <div class="card">
```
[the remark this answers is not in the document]
DaoyuT: Yep, should be a job for ngSwitch

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +50,105 @@ export class CreateSurveyComponent implements OnInit {
...
+  }
+
+  job(): Job | undefined {
+    if (this.currentSurvey?.jobs.size ?? 0 > 0) {
+      return this.currentSurvey?.jobs.values().next().value;
+    }
+    return undefined;
+  }
+
+  isNameSurvey(): boolean {
```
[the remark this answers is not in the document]
DaoyuT: We do have it, it's called `SetupPhase` and `getSetupPhase` now, are you suggesting to rename to `CreateSurveyStep`?

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -40,22 +47,49 @@ describe('CreateSurveyComponent', () => {
...
+    /* acl= */ Map()
+  );
+  const surveyWithoutJob = new Survey(
     surveyId,
     title,
     description,
     /* jobs= */ Map(),
     /* acl= */ Map()
   );
+  const mockJob = new Job(
```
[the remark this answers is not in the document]
DaoyuT: Oops, missed this one.

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -15,32 +15,31 @@
...
   selector: 'create-survey',
   templateUrl: './create-survey.component.html',
   styleUrls: ['./create-survey.component.scss'],
 })
 export class CreateSurveyComponent implements OnInit {
-  readonly titleControlKey = 'title';
-  readonly descriptionControlKey = 'description';
-  private currentSurveyId: string | null = null;
-  formGroup: FormGroup;
+  currentSurveyId: string | null = null;
```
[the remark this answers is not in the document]
DaoyuT: Yes, we should use `string | undefined` more than `string | null`, but this is coming from `getSurveyId$` in `NavigationService`, and it has `| null` in a bunch of other places. We should clean this up later, it would be out of the scope of this PR.

web/src/app/pages/create-survey/create-survey.component.ts
DaoyuT: 👍

web/src/app/pages/create-survey/create-survey.component.ts
DaoyuT: Oh, I tried `nameJob: NameJobComponent?;` and it did not compile, and it should be `nameJob?: NameJobComponent;`. Done.

web/src/app/models/survey.model.ts
DaoyuT: As stated in the comments, `undefined` should be considered as `DONE`. I like to have both `undefined` and `DONE` that both means `DONE`, we should use `DONE` explicitly and allow `undefined` in the meantime, also the enum `DONE` serves as an anchor for `undefined` to related to.

gino-m: The problem is that we don't know yet what future states will be, but it's fairly certain "DONE" will be confusing once we introduce new ones. In general it's more robust to have only one way to represent a particular state. Did you want this for use in switch statements? In those cases you could also use the default branch of switch to cover this case.

DaoyuT: No, this is not for template, we already agreed on extracting what's for UI and not store that in the backend.

Why is `DONE` confusing? To me, having to remember `undefined` means `DONE` while there is only one state in this enum called `INITIAL_SETUP ` is not straightforward.

web/src/app/pages/create-survey/create-survey.component.html
```diff
@@ -13,23 +13,22 @@
...
-        Description
-      </mat-label>
-      <input matInput [formControlName]="descriptionControlKey" placeholder="Survey description" />
-    </mat-form-field>
-  </form>
-  <button id="back-button" mat-stroked-button (click)="back()">Back</button>
-  <button id="continue-button" mat-flat-button color="primary" (click)="continue()">Continue</button>
-</div>
\ No newline at end of file
+<div class="container" [ngSwitch]="setupPhase">
```
nwkotto, line 17: Super nit, but why is ngSwitch on the container and not the card? Seems like it should be the immediate parent of the things being switched, at least for readability.

web/src/app/models/survey.model.ts
gino-m: That's also true. But it may be misleading or unclear that the survey is in a "Done" state since it's specific to the survey creation flow. I wonder if it's not just simpler and clearer to only show the create flow when one of the required elements in survey are missing? If someone leaves when they're on the "review" page, they'd just get bounced to the "edit" flow, which isn't necessarily the wrong thing to do either. My thinking here is I'm trying to avoid needing the data model to reflect anything specific to the creation flow, but rather having the creation flow react to the data in the survey. Curious to hear what you think.

web/src/app/pages/create-survey/name-job/name-job.component.ts
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Component, Input, OnChanges} from '@angular/core';
+import {FormGroup, FormBuilder} from '@angular/forms';
+
+@Component({
+  selector: 'name-job',
+  templateUrl: './name-job.component.html',
```
nwkotto, line 22: This name feels a little clunky to me, and it seems like it restricts it pretty narrowly to the job name. Maybe something like job-details-form instead?

web/src/app/pages/create-survey/name-survey/name-survey.component.ts
```diff
@@ -0,0 +1,56 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Component, Input, OnChanges} from '@angular/core';
+import {FormGroup, FormBuilder} from '@angular/forms';
+
+@Component({
+  selector: 'name-survey',
```
nwkotto, line 21: Ditto on the naming of the component. This is a little difficult to parse, and feels restrictive. Maybe something like survey-details-form instead?

nwkotto (dismissed):

web/src/app/models/survey.model.ts
DaoyuT: Chatted over GVC, we decided to remove the `SurveyState` and judge whether the survey is still `INITIAL_SETUP` or not based on the content of the survey.

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -89,95 +128,166 @@ describe('CreateSurveyComponent', () => {
...
-  it('navigates to survey list page when back button is clicked', () => {
-    const backButton = fixture.debugElement.query(By.css('#back-button'))
-      .nativeElement as HTMLElement;
-    backButton.click();
+  describe('Name Survey', () => {
+    beforeEach(() => {
+      component.setupPhase = SetupPhase.NAME_SURVEY;
+    });
+
+    it('creates new survey with title and description after clicking continue when no current survey', fakeAsync(() => {
```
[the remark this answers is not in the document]
DaoyuT: Sg, we can try this.

web/src/app/pages/create-survey/create-survey.component.html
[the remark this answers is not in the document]
DaoyuT: Good catch, I meant to added in card div, my mistake

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -89,95 +128,166 @@ describe('CreateSurveyComponent', () => {
...
+      expect(surveyServiceSpy.createSurvey).toHaveBeenCalledOnceWith(
+        newTitle,
+        newDescription
+      );
+      expect(
+        navigationServiceSpy.navigateToCreateSurvey
+      ).toHaveBeenCalledOnceWith(newSurveyId);
+    }));
+
+    it('updates title and description after clicking continue when given current survey', fakeAsync(() => {
```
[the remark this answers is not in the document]
DaoyuT: Done.

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -89,95 +128,166 @@ describe('CreateSurveyComponent', () => {
...
+      fixture.detectChanges();
+
+      const newTitle = 'newTitle';
+      const newDescription = 'newDescription';
+      const nameSurveyComponent = component.nameSurvey!;
+      nameSurveyComponent.formGroup.controls[
+        nameSurveyComponent.titleControlKey
+      ].setValue(newTitle);
+      nameSurveyComponent.formGroup.controls[
+        nameSurveyComponent.descriptionControlKey
```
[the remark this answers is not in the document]
DaoyuT: I tried to grab the `input` by css selectors and send keys into it but for some reason, that did not work...

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -89,95 +128,166 @@ describe('CreateSurveyComponent', () => {
...
+      surveyId$.next(surveyId);
+      activeSurvey$.next(surveyWithoutTitle);
+      tick();
+      fixture.detectChanges();
+
+      const newTitle = 'newTitle';
+      const newDescription = 'newDescription';
+      const nameSurveyComponent = component.nameSurvey!;
+      nameSurveyComponent.formGroup.controls[
+        nameSurveyComponent.titleControlKey
```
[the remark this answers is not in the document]
DaoyuT: I tried to grab the input by css selectors and send keys into it but for some reason, that did not work...

web/src/app/pages/create-survey/name-job/name-job.component.ts
DaoyuT: Sg, renamed to `job-details` to keep it short.

web/src/app/pages/create-survey/name-survey/name-survey.component.ts
DaoyuT: Sg, renamed to `survey-details` to keep it short.

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +51,97 @@ export class CreateSurveyComponent implements OnInit {
...
+    const name = this.nameJob!.toJobName();
+    if (this.currentSurvey!.jobs.size > 0) {
+      const existingJob = this.currentSurvey!.jobs.values().next().value;
+      await this.jobService.addOrUpdateJob(
+        this.currentSurveyId!,
+        existingJob.copyWith({name})
+      );
+    } else {
+      const newJob = this.jobService.createNewJob();
+      await this.jobService.addOrUpdateJob(
```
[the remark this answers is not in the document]
DaoyuT: Sg, done.

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +51,97 @@ export class CreateSurveyComponent implements OnInit {
...
+    }
+  }
+
+  @ViewChild('nameJob')
+  nameJob?: NameJobComponent;
+
+  private async saveJobName(): Promise<void> {
+    const name = this.nameJob!.toJobName();
+    if (this.currentSurvey!.jobs.size > 0) {
+      const existingJob = this.currentSurvey!.jobs.values().next().value;
```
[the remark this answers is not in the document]
DaoyuT: It's not checked anywhere, I'm not sure what we should do if there is more than one at this point, it shouldn't happen realistically. Added a comment explaining

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +51,97 @@ export class CreateSurveyComponent implements OnInit {
...
         description
       );
-      this.navigationService.navigateToCreateJob(this.currentSurveyId);
     } else {
       const createdSurveyId = await this.surveyService.createSurvey(
         title,
         description
       );
-      this.navigationService.navigateToCreateJob(createdSurveyId);
+      this.navigationService.navigateToCreateSurvey(createdSurveyId);
```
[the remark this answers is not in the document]
DaoyuT: Sg, extracted this line.

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -89,95 +128,166 @@ describe('CreateSurveyComponent', () => {
...
 
-    expect(navigationServiceSpy.navigateToSurveyList).toHaveBeenCalled();
+    it('navigates to survey list page when back button is clicked', () => {
+      clickBackButton(fixture);
+
+      expect(navigationServiceSpy.navigateToSurveyList).toHaveBeenCalled();
+    });
+  });
+
+  describe('Name Job', () => {
```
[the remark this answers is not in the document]
DaoyuT: Added describes, punted on typing the values

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -51,39 +51,97 @@ export class CreateSurveyComponent implements OnInit {
...
-      .getActiveSurvey$()
-      .subscribe(survey => this.loadExistingSurvey(survey));
+    this.surveyService.getActiveSurvey$().subscribe(survey => {
+      this.currentSurvey = survey;
+      this.setupPhase = this.getSetupPhase(survey);
+    });
+  }
+
+  private getSetupPhase(survey: Survey): SetupPhase {
+    if (survey.title && survey.title.trim().length > 0) {
```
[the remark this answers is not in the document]
DaoyuT: Good idea, done.

gino-m (approved):
LGTM! ✅

nwkotto (approved):
Changes look good. Thanks!

PR #1066, opened 2023-03-01
Moved share dialog to components/ and added share button to new create flow

npower:
#1054


web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -277,37 +295,33 @@ describe('CreateSurveyComponent', () => {
       }));
     });
 
-    describe('when active survey has a job', () => {
```
DaoyuT, line 280: Did you meant to delete this test case?

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -277,37 +295,33 @@ describe('CreateSurveyComponent', () => {
...
+      tick();
+      fixture.detectChanges();
+    }));
+
+    it('goes back to job details component after back button is clicked', () => {
+      clickBackButton(fixture);
+
+      expect(component.setupPhase).toBe(SetupPhase.JOB_DETAILS);
+
+      // TODO: figure out why these are not suceeding
```
DaoyuT, line 320: qq: have you tried `tick()` & `flush()` after importing those dialog dependencies into this test file?

DaoyuT (commented):
LGTM overall, just a few nits.

web/src/app/pages/create-survey/create-survey.component.spec.ts
npower: With the new logic, "when the active survey has a job" the setupPhase will be the Review page which breaks this test

'when active survey and job defined' above covers what happens now (I'm not sure what the protocol is for tests broken by intermediate states of implementation)

web/src/app/pages/create-survey/create-survey.component.spec.ts
npower: I tried both and the test still fails

web/src/app/pages/create-survey/create-survey.component.spec.ts
DaoyuT: I see, can we bring this test back and remove the newly added condition in `getSetupPhase` function?

It's a little hard to imagine with some missing phases, but I think the idea logic for `getSetupPhase` should be:
No survey/no survey title --> SURVEY_DETAILS
Survey has title/survey has a job --> JOB_DETAILS
Survey has a job without LOI --> DEFINE_LOIS
Survey has a job with LOI but without tasks --> DEFINE_TASKS
Survey has a job with LOI and tasks --> redirect to edit survey page (I am working on this)

So technically, this `getSetupPhase` function is used when user did not finish the initial setup and quit, and we can bring them to where they were left off. Gino and I had some discussion and decided that we don't have to bring them back to the REVIEW phase if they left at the REVIEW phase last time. So you can remove that chunk in `getSetupPhase`. Sorry, should have caught that earlier.

npower: this should be fixed now!

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -225,6 +229,21 @@ describe('CreateSurveyComponent', () => {
     });
   });
 
+  describe('when active survey has finished setup', () => {
```
DaoyuT, line 232: This is repeated.

DaoyuT (approved):

DaoyuT (dismissed):
LGTM

DaoyuT (approved):
LGTM

PR #1075, opened 2023-03-13
Redirect from create survey page to edit survey page if finished

DaoyuT:
Redirect from create survey page to edit survey page if the setup is finished.

### DIFF
--- a/web/src/app/pages/create-survey/create-survey.component.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.ts
@@ -62,7 +62,6 @@ export class CreateSurveyComponent implements OnInit {
   }
 
   private isSetupFinished(survey: Survey): boolean {
-    // To make it simple we are not checking the LOIs here since defining tasks is the step after defining LOIs.
     return this.hasTitle(survey) && this.hasJob(survey) && this.hasTask(survey);
   }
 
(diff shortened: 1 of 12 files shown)


web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -52,18 +52,38 @@ export class CreateSurveyComponent implements OnInit {
...
+      if (this.isSetupFinished(survey)) {
+        this.navigationService.navigateToEditSurvey(survey.id);
+        return;
+      }
       this.currentSurvey = survey;
       this.setupPhase = this.getSetupPhase(survey);
     });
   }
 
+  private isSetupFinished(survey: Survey): boolean {
```
nwkotto, line 64: What about LOIs?

web/src/app/services/navigation/navigation.service.ts
```diff
@@ -236,6 +237,11 @@ export class NavigationService {
     this.router.navigateByUrl(url, {replaceUrl: true});
   }
 
+  navigateToEditSurvey(surveyId: string): void {
+    const url = `${NavigationService.SURVEYS_SEGMENT}/${surveyId}/${NavigationService.SURVEYS_EDIT}`;
```
nwkotto, line 241: Nit: Maybe we should create a utility file that we can share between the navigation service and the route configs to ensure that the routes defined here and there are consistent.

web/src/app/pages/create-survey/create-survey.component.ts
DaoyuT: We don't need to since task is the next step of LOI, so I wanted to make it simple and only check tasks. I added a comment to explain why.

web/src/app/services/navigation/navigation.service.ts
DaoyuT: Sg, added a test.

nwkotto (approved):

PR #1087, opened 2023-04-24
Disable drawing tools for now.

DaoyuT:


web/src/app/pages/main-page-container/main-page/map/map.component.ts
```diff
@@ -152,6 +157,10 @@ export class MapComponent implements AfterViewInit, OnDestroy {
     if (this.disableMapClicks) {
       return;
     }
+    if (!this.shouldEnableDrawingTools) {
```
nwkotto, line 160: I think a slightly cleaner way you could do this is to set editMode to None if drawing tools are not enabled. So you could use a ternary expression below to set to none if drawing tools not enabled or get edit modes otherwise.

web/src/app/pages/main-page-container/main-page/map/map.component.html
```diff
@@ -18,28 +18,31 @@
   height="100%"
   width="100%"
   [options]="mapOptions"
-  (mapClick)="onMapClick($event)">
+  (mapClick)="onMapClick($event)"
+>
```
nwkotto, line 22: Do we want these changes to be included? Was this an accident?

nwkotto (approved):

web/src/app/pages/main-page-container/main-page/map/map.component.html
DaoyuT: It came from the linter.

web/src/app/pages/main-page-container/main-page/map/map.component.ts
DaoyuT: Sorry, I merged before I saw this. This sounds good, I'll send another PR to address this.

PR #1089, opened 2023-04-27
Npower/1081/initialize map satellite hybrid

npower:
Turned on mapTypeControl so that users can switch to "Satellite" mode, because setting the map to hybrid doesn't seem to work for everyone. 

Before change:
77cQFKShqbjRGm8

After change:
7MaZ6hJJYyaAdz7

After change, with 'Satellite' chosen:
7EASKkWXELFG6rT


DaoyuT (approved):
LGTM

PR #1091, opened 2023-05-17
Select LOI when LOI is clicked in side panel

DaoyuT:
and highlight LOI in side panel when selected

#1073

Screenshot 2023-05-17 at 12 18 49 PM
The highlighting style is only temporary, it will be finalized later along with other parts

### DIFF
--- a/web/src/app/components/job-list-item/job-list-item.component.spec.ts
+++ b/web/src/app/components/job-list-item/job-list-item.component.spec.ts
@@ -23,6 +23,7 @@ import {
   MatTreeHarness,
   MatTreeNodeHarness,
 } from '@angular/material/tree/testing';
+import {MatButtonHarness} from '@angular/material/button/testing';
 import {DataStoreService} from 'app/services/data-store/data-store.service';
 import {JobListItemComponent} from './job-list-item.component';
 import {MatIconModule} from '@angular/material/icon';
@@ -34,7 +35,6 @@ import {MatDialogModule} from '@angular/material/dialog';
 import {Router} from '@angular/router';
 import {MatTreeModule} from '@angular/material/tree';
 import {CdkTreeModule} from '@angular/cdk/tree';
-import {By} from '@angular/platform-browser';
 import {SubmissionService} from 'app/services/submission/submission.service';
 import {AuditInfo} from 'app/models/audit-info.model';
 import {Submission} from 'app/models/submission/submission.model';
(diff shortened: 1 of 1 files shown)


web/src/app/components/job-list-item/job-list-item.component.spec.ts
```diff
@@ -238,4 +251,32 @@ describe('JobListItemComponent', () => {
...
   });
+
+  it('should highlight LOI when it is selected', () => {
+    const lois = createLois(1);
+    lois$.next(lois);
+
+    locationOfInterestId$.next(lois.first()!.id);
+    fixture.detectChanges();
+
+    const loiNodeElement = fixture.debugElement.query(By.css('mat-tree-node'));
```
amysorto, line 262: It's best to use the tree test harness here instead. The following should work instead:
```
const loiNodeElement = await loader.getHarness(MatTreeNodeHarness);
expect (await (await loiNodeElement.host()).hasClass('tree-node-selected');
```

web/src/app/components/job-list-item/job-list-item.component.spec.ts
```diff
@@ -238,4 +251,32 @@ describe('JobListItemComponent', () => {
...
+      jasmine.objectContaining({'tree-node-selected': true})
+    );
+  });
+
+  it('should select LOI when LOI is clicked', () => {
+    const lois = createLois(1);
+    lois$.next(lois);
+    const loiId = lois.first()!.id;
+
+    const selectLoiButton = fixture.debugElement.query(
```
amysorto, line 273: For this line, it can be replaced with test harnesses:
```
const selectLoiButton = await loader.getHarness(MatButtonHarness.with({selector: '#select-loi'}));
```

web/src/app/components/job-list-item/job-list-item.component.spec.ts
```diff
@@ -238,4 +251,32 @@ describe('JobListItemComponent', () => {
...
+      By.css('mat-tree-node #select-loi')
+    ).nativeElement as HTMLElement;
+    selectLoiButton.click();
```
amysorto, line 276: this would become `await selectLoiButton.click();` once the button is a test harness

amysorto (approved):
LGTM, the only comments are to use test harnesses for the tests: https://material.angular.io/guide/using-component-harnesses#benefits-of-component-test-harnesses.

web/src/app/components/job-list-item/job-list-item.component.spec.ts
DaoyuT: This is exactly what I needed, and I don't need that `fixture.detectChanges();` no more. Great tip! Thanks! Just had to add that `toBeTrue()` at the end, otherwise jasmine gives me a warning.

web/src/app/components/job-list-item/job-list-item.component.spec.ts
DaoyuT: Great tip! Applied, thanks!

web/src/app/components/job-list-item/job-list-item.component.spec.ts
DaoyuT: Done.

amysorto (approved):

PR #1093, opened 2023-05-22
Npower/1079/survey list styling

npower:
Updated the styling of the survey list page to match the mock. Also changed the "Start new survey" card to open the new create survey page`/surveys/create`

4YReKT5QdazWwkv


DaoyuT (approved):
Thanks for the PR, LGTM

PR #1100, opened 2023-06-12
Add submission detail panel

amysorto:
Will currently work for text based questions

Here is a screenshot of this PR changes if you click on a submission:
image


web/src/app/pages/main-page-container/main-page/main-page.component.html
```diff
@@ -20,12 +20,13 @@
...
     <mat-sidenav-container class="main-page-container">
       <mat-sidenav-content class="mat-sidenav-content">
+        <submission-detail-panel *ngIf="showSubmissionPanel"
+          [submission]="selectedSubmission"
+          (closePanel)="closeSelectedSubmissionDetailPanel()">
+        </submission-detail-panel>
         <ground-map
           [shouldEnableDrawingTools]="shouldEnableDrawingTools"
         ></ground-map>
-        <div *ngIf="shouldEnableDrawingTools">
```
DaoyuT, line 26: Is there a reason to delete this?

I believe `shouldEnableDrawingTools` is currently set to `false` because it's temporary disabled and this shouldn't affect others.

web/src/app/components/job-list-item/job-list-item.component.ts
```diff
@@ -201,7 +201,12 @@ export class JobListItemComponent implements OnInit, OnDestroy {
   }
 
   selectLoi(node: DynamicFlatNode) {
-    this.navigationService.selectLocationOfInterest(node.loi!.id);
+    if (node.loi) {
```
DaoyuT, line 204: If I understand correctly if `node` has `loi`, that means it's a `loi node` and thus we select LOI, right? If so, can we do like `if (isLoiNode(node))`? It's a very small change, I think it helps the reader understand how DynamicFlatNode works.

Same thing for below.

DaoyuT (changes requested):
Looks good overall, just a few nits, good work!

web/src/app/pages/main-page-container/main-page/main-page.component.html
amysorto: Added it back!

DaoyuT (approved):
LGTM

DaoyuT (approved):
LGTM

PR #1101, opened 2023-06-16
Fix theming to make sharing styles more consistent

amysorto:
Theming allows you to share color and typography related styles across the app to make things consistent. This PR consists of the following changes:
* Create new theme file that uses updated theming syntax
* Move global styles and variables to the theme file
* For each component that has typography and color related styles, make a new file to place those styles in with the name of `_componentname.component-theme.scss`. These files are formatted specifically as listed in the [material docs](https://material.angular.io/guide/theming-your-components) so that they can be imported into the theme file. For each file, these are the more specific changes:
  * Use theme color palette when possible
  * Comment M3 variable names for future changes. Theming currently does not support M3 so those colors are not in the palette.
  * Add a TODO for any color that does not map to the color palette or M3
  * Use application typography when applicable to make sure the app is on the same font
  * Add a TODO comment to create [typography levels](https://material.angular.io/guide/typography#define-a-level) to further standarize typography across the application.
  * Fix any visual differences if needed (for example, sidenav color was not matching the header color anymore)
* Change any `css` files I encountered to `scss` files

This PR has a lot of file changes, but most of them are just moving color and typography related styles into a new file. 

Below are some screenshots to help understand to visual changes:

Before
image

After
image

Before
image

After
image


DaoyuT (approved):
Looks great, thank you, Amy! I've been wanting to do this but did not have the knowledge. And thanks for the detailed description, it explained a lot.

PR #1102, opened 2023-06-23
Add left nav menu to edit survey page.

DaoyuT:
Screenshot 2023-06-23 at 3 38 44 PM
Screenshot 2023-06-23 at 3 38 58 PM

### DIFF
--- a/web/src/app/pages/edit-survey/edit-survey.component.html
+++ b/web/src/app/pages/edit-survey/edit-survey.component.html
@@ -20,9 +20,9 @@
   <div class="container">
     <mat-nav-list class="left-menu">
       <a mat-list-item id="survey-button" [routerLink]="['./survey']">
-        <mat-icon class="material-symbols-outlined button-icon"
-          >content_paste</mat-icon
-        >
+        <mat-icon class="material-symbols-outlined button-icon">
+          content_paste
+        </mat-icon>
         <span>Survey</span>
       </a>
 
@@ -39,7 +39,7 @@
         mat-list-item
         [routerLink]="jobRouterLink(job.id)"
       >
-        <span matLine>{{ job.name }} extra extra extra</span>
+        <span matLine>{{ job.name }}</span>
         <button mat-icon-button (click)="onMenu($event)">
           <mat-icon>more_vert</mat-icon>
         </button>
--- a/web/src/app/pages/edit-survey/edit-survey.component.html
+++ b/web/src/app/pages/edit-survey/edit-survey.component.html
@@ -14,4 +14,40 @@
   limitations under the License.
 -->
 
-<h1 id="temp-title">Edit survey {{ (activeSurvey$ | async)?.id }}</h1>
+<div class="page">
+  <ground-header> </ground-header>
+
+  <div class="container">
+    <mat-nav-list class="left-menu">
+      <a mat-list-item id="survey-button" [routerLink]="['./survey']">
+        <mat-icon class="material-symbols-outlined button-icon"
+          >content_paste</mat-icon
+        >
+        <span>Survey</span>
+      </a>
+
+      <a mat-list-item id="share-button" [routerLink]="['./survey']">
+        <mat-icon class="material-symbols-outlined button-icon">group</mat-icon>
+        <span>Sharing</span>
+      </a>
+
+      <h3 matSubheader class="item job-label">Jobs</h3>
+
+      <a
+        *ngFor="let job of jobs(); index as i"
+        id="job-{{ i }}"
+        mat-list-item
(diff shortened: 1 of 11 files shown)


web/src/app/pages/edit-survey/edit-survey.component.spec.ts
```diff
@@ -105,10 +135,32 @@ describe('EditSurveyComponent', () => {
       fixture.detectChanges();
     }));
 
-    it('displays survey id in title', () => {
-      const tempTitle = fixture.debugElement.query(By.css('#temp-title'))
-        .nativeElement.innerText;
-      expect(tempTitle).toContain(surveyId);
+    it('displays left menu items with router link', () => {
+      const surveyButton = fixture.debugElement.query(By.css('#survey-button'))
```
nwkotto, line 139: Nit: Could you use queryAll and map to get these labels and links with a little less repetition?

Or parametrizing could also be the way to go. Internally we have the concept of "pit" (parametrized it), though not sure if that exists out in the wild. There's also the old fashioned way of parametrizing: https://jdkandersson.com/2019/06/30/parametrize-angular-jasmine-tests/

But I do feel that there are too many expects here for one test. I think it would be cleaner to have separate it blocks for each of the links, one way or another.

web/src/app/pages/edit-survey/edit-survey.component.spec.ts
```diff
@@ -65,7 +89,13 @@ describe('EditSurveyComponent', () => {
     activeSurvey$ = new Subject<Survey>();
     surveyServiceSpy.getActiveSurvey$.and.returnValue(activeSurvey$);
 
+    const routes = [
```
nwkotto, line 92: Are these routes being used for anything in these tests? Do they populate the router links or something?

web/src/app/pages/edit-survey/edit-survey.component.html
```diff
@@ -14,4 +14,40 @@
...
 
-<h1 id="temp-title">Edit survey {{ (activeSurvey$ | async)?.id }}</h1>
+<div class="page">
+  <ground-header> </ground-header>
+
+  <div class="container">
+    <mat-nav-list class="left-menu">
+      <a mat-list-item id="survey-button" [routerLink]="['./survey']">
+        <mat-icon class="material-symbols-outlined button-icon"
+          >content_paste</mat-icon
```
nwkotto, line 24: Nit: This html formatting is a little funky -- is there a reason for this?

web/src/app/pages/edit-survey/edit-survey.component.html
```diff
@@ -14,4 +14,40 @@
...
+
+      <h3 matSubheader class="item job-label">Jobs</h3>
+
+      <a
+        *ngFor="let job of jobs(); index as i"
+        id="job-{{ i }}"
+        mat-list-item
+        [routerLink]="jobRouterLink(job.id)"
+      >
+        <span matLine>{{ job.name }} extra extra extra</span>
```
nwkotto, line 42: extra extra extra?

nwkotto (commented):
Nice! One thing is that it looks like you're not testing that the correct components are getting loaded into the router outlets, unless I'm missing that. Could you add that, if it's not too difficult?

web/src/app/pages/edit-survey/edit-survey.component.html
DaoyuT: Ooops, thanks for pointing out!

web/src/app/pages/edit-survey/edit-survey.component.html
DaoyuT: This was what the formatter gave me, I made `content_paste` its own line, now it looks better.

web/src/app/pages/edit-survey/edit-survey.component.spec.ts
DaoyuT: It was used by the `RouterTestingModule`, I thought I need that for the test environment, but it's not actually necessary, I removed it.

web/src/app/pages/edit-survey/edit-survey.component.spec.ts
DaoyuT: Done. I like `pit`, it looks neat, thanks!

DaoyuT:
> Nice! One thing is that it looks like you're not testing that the correct components are getting loaded into the router outlets, unless I'm missing that. Could you add that, if it's not too difficult?

Thanks for reviewing! I gave it a try, but could not verify component in the router-outlet. I think asserting the router link is good enough since we can trust the Angular router to do its job.

nwkotto (approved):
Good stuff!

Issue #934, opened 2022-08-25
[Sign-in] Cannot sign in after recovery from "quota exceeded"


gino-m:
Accounts created while Firebase is blocked due to quota limits being reached become unusable. This is likely because the `onCreateUser` Cloud Function trigger is only called when the user's Firebase account is created. If the db in inaccessible at that time, the trigger fails, and no profile is created under `users/`. 

To fix this, we should update the user's profile on each sign in. This will also fix the issue of the user's profile never being updated after first sign in that current leads to stale user full name, profile pic, or other data. A [new?] trigger `beforeSignIn` can be used for this purpose.

@atalekar FYI

PR #1104, opened 2023-07-05
[WIP] Invoke new Callable Cloud Function on sign-in to update user profile in db

gino-m:
Towards #934.


functions/src/profile-refresh.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright 2020 Google LLC
```
shobhitagarwal1612, line 2: nit: 2023

shobhitagarwal1612 (dismissed):

Issue #1053, opened 2023-02-16
Integrate task editor into "survey edit" flow


os-micmec:

gino-m:
Duplicate of #1052

PR #1114, opened 2023-07-18
Add logic to retrieve correct setup phase

os-micmec:
Now that we have the task definition component, we need to incorporate that into retrieving the right setup phase.

Towards #1053


web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -36,13 +38,17 @@ import {ShareSurveyComponent} from './share-survey/share-survey.component';
 export class CreateSurveyComponent implements OnInit {
   currentSurveyId: string | null = null;
   currentSurvey?: Survey;
+  // TODO: when we refresh, the setupPhase below is always displayed for a split of a second.
```
gino-m, line 41: Please add issue number in TODO(#..).

gino-m (changes requested):
LG, just maybe revert the two package-lock.json files which I think got updated accidentally.

web/src/app/pages/create-survey/create-survey.component.ts
gino-m: Gentle ping!

os-micmec: Done!

gino-m (approved):
Thanks for the fix!

os-micmec:
Also removed the package-lock files

PR #1134, opened 2023-08-25
Fix incorrect ordering of X and Y coordinates

### LINKED ISSUE #1133: Imported polygons show up in wrong location
It appears X and Y are reversed

gino-m:
Fixes #1133.

Before:
~Latitude = X, Longitude = Y~
image

After:
Latitude = Y, Longitude = X
image

@nwkotto PTAL?

### DIFF
--- a/web/src/app/converters/geometry-converter.spec.ts
+++ b/web/src/app/converters/geometry-converter.spec.ts
@@ -25,6 +25,7 @@ import {toGeometry, GEOMETRY_TYPES} from './geometry-converter';
 import {GeoPoint} from 'firebase/firestore';
 import {GeometryType} from 'app/models/geometry/geometry';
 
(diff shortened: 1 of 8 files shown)


nwkotto (changes requested):
Are there no tests that need to be updated to make this change? Or if not, can you add a simple test or two? I can also take that on if you don't have time.

gino-m:
> [Aug 25, 2023 ](#pullrequestreview-1596290825)
> [ View reviewed changes ](/google/ground-platform/pull/1134/files/9eb7a3b37dd3e794855e8224138dc03dcebdbd33)

Sure. I fixed the existing tests, but left out creating a test for LoiService since that functionality isn't currently used. We can revisit once we introduce interactive map edits.

nwkotto (approved):

Issue #1139, opened 2023-09-01
Update Angular and migrate to MDC based components


amysorto:
As a part of the first step of #1135, the app needs to be updated to Angular v16 (latest major version). This is so we can use the new MDC based components that will support M3 in a future version. I am splitting this work out into a separate issue since it's relelvant to do right now.

The following steps to get to this state are:
- [x] Replace usages of [flex-layout](https://github.com/angular/flex-layout) with CSS flexbox instead to unblock upgrade since it's [deprecated in v15](https://blog.angular.io/modern-css-in-angular-layouts-4a259dca9127).
- [x] Update Angular and Angular material from [v14 (current) to v15](https://update.angular.io/?v=14.0-15.0).
- [x] Run [schematic for migrating to MDC based components](https://rc.material.angular.io/guide/mdc-migration#2-run-the-migration-tool)
  - [x] button
  - [x] card
  - [x] checkbox
  - [x] dialog
  - [x] form field, input, select, autocomplete (must be migrated together)
  - [x] list
  - [x] menu
  - [x] progress bar
  - [x] progress spinner
  - [x] radio
  - [x] slide toggle
  - [x] snack bar
  - [x] tooltip
- [x] Verify visual changes and fix any issues
- [ ] Update Angular and Angular material from [v15 to v16 (most recent major version)](https://update.angular.io/?v=15.0-16.0)

gino-m:
@jcqli I think the last task "Update Angular and Angular material from [v15 to v16 (most recent major version)](https://update.angular.io/?v=15.0-16.0)" is blocked on replacing the 3rd party upload component we use, since last I checked it wasn't compatible with newer deps. Could you kindly split that one off into its own bug and mark this as done?

amysorto:
I think this issue can be closed! The non MDC components were removed in V17, which is the version we are on now

gino-m:
@amysorto Whoohoo!!

PR #1143, opened 2023-09-01
Migrate button, card, and checkbox to MDC based components

amysorto:
Part of #1139.

The TODOs added in this PR will be addressed once all the components are migrated.


gino-m (approved):
LGTM. Thanks for moving us forward!

PR #1155, opened 2023-09-05
1141 - Remove "=s" suffix from profile photo url

### LINKED ISSUE #1141: Photo URLs are stored to db with resize param
This makes it insufficient to append =s<size> . We can update `mergeUserProfile` in `datastore.ts` to remove this suffix.

rfontanarosa:
close #1141

Screenshot 2023-09-05 alle 21 58 01
Screenshot 2023-09-05 alle 21 57 34

### DIFF
--- a/functions/src/common/datastore.ts
+++ b/functions/src/common/datastore.ts
@@ -61,10 +61,13 @@ export class Datastore {
    * These attributes are merged with other existing ones if already present.
    */
   async mergeUserProfile(user: functions.auth.UserRecord) {
-    const { uid, email, displayName, photoURL } = user;
+    const {uid, email, displayName, photoURL} = user;
     await this.db_
       .doc(`users/${uid}`)
-      .set({ email, displayName, photoURL }, { merge: true });
+      .set(
+        {email, displayName, photoURL: photoURL?.replace(/=s..-[c]$/g, '')},
+        {merge: true}
+      );
   }
 
   async fetch_(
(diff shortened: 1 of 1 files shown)


functions/src/common/datastore.ts
```diff
@@ -61,10 +61,13 @@ export class Datastore {
    * These attributes are merged with other existing ones if already present.
    */
   async mergeUserProfile(user: functions.auth.UserRecord) {
-    const { uid, email, displayName, photoURL } = user;
+    const {uid, email, displayName, photoURL} = user;
     await this.db_
       .doc(`users/${uid}`)
-      .set({ email, displayName, photoURL }, { merge: true });
+      .set(
+        {email, displayName, photoURL: photoURL?.replace(/=s..-[c]$/g, '')},
```
gino-m, line 68: Please move `photoURL?.replace(/=s..-[c]$/g, '')` into its own function (e.g., `trimPhotoResizeSuffix`) and add a note explaining what the regex will trim.

Also, I just tested and the `-c` part isn't required, and the size doesn't necessarily need to be two digits. Since we're not sure of the exact syntax, perhaps we can trim more aggressively with `/=s.*$/`? The preceding ID appears to be alphanumeric only, so I think we're probably ok to assume "=" won't appear in the main part of the URL.

gino-m (changes requested):
Thanks! Two suggestions, otherwise LG!

functions/src/common/datastore.ts
rfontanarosa: added a static function named trimPhotoURLSizeSuffix with the following comment:

`  /**
   * Removes a possible size suffix from the user's photo URL.
   *
   * @param photoURL The user's photo URL.
   */`

gino-m (approved):

gino-m:
Nice work, thanks for the fix!

Issue #941, opened 2022-08-25
Auth not correctly implemented on some import/export Cloud Functions


gino-m:
- [ ] Cloud Functions should be using the user credentials, currently via the user's session
- [ ] Existing Firebase rules should be applied accordingly to limit access to data.
- [ ] Audit info (created, lastModified, and owner) should be set accordingly.

Originally reported as #720.

gino-m:
Duplicate of #753

PR #1156, opened 2023-09-05
Authenticate calls to HTTP Cloud Functions

gino-m:
Towards #941.

Server-side Cloud Functions:
- [x] Use auth header or session cookie if available
- [x] Add `sessionLogin` endpoint to generate session cookie on deman
- [x] Require auth for HTTP functions

Web client-side:
- [x] Pass auth header to `importCsv` via POST 
- [x] Call `sessionLogin` to generate session cookie before calling `exportCsv`via GET

@os-micmec and/or @DaoyuT PTAL?

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -6173,41 +6173,6 @@
         "node": ">=14"
       }
     },
-    "node_modules/@pnpm/config.env-replace": {
-      "version": "1.1.0",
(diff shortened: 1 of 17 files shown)


web/package-lock.json
os-micmec, line 1: genuine question, but do we need to commit these?

os-micmec (approved):

PR #1161, opened 2023-09-07
1147 - Survey list style

### LINKED ISSUE #1147: Layout and styling of survey list incorrect and incomplete
- [x] Position and font of Surveys heading
- [x] Position and size of logo
- [x] User avatar missing - we should show existing user widget in top right corner
- [x] Background color doesn't cover whole page, only up to end of survey list results
- [x] Size, color, spacing, and alignment of cards incorrect
- [x] Cards cut off on right size on different screen sizes; cards should flow based on available space

App URI:
/surveys

Actual:
image

Expected:
image

@rfontanarosa PTAL?

rfontanarosa:
close #1147

https://github.com/google/ground-platform/assets/10047945/e11a3214-ab1d-4d0a-8c91-6e2f23cdcd02

### DIFF
--- a/web/src/app/components/survey-list/_survey-list.component-theme.scss
+++ b/web/src/app/components/survey-list/_survey-list.component-theme.scss
@@ -5,12 +5,9 @@
   // TODO: Use theme colors once GM3 is supported (or custom palette is defined)
   // The colors defined below are a part of the GM3 color schemes, while the
   // theme colors are based on M2 palettes.
-  .surveys-page {
-    background-color: #edeee9; /* M3/sys/light/surface-container */
-  }
 
-  .surveys-page-title {
-    color: #1a1c19; /* M3/sys/light/on-surface */
+  .page {
+    background-color: #edeee9;
   }
 
   .survey-card {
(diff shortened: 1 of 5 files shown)


web/src/app/components/survey-list/_survey-list.component-theme.scss
```diff
@@ -5,28 +5,31 @@
...
+  .page {
+    background-color: #edeee9;
   }
 
-  .survey-card {
-    background-color: #f9faf4; /* M3/sys/light/surface */
-  }
+  .mat-mdc-card {
+    &.survey-card {
+      background-color: #f9faf4; /* M3/sys/light/surface */
```
gino-m, line 15: Are there SCSS variables we can refer to here and below instead of hard-coding the colors?

gino-m (changes requested):
Nice work! One nit, @amysorto could you also PTAL?

web/src/app/components/survey-list/_survey-list.component-theme.scss
rfontanarosa: I left it as they already are because this TODO comment:  

  // TODO: Use theme colors once GM3 is supported (or custom palette is defined)
  // The colors defined below are a part of the GM3 color schemes, while the
  // theme colors are based on M2 palettes.

gino-m (approved):

Issue #1096, opened 2023-06-02
Wrong page when starting a new survey and looking at demo one


os-micmec:
**Describe the bug**
We get this page 
<img width="632" alt="Screenshot 2023-06-02 at 11 52 56 AM" src="https://github.com/google/ground-platform/assets/92086423/1d88cc4b-d6d8-457c-8d5f-f01e8cb0439b"> when interacting between Demo survey and Start a new survey workflow.

**To Reproduce**
1. Run `npm run start:local
2. Go to the http://localhost:5000/surveys page
3. Click on "Demo Survey"
<img width="764" alt="Screenshot 2023-06-02 at 11 53 45 AM" src="https://github.com/google/ground-platform/assets/92086423/d434158b-6a4c-4e4f-bf4a-0a0dfe7bb550">
5. Once the Demo Survey loads, navigate back with the browser left arrow to go back to the surveys page.
6. Now click on "Start new survey"
7. You should see the wrong page load up.

**Expected behavior**
On step 6, we should see 
<img width="1311" alt="Screenshot 2023-06-02 at 11 56 25 AM" src="https://github.com/google/ground-platform/assets/92086423/46ae2b07-3c62-40a9-93cb-dee4133f11e4"> instead of
<img width="632" alt="Screenshot 2023-06-02 at 11 52 56 AM" src="https://github.com/google/ground-platform/assets/92086423/1d88cc4b-d6d8-457c-8d5f-f01e8cb0439b">

gino-m:
On dev server we see this when creating survey:

<img width="1468" alt="image" src="https://github.com/google/ground-platform/assets/228050/5c88299b-f218-44d5-802d-34b77087c59d">

PR #1167, opened 2023-09-12
Refactor survey creation flow

### LINKED ISSUE #1119: First setup phase is always showing first
**Describe the bug**
When we are in the process of creating/editing a survey, we have code that lands us on the most relevant (current) phase of the process depending on what data was already submitted. The problem is that, while the UI is deciding what is the most relevant phase, we are showing the initial one to the user.

**To Reproduce**
1. Create e new survey
2. Add a survey name and description and go to the next phase
3. Refresh. You should see the first phase for a split of a second

**Expected behavior**
We should have a loading bar while we are deciding state.

DaoyuT:
Fix #1119 & #1096
ALso fix:
 - rename css class name to be more specific to avoid overwritten by other rule with the same name
 - redesign survey list navigation flow (see pic in PR)
 - only replace url when navigating from 1st step to 2nd step in survey creation flow

survey list navigation flow:
Survey list navigation flow drawio

### DIFF
--- a/web/src/app/components/survey-list/survey-list.component.ts
+++ b/web/src/app/components/survey-list/survey-list.component.ts
@@ -51,8 +51,12 @@ export class SurveyListComponent implements OnInit, OnDestroy {
     );
   }
 
-  onSurveyClicked(index: number) {
-    this.navigationService.selectSurvey(this.surveys.get(index)!.id);
+  onSurveyClicked(clickedSurvey: Survey) {
+    if (this.isSetupFinished(clickedSurvey)) {
+      this.navigationService.selectSurvey(clickedSurvey.id);
+    } else {
+      this.navigationService.navigateToCreateSurvey(clickedSurvey.id);
+    }
   }
 
   onNewSurvey() {
@@ -63,5 +67,22 @@ export class SurveyListComponent implements OnInit, OnDestroy {
    */
   ngOnDestroy(): void {
     this.subscription.unsubscribe();
+  }
+
+  private isSetupFinished(survey: Survey): boolean {
+    // To make it simple we are not checking the LOIs here since defining tasks is the step after defining LOIs.
+    return this.hasTitle(survey) && this.hasJob(survey) && this.hasTask(survey);
+  }
+
+  private hasTitle(survey: Survey): boolean {
+    return survey.title.trim().length > 0;
+  }
+
+  private hasJob(survey: Survey): boolean {
+    return survey.jobs.size > 0;
+  }
+
+  private hasTask(survey: Survey): boolean {
+    return survey.jobs.values().next().value.tasks.size > 0;
   }
 }
--- a/web/src/app/pages/create-survey/create-survey.component.spec.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.spec.ts
@@ -192,7 +192,7 @@ describe('CreateSurveyComponent', () => {
   it('shows spinner when survey not loaded', () => {
     const spinner = fixture.debugElement.query(By.css('#loading-spinner'))
       .nativeElement as HTMLElement;
-    // TODO(daoyu): replace it with a spinner component
+    // TODO(#1170): replace it with a spinner component
     expect(spinner.innerHTML).toEqual('Nothing');
   });
 
--- a/web/src/app/pages/create-survey/create-survey.component.spec.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.spec.ts
@@ -53,7 +53,7 @@ describe('CreateSurveyComponent', () => {
   let navigationServiceSpy: jasmine.SpyObj<NavigationService>;
   let route: ActivatedRouteStub;
   let activeSurvey$: Subject<Survey>;
-  let loisActiveSurvey: Immutable.List<LocationOfInterest>;
+  let lois: Immutable.List<LocationOfInterest>;
   let surveyServiceSpy: jasmine.SpyObj<SurveyService>;
   let jobServiceSpy: jasmine.SpyObj<JobService>;
   let loiServiceSpy: jasmine.SpyObj<LocationOfInterestService>;
@@ -152,10 +152,10 @@ describe('CreateSurveyComponent', () => {
       'LocationOfInterestService',
       ['getLocationsOfInterest$']
     );
-    loisActiveSurvey = Immutable.List();
+    lois = Immutable.List();
     loiServiceSpy.getLocationsOfInterest$.and.returnValue(
       new Observable(observer => {
-        observer.next(loisActiveSurvey);
+        observer.next(lois);
         observer.complete();
       })
     );
--- a/web/src/app/pages/create-survey/create-survey.component.spec.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.spec.ts
@@ -134,7 +134,7 @@ describe('CreateSurveyComponent', () => {
       'getActiveSurvey$',
       'updateTitleAndDescription',
       'createSurvey',
-      'getCurrentSurvey',
+      'getActiveSurvey',
     ]);
     surveyServiceSpy.createSurvey.and.returnValue(
       new Promise(resolve => resolve(newSurveyId))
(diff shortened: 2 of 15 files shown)


web/src/app/services/survey/survey.service.ts
```diff
@@ -61,6 +61,10 @@ export class SurveyService {
     this.activeSurvey$.subscribe(survey => (this.currentSurvey = survey));
   }
 
+  getCurrentSurvey(): Survey {
```
gino-m, line 64: Nit: I think active vs current isn't unclear a first glance. Really this is exposing the non-reactive version of `activeSurvey$`. Maybe just expose it as a readonly property called `activeSurvey`, perhaps a la https://stackoverflow.com/a/74165349/9986466 ? Same throughout.

web/src/app/services/navigation/navigation.service.ts
```diff
@@ -232,11 +232,11 @@ export class NavigationService {
     this.router.navigate([NavigationService.SURVEYS_SEGMENT]);
   }
 
-  navigateToCreateSurvey(surveyId: string | null): void {
+  navigateToCreateSurvey(surveyId: string | null, replaceUrl = false): void {
     const url = `${NavigationService.SURVEYS_SEGMENT}${
       surveyId ? `/${surveyId}` : ''
     }/${NavigationService.SURVEYS_CREATE}`;
-    this.router.navigateByUrl(url, {replaceUrl: true});
+    this.router.navigateByUrl(url, {replaceUrl: replaceUrl});
```
gino-m, line 239: Nit: `{replaceUrl: replaceUrl}` -> `{replaceUrl}`

web/src/app/pages/edit-survey/edit-survey.component.ts
```diff
@@ -20,36 +20,43 @@ import {SurveyService} from 'app/services/survey/survey.service';
...
-    this.surveyService
-      .getActiveSurvey$()
-      .subscribe(survey => (this.survey = survey));
     navigationService.init(route);
   }
 
   ngOnInit(): void {
     this.navigationService.getSurveyId$().subscribe(surveyId => {
       if (surveyId) {
+        this.currentSurveyId = surveyId;
```
gino-m, line 45: `activeSurveyId`? (see below)

web/src/app/pages/edit-survey/edit-survey.component.ts
```diff
@@ -20,36 +20,43 @@ import {SurveyService} from 'app/services/survey/survey.service';
...
         this.surveyService.activateSurvey(surveyId);
       }
     });
+    this.surveyService
```
gino-m, line 49: `const survey = await firstValueFrom(thisSurvey..pipe..filter(..))`;

web/src/app/pages/edit-survey/edit-survey.component.html
```diff
@@ -17,7 +17,9 @@
 <div class="page">
   <ground-header></ground-header>
 
-  <div class="container">
+  <!-- TODO(daoyu): replace it with a spinner component -->
```
gino-m, line 20: Instead of doing this in every page, can we create a component that wraps pages which require an active survey? The spinner and branching logic could be handled there. It could replace the top `div` and `ground-header` as well, eg

```
<ground-active-survey>
  <mat-nav-list class="left-menu">
  ...
```

If you don't intend to implement the spinner in this PR, please add a bug ref to the TODOs.

gino-m (changes requested):
Thanks, @DaoyuT!  A few [hopefully helpful] suggestions.

web/src/app/pages/create-survey/create-survey.component.spec.ts
```diff
@@ -53,7 +53,7 @@ describe('CreateSurveyComponent', () => {
   let navigationServiceSpy: jasmine.SpyObj<NavigationService>;
   let route: ActivatedRouteStub;
   let activeSurvey$: Subject<Survey>;
-  let loisActiveSurvey$: Immutable.List<LocationOfInterest>;
+  let loisActiveSurvey: Immutable.List<LocationOfInterest>;
```
gino-m, line 56: Do we need to specified `ActiveSurvey` here? IIrc all data in our model is within the context of the current survey, so shouldn't be necessary.

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -58,25 +58,34 @@ export class CreateSurveyComponent implements OnInit {
...
+      this.currentSurveyId = surveyId
+        ? surveyId
+        : NavigationService.SURVEY_ID_NEW;
+      this.surveyService.activateSurvey(this.currentSurveyId);
     });
+
     this.surveyService
       .getActiveSurvey$()
-      .pipe(first())
+      .pipe(
```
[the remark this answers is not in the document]
DaoyuT: Done, thanks.

web/src/app/services/navigation/navigation.service.ts
DaoyuT: Done, thanks

web/src/app/pages/edit-survey/edit-survey.component.ts
DaoyuT: Done, thanks

web/src/app/pages/edit-survey/edit-survey.component.html
DaoyuT: AFAIK, Angular does not allow you to put Dom inside a component. You'd have to use `ng-template` like
```
<ground-active-survey [template]="temp" >
</ground-active-survey>

<ng-template #temp>
  <mat-nav-list class="left-menu">
  ...
</ng-template>
```
My past experience with `ng-template` is terrible(lots of limitations). And there aren't going to be many pages, I feel repeating it is much easier than creating a component and dealing with `ng-template`, so I chose this way.

We could extract this out to a component later if we start to have too many repetitions and feel confident with `ng-template`.

web/src/app/pages/create-survey/create-survey.component.spec.ts
DaoyuT: Make sense, I originally wanted to remove the dollar sign, but I think `lois` is good enough and it's more concise.

web/src/app/components/survey-list/survey-list.component.ts
```diff
@@ -64,4 +68,21 @@ export class SurveyListComponent implements OnInit, OnDestroy {
   ngOnDestroy(): void {
     this.subscription.unsubscribe();
   }
+
+  private isSetupFinished(survey: Survey): boolean {
```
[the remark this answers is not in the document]
DaoyuT: I would like it to be in a more common place so that I don't have to repeat it twice. But then the router or navigation service would have to get the survey to know where to route/navigate... Let's discuss

web/src/app/pages/edit-survey/edit-survey.component.html
DaoyuT: Content projection is ALL YOU NEED

web/src/app/components/survey-list/survey-list.component.ts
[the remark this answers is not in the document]
DaoyuT: Add a state to survey, draft, complete, (testing).

if draft, go to creation flow 1st step with draft survey (from list/edit page)
if complete, go to edit page from create page, go to map page from list page

web/src/app/pages/edit-survey/edit-survey.component.ts
DaoyuT: Done.

web/src/app/services/survey/survey.service.ts
DaoyuT: Done.

web/src/app/pages/edit-survey/edit-survey.component.html
DaoyuT: Created bug #1170.

web/src/app/components/survey-list/survey-list.component.ts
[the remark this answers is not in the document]
DaoyuT: Created bug #1171

gino-m (approved):
LGTM! To save another round of comments I added the bug #1170 to TODOs myself, hope it's ok. Feel free to merge when ready!

DaoyuT:
> LGTM! To save another round of comments I added the bug #1170 to TODOs myself, hope it's ok. Feel free to merge when ready!

Merged via my phone on a train, feels good 😎

PR #1176, opened 2023-09-19
Show spinner while survey is activating.

### LINKED ISSUE #1162: Show spinner while survey is activating

DaoyuT:
Close #1162 

Screenshot:
Screenshot 2023-09-18 at 11 48 08 PM

### DIFF
--- a/web/src/app/pages/create-survey/_create-survey.component-theme.scss
+++ b/web/src/app/pages/create-survey/_create-survey.component-theme.scss
@@ -19,6 +19,10 @@
 
   // TODO(amysorto): use typography-config once typography levels defined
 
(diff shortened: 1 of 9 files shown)


gino-m:
Thanks! My instinct tells me we shouldn't have "Loading survey..." text below the spinner. @vittorino for UX guidance.

vittorino:
It's ok to have the message there. I thought maybe we could also show the survey skeleton during loading (see image).
image

gino-m:
Good idea! Let's save for future with, since this spinner is used on multiple screens so multiple skeletons would be needed.

gino-m (approved):

PR #1182, opened 2023-09-21
Hack up survey JSON editor for debugging

gino-m:
Hacking up an editor to help us debug, tweak, backup, and restore our survey definitions.

image

@DaoyuT PTAL?

### DIFF
--- a/web/src/app/services/data-store/data-store.service.ts
+++ b/web/src/app/services/data-store/data-store.service.ts
@@ -35,6 +35,7 @@ import {Task} from 'app/models/task/task.model';
 
 const SURVEYS_COLLECTION_NAME = 'surveys';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type JsonBlob = {[field: string]: any};
 
 // TODO: Make DataStoreService and interface and turn this into concrete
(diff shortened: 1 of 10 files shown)


web/src/app/services/data-store/data-store.service.ts
```diff
@@ -66,6 +66,27 @@ export class DataStoreService {
...
+      await firstValueFrom(
+        this.db.collection(SURVEYS_COLLECTION_NAME).doc(id).get()
+      )
+    ).data();
+  }
+
+  /**
+   * Updates the raw survey object in the db. Used for debbuging only.
+   */
+  async saveRawSurvey(id: string, data: any) {
```
DaoyuT, line 83: nit: `data: any` <--- Could this be string?

DaoyuT (changes requested):
We should have test for this new component ideally, but given this is a non-prod tool, I'll let it slide. Left one nit, LGTM other than that.

web/src/app/services/data-store/data-store.service.ts
gino-m: How about this?

DaoyuT (approved):
LGTM, there seem to be a lint issue which is failing the presubmit

DaoyuT (approved):

Issue #1188, opened 2023-09-27
Clarify product behavior and data model for "suggested LOIs" and "conditional tasks"


gino-m:
Specifically:

* Define how behavior of "systematic sampling" (survey organizers specify LOIs *a priori*, before data is collected) differ from "ad hoc sampling" (LOIs are added while collecting data) and "hybrid sampling" ("both" option in create survey job permissions wizard page) wrt:
  * How are LOIs flagged based on whether they were defined *a priori* or *ad hoc*? How is does is that represented to the user and in the db?
  * How are task which only apply to"ad hoc" data collection defined via the UI and in the db?
* This also begs the question of how "conditional tasks" will be represented in the db and to the user. For V1.0, we need at least two types of conditions:
  * Show task only when previous answer is one of X values
  * Show task only when adding a new LOI
  * How do we define these two conditions in a way which leaves room for other types of conditions in the future? e.g. for values expressions and on repeated groups of tasks

Also:

- [ ] How do we differentiate LOIs from other reported geometries on the map?

PR #1190, opened 2023-09-27
Integrate LOI Permissions component into onboarding flow

nwkotto:
image

Also refactored progress indication to make it dynamic.

### DIFF
--- a/web/src/app/pages/create-survey/create-survey.component.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.ts
@@ -60,7 +60,7 @@ export class CreateSurveyComponent implements OnInit {
     route: ActivatedRoute
   ) {
     navigationService.init(route);
-    this.permissionsOptionSelected(LoiPermissionsOption.SURVEY_ORGANIZERS);
+    this.onLoiPermissionsChange(LoiPermissionsOption.SURVEY_ORGANIZERS);
   }
 
   async ngOnInit(): Promise<void> {
@@ -217,7 +217,7 @@ export class CreateSurveyComponent implements OnInit {
     this.survey = this.surveyService.getActiveSurvey();
   }
 
-  permissionsOptionSelected(permissionsOption: LoiPermissionsOption) {
+  onLoiPermissionsChange(permissionsOption: LoiPermissionsOption) {
     this.loiPermissionsOption = permissionsOption;
     this.skipLoiSelection =
       permissionsOption === LoiPermissionsOption.DATA_COLLECTORS;
@@ -261,6 +261,7 @@ export class CreateSurveyComponent implements OnInit {
     await this.jobService.addOrUpdateJob(this.surveyId!, job.copyWith({name}));
   }
 
+  // TODO: Move LOI permissions saving to job service.
   private async saveLoiPermissions() {
     if (!this.loiPermissionsOption) return;
 
--- a/web/src/app/pages/create-survey/create-survey.component.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.ts
@@ -244,7 +244,7 @@ export class CreateSurveyComponent implements OnInit {
   @ViewChild('loiPermissions')
   loiPermissions?: LoiPermissionsComponent;
 
-  private getExistingJob(): Job {
+  private getFirstJob(): Job {
     // there should only be at most one job attached to this survey at this
     // point when user is still in the survey creation flow.
     return this.survey!.jobs.values().next().value;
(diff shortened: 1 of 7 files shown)


web/src/app/pages/create-survey/loi-permissions/loi-permissions.component.ts
```diff
@@ -35,8 +43,12 @@ interface OptionCardConfig {
   templateUrl: './loi-permissions.component.html',
   styleUrls: ['./loi-permissions.component.scss'],
 })
-export class LoiPermissionsComponent {
-  readonly loiPermissionsControlKey = 'loiPermissionsMethod';
+export class LoiPermissionsComponent implements OnInit, OnChanges {
+  @Input() permissionsOption!: LoiPermissionsOption;
```
gino-m, line 47: Nit: Does this refer to the original select option when the component is initialized? If so maybe we can call it something that makes that obvious, e.g. `originalSelection`.

web/src/app/pages/create-survey/loi-permissions/loi-permissions.component.ts
```diff
@@ -35,8 +43,12 @@ interface OptionCardConfig {
...
+  @Output() permissionsOptionSelected: EventEmitter<LoiPermissionsOption> =
```
gino-m, line 48: Following naming convention in Angular docs, I think this would be called `select`. Note I'm intentionally omitting "loiPermission" since it's clear from context. ,and since its name is likely to change once we clarify behavior of this feature (#1188).

web/src/app/pages/create-survey/loi-permissions/loi-permissions.component.ts
```diff
@@ -66,7 +78,23 @@ export class LoiPermissionsComponent {
...
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes['permissionsOption']) {
+      this.formGroup.controls.loiPermissionsOption.setValue(
+        this.permissionsOption
+      );
+    }
+  }
+
+  loiPermissionsOptionCardClicked(value: LoiPermissionsOption) {
```
gino-m, line 97: Nit: `onSelect()` for consistency

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -200,6 +217,12 @@ export class CreateSurveyComponent implements OnInit {
     this.survey = this.surveyService.getActiveSurvey();
   }
 
+  permissionsOptionSelected(permissionsOption: LoiPermissionsOption) {
```
gino-m, line 220: Nit: `onLoiPermissionsChange` or similar

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -218,18 +241,44 @@ export class CreateSurveyComponent implements OnInit {
   @ViewChild('jobDetails')
   jobDetails?: JobDetailsComponent;
 
+  @ViewChild('loiPermissions')
+  loiPermissions?: LoiPermissionsComponent;
+
+  private getExistingJob(): Job {
```
gino-m, line 247: The current designs allow multiple jobs to be added through the creation flow, so we should have a reference to the current job. @DaoyuT is this supported by the current impl?

web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -218,18 +241,44 @@ export class CreateSurveyComponent implements OnInit {
...
-      // there should only be at most one job attached to this survey at this point when user is still in the survey creation flow
-      job = this.survey!.jobs.values().next().value;
+      job = this.getExistingJob();
     } else {
       job = this.jobService.createNewJob();
     }
     await this.jobService.addOrUpdateJob(this.surveyId!, job.copyWith({name}));
   }
 
+  private async saveLoiPermissions() {
```
gino-m, line 264: Let's move this into the service?

gino-m (changes requested):
Thanks @nwkotto! A few suggestions, and a question for @DaoyuT.

web/src/app/pages/create-survey/create-survey.component.ts
gino-m: We may not have implemented it yet, but on the last page of the creation flow you should be able to create another job:

image

If this isn't implemented yet, could you please add a `TODO(#bugno)`? To make the role of the function a little clearer we could also call it `getFirstJob()`.

gino-m (dismissed):
Responded to your responses asking for a few TODOs. Overall LGTM.

web/src/app/pages/create-survey/create-survey.component.ts
gino-m: Yes

gino-m (approved):

os-micmec (approved):

Issue #1194, opened 2023-10-01
Potential race condition when stepping through creation flow


gino-m:
Instead of trying to wait for remote update, the create survey component should update the state of survey internally.

@os-micmec @DaoyuT

gino-m:
@DaoyuT Can we close this?

gino-m:
@rfontanarosa This may have been the issue you were experiencing.

Issue #1191, opened 2023-09-29
Wrap pages which depend on survey in a container component


gino-m:
@DaoyuT @os-micmec FYI I'm giving this a try now

gino-m:
My original idea creating a container component which includes child content using content projects (`<ng-content>`), and passing the current survey to children using Angular's `@Input` fell through, since `ng-content` projects the child's HTML without implementing any of Angular's data binding and related updates. tl;dr: `ng-content` doesn't work with `@Input`. Closing in favor of #1191.

PR #1196, opened 2023-10-02
[DRAFT] Define strategy and components for reactive and static rendering of active survey 

gino-m:
This PR refactors `main-page-container` into a new shared container component which can act as a guard for survey loading. Since `null`s may now appear in the active survey stream, specialized accessors have been added to `SurveyService`  for fetching the current survey reactively (UI reacts to changes to the survey in the remote db) and statically (latest snapshot is loaded once when a component is initialized). This PR suggests the following design:

1. Each top-level page requiring a `survey` gets wrapped in a `<ground-survey-page-container>` component. The wrapper ensures child components are only rendered when there is a non-null survey, and take care of the loading spinner and errors (TBD) during loading.
2. Pages and components needing to react to remote changes in real-time call `SurveyService#requireActiveSurvey$()`   to get a stream which emits the latest snapshot once when called, then once on each remote change. Those components can use `pipe` in conjunction with Rx operators like `map` and `filter` to obtain streams for specific fields in the survey, which in turn can be rendered using `async` pipe in templates. If pipes are used instead of `subscribe()`, initialization of derived `$` streams can be done in the component's constructor, for example:

    ```ts
    constructor(surveyService: SurveyService) {
      const survey$ = surveyService.requireActiveSurvey$();
      this.title$ = survey$.pipe(map(s => s.title));
      this.description$ = survey$.pipe(map(s => s.description));
    }
    ```
4. Pages and components which are expected to manipulate a static instance of Survey (i.e., create/edit survey components) can call `SurveyService#getActiveSurvey()!` in `ngOnInit()`, and operate on that instance for the remaining lifecycle of the component. For ex:

    ```ts
    ngOnInit() {
      this.survey = this.surveyService.getActiveSurvey()!;
    }
    ```

Creation flow pages will be updated in subsequent PRs.

Note: My original idea of creating a container component which includes child content using content projects (`<ng-content>`), and passing the current survey to children using Angular's `@Input` fell through, since `ng-content` projects the child's HTML without implementing any of Angular's data binding and related updates. tl;dr: `ng-content` doesn't work with `@Input`.

Towards #1191.

@rfontanarosa @nwkotto @amysorto  FYI

This was accidentally merged in #1193 and reverted in #1194. @DaoyuT PTAL


gino-m:
@DaoyuT @os-micmec I just realized that the approach in #3 only works if `<ground-survey-page-container>` is applied *one level up* the component hierarchy; simply dropping it in a component's template guard calls in `ngOnInit()` from unloaded surveys; only the components contained therein. Any ideas/suggestions?

gino-m:
This isn't really a big improvement, abandoning this PR for now.

In the future we should probably move towards the model of "dumb" hermetic UI components and "smart" containers as described here:
https://blog.angular-university.io/angular-2-smart-components-vs-presentation-components-whats-the-difference-when-to-use-each-and-why/

Issue #1212, opened 2023-10-07
Top bugs to fix for live demo


gino-m:
P0-P1

- [x] Edit survey -> Job - header and Tasks/Sitets toggle missing (related to #1071) @rfontanarosa 
- [x] Gears icon to access edit survey missing @rfontanarosa 
- [ ] #1234
- [ ] #1235
- [ ] #1236
- [x] Create survey -> Sharing -> Continue button has no effect @rfontanarosa 
- [x] Create survey -> Survey settings layout padding/spacing incorrect @rfontanarosa 
- [x] Edit survey -> Edit job -> Clicking job in left panel doesn't update tasks and LOIs. Refreshing page updates tasks @rfontanarosa 
- [ ] #1237
- [ ] Create/edit survey -> Edit tasks -> #1213
- [ ] #1238
- [x] Submission list -> Add "x" to close panel @rfontanarosa 
- [x] Default LOI labels are incorrect - remove index number, replace with "Unnamed area", "Unnamed point" @rfontanarosa 
- [ ] #1224 
- [ ] #1225

@rfontanarosa @os-micmec @DaoyuT @amysorto @nwkotto If you have time before Tuesday to fix any of these could you please add your username next to the item about? Lmk if anything isn't clear, happy to provide more details. Thank you! 🙏 

P2

- [x] Edit survey -> Select "Survey settings" page by default
- [ ] #1239
- [x] Update edit survey URL - should be `/survey/` and not `/surveys/`
- [ ] #1240
- [x] Sharing screen styling @amysorto
- [ ] #1241
- [ ] #1242
- [ ] #1222
- [ ] #1223

P3

- [ ] #1243
- [ ] `#o=` -> `/site/<id>/submission/<id>`

gino-m:
Closing in favor of individual bugs

PR #1218, opened 2023-10-09
1212 - Added gear icon (edit survey link)

rfontanarosa:
part of #1212

https://github.com/google/ground-platform/assets/10047945/94e8fd00-a873-4d57-bfc9-ca88d95d7b57

### DIFF
--- a/web/src/app/components/header/header.component.ts
+++ b/web/src/app/components/header/header.component.ts
@@ -23,9 +23,23 @@ import {NavigationService} from 'app/services/navigation/navigation.service';
   styleUrls: ['./header.component.scss'],
 })
 export class HeaderComponent {
+  surveyId = '';
+  visible = false;
+
   constructor(public navigationService: NavigationService) {}
+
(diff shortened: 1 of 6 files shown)


web/src/app/components/header/header.component.ts
```diff
@@ -23,9 +23,23 @@ import {NavigationService} from 'app/services/navigation/navigation.service';
   styleUrls: ['./header.component.scss'],
 })
 export class HeaderComponent {
+  surveyId = '';
+  visible = false;
```
gino-m, line 27: This name, implies it controls the visibility of the whole header component rather than just the "survey settings" button.  But a more general question - why do we need to check the survey id at all? Can we just add an input and let the parent component decide when to show the button?

I'll approve for now since we're in a crunch, but if you agree with the above please add a code health issues and `TODO(#issueno)` to simplify this.

gino-m (approved):
Approval, with one caveat.

PR #1228, opened 2023-10-11
Update default LOI naming

rfontanarosa:
part of #1212

Screenshot 2023-10-11 alle 15 16 34


web/src/app/services/loi/loi.service.ts
```diff
@@ -78,26 +79,17 @@ export class LocationOfInterestService {
...
-    index,
-  }: {
-    loi: LocationOfInterest;
-    index: number;
-  }): string {
-    const geometryType = loi.geometry?.geometryType
-      .toLocaleLowerCase()
-      .replace(/_/g, ' ');
-    return `Unnamed ${geometryType} ${index + 1}`;
+  static getAnonymousDisplayName(loi: LocationOfInterest): string {
```
gino-m, line 82: Let's eventually rename this to "getDefaultDisplayName"

web/src/app/services/loi/loi.service.ts
```diff
@@ -78,26 +79,17 @@ export class LocationOfInterestService {
...
+    const geometryType = loi.geometry?.geometryType;
+    return `Unnamed ${geometryType === GeometryType.POINT ? 'point' : 'area'}`;
```
gino-m, line 84: And let's just define two strings, "Unnamed point" and "Unnamed area", instead of using a string template, to make it easier to translate later.

gino-m (approved):
LGTM! A few suggestions for future work..

PR #1231, opened 2023-10-11
Fix padding on edit survey screen

amysorto:
image


gino-m (approved):
The top padding doesn't look correct but still a big improvement, thanks!

PR #1273, opened 2023-11-03
Rename results to data and Submission to SubmissionData for submission model

### LINKED ISSUE #1261: Rename "results" and "task data" to "submission data"
Currently the model has the submission responses as `results` in the model: https://github.com/google/ground-platform/blob/master/web/src/app/models/submission/submission.model.ts#L29.

The `Submission` class will be renamed to `SubmissionData`, and `results` will be renamed to `data`.

Also needs updating here for the cloud function: https://github.com/google/ground-platform/blob/master/functions/src/export-csv.ts#L101

amysorto:
Fixes #1261

### DIFF
--- a/functions/src/export-csv.ts
+++ b/functions/src/export-csv.ts
@@ -104,7 +104,8 @@ export async function exportCsvHandler(
       row.push(location['_latitude'] || '');
       row.push(location['_longitude'] || '');
       row.push(toWkt(loi.get('geoJson')) || '');
-      const data = submission['data'] || {};
+      // TODO(#1288): Clean up remaining references to old responses field
+      const data = submission['data'] || submission['responses'] || submission['results'] || {};
       tasks.forEach((task, taskId) => row.push(getValue(taskId, task, data)));
       csvStream.write(row);
     });
--- a/functions/src/export-csv.ts
+++ b/functions/src/export-csv.ts
@@ -104,8 +104,8 @@ export async function exportCsvHandler(
       row.push(location['_latitude'] || '');
       row.push(location['_longitude'] || '');
       row.push(toWkt(loi.get('geoJson')) || '');
-      const responses = submission['responses'] || {};
-      tasks.forEach((task, taskId) => row.push(getValue(taskId, task, responses)));
+      const data = submission['data'] || {};
+      tasks.forEach((task, taskId) => row.push(getValue(taskId, task, data)));
       csvStream.write(row);
     });
   });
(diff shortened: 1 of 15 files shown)


web/src/app/models/submission/submission.model.ts
```diff
@@ -21,28 +21,28 @@ import {Job} from 'app/models/job.model';
 
 import {Result} from './result.model';
 
-export class Submission {
+export class SubmissionData {
```
gino-m, line 24: Hu @amysorto thanks for the fix! I think a `Submission` is still a submission, but it has data associated with it.  If we rename it, submissiondata has data, which doesn't sound right. Can we revert this change?

web/src/app/models/submission/submission.model.ts
```diff
@@ -21,28 +21,28 @@ import {Job} from 'app/models/job.model';
...
   constructor(
     readonly id: string,
     readonly loiId: string,
     readonly job: Job | null,
     readonly created: AuditInfo,
     readonly lastModified: AuditInfo,
-    readonly results: Map<string, Result>
+    readonly data: Map<string, Result>
```
gino-m, line 31: You could instead create a typealias `SubmissionData` for `Map<string, Result>`

gino-m (changes requested):
Thanks for the cleanup! One main suggestion, PTAL.

gino-m (dismissed):
LGTM, modulo merge conflicts. Thanks!

gino-m:
Gentle ping!

functions/src/export-csv.ts
```diff
@@ -104,8 +104,8 @@ export async function exportCsvHandler(
       row.push(location['_latitude'] || '');
       row.push(location['_longitude'] || '');
       row.push(toWkt(loi.get('geoJson')) || '');
-      const responses = submission['responses'] || {};
-      tasks.forEach((task, taskId) => row.push(getValue(taskId, task, responses)));
+      const data = submission['data'] || {};
```
gino-m, line 107: Should we also maintain backwards compatibility with 'responses' here?

Please add TODO with reference to #1288 to clean this up.

web/src/app/converters/firebase-data-converter.ts
```diff
@@ -397,18 +400,21 @@ export class FirebaseDataConverter {
     if (data === undefined) {
       throw Error(`Submission ${id} does not have document data.`);
     }
+    // Support submissions that have results or responses fields instead of data
+    // before model change.
+    const submissionData = data.data ?? data.results ?? data.responses;
```
gino-m, line 405: Please add TODO with reference to #1288 to clean this up.

gino-m (changes requested):

gino-m:
@amysorto I made the equivalent change in Android in https://github.com/google/ground-android/pull/2059. PTAL?

gino-m (approved):

PR #1299, opened 2023-11-29
Aadd support for other option in multiple choice lists

### LINKED ISSUE #697: Add support for "Other" option in multiple choice lists
image

rfontanarosa:
close #697


gino-m:
@rfontanarosa Please also paste screenshot when ready so we can make sure we're more or less synced with UX designs.

rfontanarosa:
see https://github.com/google/ground-platform/pull/1323

PR #1333, opened 2023-12-08
Add style fixes for create survey flow

amysorto:
[cast (1).webm](https://github.com/google/ground-platform/assets/8575252/57ad8689-f2b4-4478-bfba-0bcdc7894207)

Note: the video has font changes that will be included in a follow up PR


web/src/app/pages/create-survey/loi-permissions/loi-permissions.component.ts
```diff
@@ -57,18 +58,21 @@ export class LoiPermissionsComponent implements OnInit, OnChanges {
       label: 'Predefined',
       description:
         'Data collectors gather data exclusively about preloaded data collection sites.',
+      imagePath: '../../../../assets/img/loi-preview-predefined.svg',
```
gino-m, line 61: This will break if we move the file - can we use absolute paths instead?

gino-m (approved):

gino-m (approved):

PR #1335, opened 2023-12-08
Update to correct fonts and update logo icon size

amysorto:


web/src/app/components/header/_header.component-theme.scss
```diff
@@ -6,7 +6,7 @@
   $foreground: map.get($color-config, foreground);
 
   .ground-title {
-    color: map.get($foreground, secondary-text);
+    color: #1A1C19;
```
gino-m, line 9: Do we have named color vars we can refer to instead of including the literal?

PR #1348, opened 2023-12-12
Task Editor now is not valid when there are no tasks

### LINKED ISSUE #1347: Task Editor: Not valid if no tasks

rfontanarosa:
close #1347

https://github.com/google/ground-platform/assets/10047945/11032727-c98c-4b9c-99de-066421a07e49


gino-m:
@rfontanarosa I don't think we want to create jobs without tasks, in which case the current behavior is correct.

gino-m (approved):

PR #1364, opened 2023-12-19
Add/remove special geometry task 

### LINKED ISSUE #1213: Show special "add data collection site" task in task editors
There's currently no way to change the geometry task type used for adding new data collection sites (LOIs).

* Shown only if data collectors are allowed to add new LOIs
* Can be dragged to be reordered, changed between "Pin, area, capture location"
* ~Special case of a conditional task with condition "when adding a new data collection site"~

image

Design:

Model and schema:

- [x] Add enum state to `Job`: `dataCollectionStrategy` one of `PREDEFINED`, `AD_HOC`, `BOTH`.  Link to data collection strategy selector.
- ~[ ] For backward compat with dev db: if `allowedLoiTypes` is set, set `dataCollectionStrategy` to `BOTH`, not set, `PREDEFINED`. Add TODO to remove.~ <-- only used in UI
- ~[ ] Add boolean state to `Job`: `dataCollectorsCanAddLois` controlled by toggle in LOI editor (see also #1260).~ <- this can be full derived from `dataCollectionStrategy`
- [x] ~When `dataCollectionStrategy` changes, update `dataCollectorsCanAddLois`: `PREDEFINED` -> `false`, otherwise `true`.~
- [x] When changing toggle: if enabled, update `dataCollectionStrategy` to `BOTH`. If disabled, update `dataCollectionStrategy` to `PREDEFINED`

For add LOI tasks:

- [x] Update task editor for geometry types: Allow "Capture location", "Drop a pin", or "Draw an area" to be changed from a drop down
- [x] Add/remove special geometry task with `addLoiTask: true`, default type `CAPTURE_LOCATION` at index 0  whenever `dataCollectorsCanAddLois` is changed (i.e., add it when `true`, remove it when `false`).
- [x] For tasks with `addLoiTask: true` , disable "Required" toggle (always on).

rfontanarosa:
close #1213

https://github.com/google/ground-platform/assets/10047945/c9872728-1ba1-4e88-8e69-7cb5e1543a5d

https://github.com/google/ground-platform/assets/10047945/19d4997b-0f9e-4821-94a5-dc4be29751eb

### DIFF
--- a/web/src/app/components/tasks-editor/task-form/task-form.component.scss
+++ b/web/src/app/components/tasks-editor/task-form/task-form.component.scss
@@ -19,7 +19,6 @@
   border: 1px solid var(--ground-sys-light-surface-variant, #DDE5D9);
   background: var(--ground-sys-light-surface-container-lowest, #FFF);
   padding: 0px 24px 24px 24px;
-  margin-top: 8px;
 
   &.selected {
     border-left: 8px solid #6DDD81;
--- a/web/src/app/components/tasks-editor/task-form/task-form.component.scss
+++ b/web/src/app/components/tasks-editor/task-form/task-form.component.scss
@@ -27,6 +27,10 @@
     padding-left: 17px;
   }
 
+  &.loi-task {
+    background: #DDE5D9;
+  }
+
   &.cdk-drag-preview {
     font-family: 'Manrope';
   }
--- a/web/src/app/components/tasks-editor/task-form/task-form.component.scss
+++ b/web/src/app/components/tasks-editor/task-form/task-form.component.scss
@@ -27,10 +27,6 @@
     padding-left: 17px;
   }
 
-  &.loi-task {
-    background: #DDE5D9;
-  }
-
   &.cdk-drag-preview {
     font-family: 'Manrope';
   }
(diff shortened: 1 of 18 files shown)


web/src/app/components/tasks-editor/task-form/task-form.component.scss
```diff
@@ -27,6 +27,10 @@
     padding-left: 17px;
   }
 
+  &.loi-task {
+    background: #DDE5D9;
```
gino-m, line 31: Reminder: Add `// TODO(#1365): ...`

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -486,6 +486,8 @@ export class DataStoreService {
    * Add a loiTask as first element, reindex the others.
    */
   addLoiTask(tasks: Map<string, Task>): Map<string, Task> {
+    if (tasks.find(task => !!task.addLoiTask)) return tasks;
```
gino-m, line 489: I think `some()` is more idiomatic in this case. Also, it may help readability a bit to make `null` check explicit:

`if (tasks.some(task => task.addLoiTask === true)) return tasks;`

gino-m (changes requested):

gino-m (dismissed):

gino-m (approved):
That was a big piece of work.. Great work!

PR #1370, opened 2023-12-21
Update Firebase lib deps

gino-m:


gino-m:
@os-micmec I think this should fix it

gino-m:
> @os-micmec I think this should fix it

Once I got JDK installed on my new machine I confirmed this does in fact fix the issue.

DaoyuT (approved):

PR #1371, opened 2023-12-22
Fit map to LOIs as LOIs update

### LINKED ISSUE #1222: Map does not pan/zoom to polygons on selection
Currently only pans for pins, does nothing for polygons.

We should zoom and pan for both pins and polygons.

nwkotto@ PTAL?

nwkotto:
Added function to the LOI service to get the boundary box for a list of LOIS; integrated into the map component to fit the map viewport to the LOIs.

Closes #1222

### DIFF
--- a/web/src/app/services/loi/loi.service.ts
+++ b/web/src/app/services/loi/loi.service.ts
@@ -21,9 +21,7 @@ import {map, switchMap} from 'rxjs/operators';
 
 import {Coordinate} from 'app/models/geometry/coordinate';
 import {GeometryType} from 'app/models/geometry/geometry';
-import {MultiPolygon} from 'app/models/geometry/multi-polygon';
 import {Point} from 'app/models/geometry/point';
-import {Polygon} from 'app/models/geometry/polygon';
 import {
   GenericLocationOfInterest,
   LocationOfInterest,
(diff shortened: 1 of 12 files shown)


web/src/app/services/loi/loi.service.ts
```diff
@@ -134,6 +136,37 @@ export class LocationOfInterestService {
...
+  static getLatLngBoundsFromLois(
+    lois: LocationOfInterest[]
+  ): google.maps.LatLngBounds | null {
+    if (!lois.length) return null;
+
+    const data = new google.maps.Data();
+    const bounds = new google.maps.LatLngBounds();
+
+    for (const loi of lois) {
+      if (loi.geometry instanceof Point) {
```
gino-m, line 148: This logic isn't specific to LOIs, and there's a bit of duplicated logic in each branch here. Perhaps we can refactor it into individual Geometry classes, with the root call in Coordinate? For ex, MultiPolygon.extendBounds would call Polygon.extendBounds, which would call Coordinate.extendBounds.

gino-m (changes requested):

web/src/app/services/loi/loi.service.ts
nwkotto: Yeah, this is a bit cleaner! Doesn't quite work out perfectly, in that polygons never seem to reference points directly for some reason, and I don't feel like it's worth it to refactor at the moment, but still a bit DRYer.

gino-m (approved):
This is awesome!

PR #1382, opened 2023-12-28
Fixed rendering of points on map

### LINKED ISSUE #1375: Point LOIs shown at wrong locations on the map
It appears lat and lng (Y and X) are being reversed. The values are correct in the db. For ex, this point in Brazil:

image

Is displayed as being in the South Atlantic:

image

nwkotto:
Closes #1375


gino-m (approved):
Nice work!

PR #1422, opened 2024-01-11
Detect modified lois from stream by comparing them and then rerender …

### LINKED ISSUE #846: [Map] Updating the feature in db does not reflect on map reactively

DaoyuT:
…them on map.

Fix #846

### DIFF
--- a/web/src/app/pages/main-page-container/main-page/map/map.component.ts
+++ b/web/src/app/pages/main-page-container/main-page/map/map.component.ts
@@ -24,6 +24,9 @@ import {
   ViewChild,
 } from '@angular/core';
 import {GoogleMap} from '@angular/google-maps';
+import {Map as ImmutableMap, List} from 'immutable';
+import {Observable, Subscription, combineLatest} from 'rxjs';
+
 import {Coordinate} from 'app/models/geometry/coordinate';
 import {MultiPolygon} from 'app/models/geometry/multi-polygon';
 import {Point} from 'app/models/geometry/point';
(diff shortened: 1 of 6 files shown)


web/src/app/pages/main-page-container/main-page/map/map.component.ts
```diff
@@ -120,13 +120,35 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
       ]).subscribe(([survey, lois, locationOfInterestId]) => {
-        this.lois = lois;
-
-        this.onSurveyAndLocationsOfInterestUpdate(
-          survey,
-          lois,
-          locationOfInterestId
-        );
+        const loisMap = ImmutableMap(lois.map(loi => [loi.id, loi]));
+        const loiIdsToRemove = this.loisMap
```
gino-m, line 124: Can we do a since deep compare rather than checking ID and geometry separately here? 

If you populate a `Set` using a set of `Immutable.fromJS(loi)`, you can simply use `subtract()` on the two sets (before and after states) to get the added/modified and removed/modified values.

gino-m (changes requested):
LG, one suggestion.

web/src/app/pages/main-page-container/main-page/map/map.component.ts
DaoyuT: That would be better, and that would require us to override `equals` function of `LocationOfInterest`. However, `LocationOfInterest` is still an abstract class with other inherent classes that we planned to remove (PointOfInterest, GeoJsonLocationOfInterest... in [loi.model.ts](https://github.com/google/ground-platform/blob/master/web/src/app/models/loi.model.ts)). So that's why I wanted to do the `Set.subtract()` thing later after we instantiate `LocationOfInterest` class and override `equals` function.

Filed #1444 & #1445 with descriptions and assigned to myself for now. Also updated the TODOs to relate to the bugs. WDYT?

gino-m: SGTM, thanks for filing those!

gino-m (approved):

PR #1425, opened 2024-01-12
Added draft survey validation + Bugfix

### LINKED ISSUE #1395: [Task editor] "Publish" button enabled when no changes made
Merely focusing on a task enabled the button:

https://github.com/google/ground-platform/assets/228050/9a45207e-7e01-4256-9cf1-f3fd0110941c

@rfontanarosa PTAL?

### LINKED ISSUE #1406: Validate changes to survey when clicking "Publish changes"
The editor currently allows publishing invalid surveys. Instead, we should show a message explaining that the survey contains errors and that the user should fix them before proceeding. @amegantz for UXW.

rfontanarosa:
close #1395
close #1406

https://github.com/google/ground-platform/assets/10047945/ea479a2f-8421-4a15-8ccd-35ec738af2a7

### DIFF
--- a/web/src/app/services/draft-survey/draft-survey.service.ts
+++ b/web/src/app/services/draft-survey/draft-survey.service.ts
@@ -15,7 +15,7 @@
  */
 
 import {Injectable} from '@angular/core';
-import {List} from 'immutable';
+import {List, Map} from 'immutable';
 import {BehaviorSubject, Observable, firstValueFrom} from 'rxjs';
 
 import {Job} from 'app/models/job.model';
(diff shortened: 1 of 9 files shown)


web/src/app/services/draft-survey/draft-survey.service.ts
```diff
@@ -80,30 +85,36 @@ export class DraftSurveyService {
...
     const currentSurvey = this.survey$.getValue();
 
-    const currentJob = currentSurvey.jobs.get(jobId);
+    const currentJob = currentSurvey.jobs.get(jobId)!;
 
-    if (currentJob) {
-      const job = currentJob?.copyWith({
-        tasks: this.dataStoreService.convertTasksListToMap(tasks),
-      });
+    const job = currentJob?.copyWith({
```
gino-m, line 93: Do we also need to generate new ids for options, and possibly other nested items? I don't think we depend on options having globally unique ids anywhere, but perhaps we should be consistent since other entities do use unique ids (Surveys, for ex) Perhaps we can file a bug and add a TODO(#issueno) here?

gino-m (changes requested):

web/src/app/services/draft-survey/draft-survey.service.ts
rfontanarosa: I already noticed this. This duplication issue involves options and tasks too. I'll fix it in separate pr.

gino-m (approved):

Issue #1441, opened 2024-01-16
Map zoomed in and not showing other LOIs when LOI id is selected in URL


DaoyuT:
To reproduce:

Go to url with survey id and loi id (e.g. http://localhost:5000/survey/YD35zLKMkhNWtWCrx5gc#f=UQkZm12xtdElQOqAHpqO) and it'll look like this:

![7TCcfvhw9FofJGs](https://github.com/google/ground-platform/assets/10445677/acdca323-28d6-4dac-ac24-269a92a53361)

PR #1442, opened 2024-01-16
Replace loadLocationOfInterest$ with lois$.

DaoyuT:
`lois$` calls `.collection(LOI_COLLECTION_NAME)...` and `loadLocationOfInterest$` calls `.collection(LOI_COLLECTION_NAME).doc(loiId)...`.

For some reason, when `loadLocationOfInterest$` is subscribed, `loi$` emmits a list of LOI of that ID and then a list of full LOIs. That's why it's causing the bug(#1441). We don't have to have `loadLocationOfInterest$` anyway, so here I replaced `loadLocationOfInterest$` with `lois$` plus a `finding LOI by its id from the list` logic.

And now the map fits correctly showing all the LOIs.
Screenshot 2024-01-16 at 3 30 48 PM

### DIFF
--- a/web/src/app/services/data-store/data-store.service.ts
+++ b/web/src/app/services/data-store/data-store.service.ts
@@ -251,37 +251,6 @@ export class DataStoreService {
   }
 
   /**
-   * Returns an Observable that loads and emits the LOI with the specified
-   * uuid.
-   *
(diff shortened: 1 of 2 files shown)


gino-m (approved):
LGTM!

PR #1467, opened 2024-01-22
Fixed direct access to the submission deep linking

### LINKED ISSUE #1455: [Submission details] Submission panel blank when opening deep links to specific submission
https://github.com/google/ground-platform/assets/10047945/321bb410-495e-4604-8afe-346f6e6c3e15

rfontanarosa:
close #1455 

https://github.com/google/ground-platform/assets/10047945/6e82dc13-2e8c-4684-a932-64e1dceacc33

### DIFF
--- a/web/src/app/pages/main-page-container/main-page/secondary-side-panel/secondary-side-panel.component.ts
+++ b/web/src/app/pages/main-page-container/main-page/secondary-side-panel/secondary-side-panel.component.ts
@@ -15,7 +15,7 @@
  */
 
 import {Component} from '@angular/core';
-import {Observable} from 'rxjs';
+import {Observable, Subscription} from 'rxjs';
 
 import {
   NavigationService,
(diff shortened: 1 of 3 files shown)


web/src/app/pages/main-page-container/main-page/secondary-side-panel/secondary-side-panel.component.ts
```diff
@@ -28,18 +28,30 @@ import {
...
+      this.navigationService
+        .getLocationOfInterestId$()
+        .subscribe(id => (this.locationOfInterestId = id || ''))
+    );
+
+    this.subscription.add(
+      this.navigationService
+        .getSubmissionId$()
+        .subscribe(id => (this.submissionId = id || ''))
+    );
```
gino-m, line 49: The two streams feeding this component's state can receive events in any order, leading to race conditions. Instead, we should have a single subscription is called only once when the URL is changed. The easiest way to do this would be to expose `route.paramMap` from `NavigationService`, subscribe to that, and call `navigationService.getSubmissionId()` and `getLocationOfInterestId()` directly in the callback.

gino-m (changes requested):
Cool, thanks! There's a bit of jank when the page reloads, which may be a symptom a race condition in the component's handling of the event. See my comment for details.

gino-m (approved):
👍🏻

PR #1469, opened 2024-01-24
[DRAFT] Upgrade firebase-tools to latest

gino-m:


gino-m:
Transitive deps need to be updated to do this. Closing for now, we can revisit later.

PR #1491, opened 2024-02-16
Add .json format and propagate error to console

### LINKED ISSUE #1489: [Import LOIs] Import fails with no explanation when filename is `.json` instead of `.geojson`
We should allow `.json` an extension, and provide an explanation when the wrong extension was provided. Also, we should log errors instead of swallowing them.

os-micmec:
Fixes #1489


gino-m (approved):
Thanks!

gino-m (approved):

PR #1492, opened 2024-02-16
Do not repeat task id when writing tasks

### LINKED ISSUE #1283: Task ids repeated inside task and in outer map in db
No need to repeat keys in the tasks, in the map should be sufficient.

image

@os-micmec @rfontanarosa FYI

DaoyuT:
Fixes #1283


gino-m (approved):
👏

PR #1493, opened 2024-02-19
Handle issues when reading invalid data in Firestore.

### LINKED ISSUE #931: Handle issues when reading invalid data in Firestore
~"Quota exceeded" was being tripped but was only visible in the Chrome debug console on web. We should display these error to the user to help debugging.~

~See also https://github.com/google/ground-android/issues/1227~

~@DaoyuT We'll need to catch errors in Angular components which use Firebase. Import/export is out of scope of this request. We call Firebase in a few ways:~

~- [ ] Create helper to display error message with simple alert~
~- [ ] Call fro Rx streams onError and on thrown exceptions~

Rescoping as following to make more tractable:

- [x] Converter functions which throw errors should instead return `T | Error` 
- [x] Handle errors in caller of converters as follows:
  - [x] If creating a List<>, just log as error and don't add to list.
  - [x] If called directly on `pipe(map())`, we can add a helper `handleErrors` which returns `OperatorFunction<T, R>` to pass to pipe to log and skip.

DaoyuT:
fix #931

### DIFF
--- a/web/src/app/converters/firebase-data-converter.ts
+++ b/web/src/app/converters/firebase-data-converter.ts
@@ -401,12 +401,20 @@ export class FirebaseDataConverter {
    * }
    * </code></pre>
    */
-  static toSubmission(job: Job, id: string, data: DocumentData): Submission {
+  static toSubmission(
+    job: Job,
+    id: string,
+    data: DocumentData
+  ): Submission | Error {
     if (job.tasks === undefined) {
-      throw Error('Job must contain at least once task');
+      return Error(
+        'Error converting to submission: job must contain at least once task'
+      );
     }
     if (data === undefined) {
-      throw Error(`Submission ${id} does not have document data.`);
+      return Error(
+        `Error converting to submission: submission ${id} does not have document data.`
+      );
     }
     // TODO(#1288): Clean up remaining references to old responses field
     // Support submissions that have results or responses fields instead of data
@@ -418,15 +426,7 @@ export class FirebaseDataConverter {
       job,
       FirebaseDataConverter.toAuditInfo(data.created),
       FirebaseDataConverter.toAuditInfo(data.lastModified),
-      Map<string, Result>(
-        keys(submissionData).map((taskId: string) => [
-          taskId as string,
-          FirebaseDataConverter.toResult(
-            job.tasks!.get(taskId)!,
-            submissionData[taskId]
-          ),
-        ])
-      )
+      FirebaseDataConverter.toResults(job, data)
     );
   }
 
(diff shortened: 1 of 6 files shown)


web/src/app/converters/firebase-data-converter.ts
```diff
@@ -465,10 +465,32 @@ export class FirebaseDataConverter {
...
   }
 
+  private static toResults(job: Job, data: DocumentData): Map<string, Result> {
+    const submissionData = data.data ?? data.results ?? data.responses;
+    return Map<string, Result>(
+      keys(submissionData)
+        .map((taskId: string) => [
+          taskId as string,
+          FirebaseDataConverter.toResult(
+            job.tasks!.get(taskId)!,
```
gino-m, line 475: We should also probably handle this case explicitly, and/or comment describing why this can never happen. I'd also suggest adding a brief comment to TSDoc that this method handles errors, i.e., caller doesn't need to catch.

Alternatively, we can have this method return `Map<String, Result | Error>` and let the caller filter and log.

web/src/app/converters/firebase-data-converter.ts
```diff
@@ -465,10 +465,32 @@ export class FirebaseDataConverter {
...
+            submissionData[taskId]
+          ),
+        ])
+        .filter(([_, resultOrError]) => {
```
gino-m, line 479: Can we refactor `filterConversionErrorAndLog` so that we can call it here?

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -455,4 +451,12 @@ export class DataStoreService {
     });
     return tasksMap;
   }
+
+  private filterConversionErrorAndLogError<T>(entityOrError: T | Error) {
```
gino-m, line 455: Nit: `filterAndLogError` (the additional "Error" is clear from context, plus there's nothing specific to "conversion" here)

web/src/app/services/submission/submission.service.ts
```diff
@@ -86,7 +86,13 @@ export class SubmissionService {
             )
           )
         )
-        .subscribe(o => this.selectedSubmission$.next(o))
+        .subscribe(o => {
+          if (o instanceof Error) {
```
gino-m, line 90: You could also use ".filter(filterAndLogError)" above to avoid needing this in subscribe.

gino-m (changes requested):
`Error--`! A few small readability suggestions, LG!

web/src/app/services/submission/submission.service.ts
DaoyuT: Done.

web/src/app/services/data-store/data-store.service.ts
DaoyuT: Done.

web/src/app/converters/firebase-data-converter.ts
DaoyuT: Added TSDoc mentioning that this method handles errors. What do you mean by `this can never happen`?

gino-m: Thanks! Re "this", I was referring to job.tasks being `null` or `undefined`. It can never happen if we checked it previously. If that's not the case, we should also handle this case.

gino-m (changes requested):
Thanks for the edits!

web/src/app/converters/firebase-data-converter.ts
DaoyuT: Got it, we did check `job.tasks` before this. I changed the TSDoc regarding this and changed the null/undefined check a bit.

gino-m (approved):
LGTM! 👍

PR #1499, opened 2024-02-21
New hardcoded add loi task id: $addLoi

### LINKED ISSUE #1496: Replace "addLoiTask" flag with hard coded ID
Task ID `_addLoi`

@rfontanarosa FYI

rfontanarosa:
close #1496


web/src/app/services/task/task.service.ts
```diff
@@ -39,6 +39,8 @@ export type TaskUpdate = {
 export class TaskService {
   private tasks$: Observable<List<Task>>;
 
+  private addLoiTaskId = '_addLoi';
```
gino-m, line 42: Nits:
* Let's define at top level as a `const`
* Let's use `$addLoi` instead of `_addLOi`, which is more ergonomic imo.

gino-m (changes requested):
Two nits, otherwise LG!

gino-m (approved):

PR #1565, opened 2024-03-06
Fix style for loi panel header

### LINKED ISSUE #1540: [Submission list panel] Minor visual glitches in header
- [ ] LOI icon in header doesn't use job color
- [ ] Wrapped text should be aligned to first row

image

@amysorto PTAL?

amysorto:
Fixes #1540

image


gino-m (approved):
Great! Note some checks are failing.

PR #1567, opened 2024-03-06
Migrate to MDC based dialog component

amysorto:
There are a lot of files are touched here, but the MDC dialog does not differ much from the legacy. Some of this is unused code, but that should likely be a separate issue for code clean up.


gino-m (changes requested):
Awesome! Can you paste a screen recording in the PR desc? Also note some checks are failing

amysorto:
> Awesome! Can you paste a screen recording in the PR desc? Also note some checks are failing

Would it make more sense to put all these PRs related to migrating to MDC into a branch, and then having a video walkthough of the entire app? Or do videos for each component?

gino-m:
> > Awesome! Can you paste a screen recording in the PR desc? Also note some checks are failing
> 
> Would it make more sense to put all these PRs related to migrating to MDC into a branch, and then having a video walkthough of the entire app? Or do videos for each component?

I think for velocity's sake we can skip the video now, since this is a cross-cutting change. Thanks!

gino-m (dismissed):

rfontanarosa (approved):

Issue #1576, opened 2024-03-07
[Import CSV] Importing CSV reports success but nothing actually imported


gino-m:
Nothing appears in db after import.

Attaching test CSV with fake data.

[Long Island Farms.csv](https://github.com/google/ground-platform/files/14527120/Long.Island.Farms.csv)

jcqli:
@os-micmec Is this closed now that #1590 is merged?

PR #1590, opened 2024-03-09
Fix import

os-micmec:
Towards #1576


functions/src/import-csv.ts
```diff
@@ -146,10 +146,11 @@ function csvRowToLocationOfInterest(row: any, jobId: string) {
       properties[columnName] = value;
     }
   }
-  const {latStr, lngStr, ...loi} = data;
+  const {lat: latStr, lng: lngStr, ...loi} = data;
   const lat = Number.parseFloat(latStr);
   const lng = Number.parseFloat(lngStr);
   if (isNaN(lat) || isNaN(lng)) {
+    console.error('unable to parse latitude and/or longitude values');
```
gino-m, line 153: This error will not be visible to the user, I don't the we need to log it in the backend

gino-m (changes requested):

gino-m (approved):
LGTM!

rfontanarosa (approved):

Issue #1329, opened 2023-12-06
[Submission details] Display elevation and accuracy for capture location


amysorto:
Right now capture location tasks are treated the same as a point. Although they have some additional data (elevation and accuracy), but it is not shown anywhere.

gino-m:
@vittorino How would this look in the submission panel?

vittorino:
here's an example of how it should look:
<img width="431" alt="image" src="https://github.com/google/ground-platform/assets/679199/c781f4f8-2536-448f-8a3c-e4f5ee6a4a81">

gino-m:
The point would also show on the map like with "Drop a pin" tasks. Should this marker also be included here, maybe with a different icon to indication GPS location instead of manual?

<img width="256" alt="image" src="https://github.com/google/ground-platform/assets/228050/49f5d2dc-508a-48e6-8d82-653a59eeb115">

gino-m:
This doesn't appear to be in the db, or the format aren't consistent. Punting to next iteration.

shobhitagarwal1612:
Yes, we do store elevation in the "Capture location" tasks.

PR #1603, opened 2024-03-12
Add additional information for capture location tasks in submission view

amysorto:
Part of #1329 
Note: Seems like the additional information for capture location tasks is no longer stored in firebase. Maybe that was a change on the Android side? So I removed the code that hid that information previously.

I'm not too sure about formatting for the coordinates, this follows how the geopoint is shown in firebase (not how it is in the mock). Open to suggestions on the right way of doing this!

Also, is there an update if the marker should be included or not?

image


gino-m (dismissed):
TODO: DROP_PIN and DRAW_AREA

gino-m (approved):

PR #1627, opened 2024-03-12
Fix error where enter key triggers delete in 'select one' or 'select multiple'

### LINKED ISSUE #1625: [Create/edit survey] Enter key in "select one" or "select multiple" option field triggers delete
**Describe the bug**
When adding a job question that is a "select one" or "select multiple" type, pressing "enter" key after typing an option will prompt users to delete that option, instead of adding a new option

**Expected behavior**
Should add a new option and focus cursor to enter text for that option

This might be due to the way that the browser is "focusing" on the next component

amysorto:
Fixes #1625

Buttons by default within `<form>` elements get automatically treated as the submit button

### DIFF
--- a/web/src/app/components/tasks-editor/task-form/edit-option/edit-option.component.html
+++ b/web/src/app/components/tasks-editor/task-form/edit-option/edit-option.component.html
@@ -41,10 +41,10 @@
   <div class="delete-button-container">
     <button
       *ngIf="index"
-      tabindex="-1"
       mat-icon-button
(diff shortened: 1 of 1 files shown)


gino-m (approved):
@amysorto Could you file a FR to add a new option on "Enter"?

amysorto:
> @amysorto Could you file a FR to add a new option on "Enter"?

https://github.com/google/ground-platform/issues/1629

PR #1631, opened 2024-03-12
Set width for mat-selects to be the longest option instead of the trigger width

### LINKED ISSUE #1623: [Create/edit survey] Add LOI task type select wrong width and style (orange checkl?)
image

@amysorto As discussed

amysorto:
Fixes #1623

[cast (3).webm](https://github.com/google/ground-platform/assets/8575252/cc62254b-700a-45e3-95dc-9b7e3ab63d66)

### DIFF
--- a/web/src/ground-theme.scss
+++ b/web/src/ground-theme.scss
@@ -163,11 +163,14 @@ snack-bar-container.notification {
   width: 1024px;
 }
 
-// TODO: Can remove once upgraded past v16.1, there is a panelWidth
+// TODO(#1633): Can remove once upgraded past v16.1, there is a panelWidth
 // input property for mat-select that can be set to null. By default
 // options are set to the select trigger width, but can be set to the
 // longest option width instead.
-// For context: https://github.com/angular/components/issues/26000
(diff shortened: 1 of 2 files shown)


web/src/ground-theme.scss
```diff
@@ -162,3 +162,12 @@ snack-bar-container.notification {
   margin: 24px;
   width: 1024px;
 }
+
+// TODO: Can remove once upgraded past v16.1, there is a panelWidth
+// input property for mat-select that can be set to null. By default
+// options are set to the select trigger width, but can be set to the
+// longest option width instead.
+// For context: https://github.com/angular/components/issues/26000
```
gino-m, line 170: Nice, thanks! Can we file a bug for this and reference here?

gino-m (changes requested):
Nit: Can we add padding to the right of the text so the SELECT down arrow isn't right up against the text?

amysorto:
> Nit: Can we add padding to the right of the text so the SELECT down arrow isn't right up against the text?

Updated screenshot:
Screenshot 2024-03-12 at 5 38 39 PM

gino-m (approved):
🔍 Mystery (almost) solved?

PR #1642, opened 2024-03-13
Make submission item text clickable

### LINKED ISSUE #1542: [Submission list] Clicking on item text in submission list has no effect
Only clicks on the ">" work. Instead, the whole item should be clickable.

@amysorto PTAL?

amysorto:
Fixes #1542


gino-m (approved):
👍 !

PR #1661, opened 2024-03-15
Improved formatting for LOI list in onboarding flow

### LINKED ISSUE #1541: [Import LOI list] Minor layout issues
- [x] LOIs in list should be left aligned
- [x] LOIs should be separated by horizontal dividers as in designs
- [x] LOI icons should be shown using the job color

@nwkotto Do you have cycles for this one? Or @amysorto perhaps we can refactor the LOI list from the main page and use it here?

nwkotto:
- Added LOI icons with the same color as the job
- Created new GrounIconModule in the process, to simplify the use of custom icons
- Added horizontal divider between loi list elements

image

Closes #1541


gino-m (approved):
LGTM, thanks for the fix!

PR #1690, opened 2024-03-28
Remove last of legacy material component styles

amysorto:
Last PR of the MDC migration, one step closer to getting to M3! 🎉


gino-m (approved):
This is great news! 🥳  Great work, @amysorto !!!

PR #1705, opened 2024-04-08
Always show user profile widget when logged in (incl. on permission denied error page)

### LINKED ISSUE #1704: [Bug] User icon is not always visible when the user is logged in
If the user is logged in but not in the passlist, it's impossible to logout because the user icon is not visible.

Screenshot 2024-04-08 alle 19 48 52

rfontanarosa:
closes #1704


gino-m (changes requested):
How does this affect when the user profile widget is shown? Looks to me like that part was unchanged?

rfontanarosa:
because the condition is inside the user widget (<div *ngIf="auth.getUser$() | async as user">).

the main problem was that the widget was under canManage condition.

gino-m (approved):
I see, the diff wasn't too clear, thanks for clarifying.

PR #1708, opened 2024-04-08
Accept only .csv & .geojson files when importing LOIs

### LINKED ISSUE #741: [Import] Make it clear which file extenstions are supported for import
Currently, there's no indication within the app itself as to what file formats are supported for importing features. The file selection dialog currently permits selecting files of any type, resulting in immediate failure on selection of an unsupported format.

https://user-images.githubusercontent.com/11237600/120385518-88090280-c2f5-11eb-8b6e-06618ddd107e.mov

DaoyuT:
Also changed the label in the dialog to indicate accepted file extensions.

Fixes #741 

Screenshot 2024-04-08 at 2 00 17 PM


gino-m (approved):
Awesome, thank you!

Issue #1711, opened 2024-04-09
Cannot import LOIs in local env


DaoyuT:
Run `npm run start:local` command to start up local env, but I cannot import LOIs.

DaoyuT:
Fixed by #1712

PR #1712, opened 2024-04-09
Fix local env so that developers can import LOIs.

DaoyuT:


functions/src/common/auth.ts
```diff
@@ -59,16 +60,26 @@ export async function setSessionCookie(req: https.Request, res: Response): Promi
   res.cookie(SESSION_COOKIE_NAME, cookie, { maxAge: expiresIn, httpOnly: true, secure: true });
 }
 
-function getRole(user:DecodedIdToken, survey: DocumentSnapshot): string | null {
-  const acl = survey.get('acl')
+const OWNER_ROLE = 'OWNER';
```
gino-m, line 63: Consider moving consts to top of file.

gino-m (changes requested):
One nit, LG!

functions/src/common/auth.ts
```diff
@@ -14,13 +14,16 @@
  * limitations under the License.
  */
 
-import {DecodedIdToken, getAuth} from 'firebase-admin/auth';
-import {DocumentSnapshot} from 'firebase-admin/firestore';
-import {https, Response} from 'firebase-functions/v1';
+import { DecodedIdToken, getAuth } from 'firebase-admin/auth';
+import { DocumentSnapshot } from 'firebase-admin/firestore';
+import { https, Response } from 'firebase-functions/v1';
+import { EmulatorIdToken } from '../handlers';
```
gino-m, line 20: You may also need to remove the space between brackets.

functions/src/common/auth.ts
DaoyuT: SG, done.

gino-m (dismissed):

functions/src/common/auth.ts
DaoyuT: Done.

gino-m (approved):

DaoyuT:
Fix #1711

PR #1717, opened 2024-04-10
 drawing-tools.component.spec DOM text reinterpreted as HTML

Shivam7-1:


gino-m (approved):
Hi @Shivam7-1, thanks for the PR. The code in question is a unit test, so there shouldn't be a way to inject malicious HTML. That said, we're really only checking for the text in the specified element, so if tests pass this should be ok to merge.

PR #1729, opened 2024-04-12
 create-survey.component.spec.ts DOM text reinterpreted as HTML

Shivam7-1:


gino-m (approved):

gino-m (approved):

Shivam7-1:
Hi @gino-m Thanks For Reviewing 
  Above PR (#1729 ) was approved recently, and I wanted to check in to see if there's anything else needed from my side to proceed with the merge. 


Thanks

PR #1737, opened 2024-04-13
Updated drawing-tools File DOM text reinterpreted as HTML

Shivam7-1:


Shivam7-1:
Hi @gino-m Could You Please Review Above PR
Thanks

gino-m:
Tests fail with:
```
Error: Error: src/app/pages/main-page-container/main-page/drawing-tools/drawing-tools.component.spec.ts:328:23 - error TS2339: Property 'innerText' does not exist on type 'Element'.
```

Shivam7-1:
Hi @gino-m Thanks For Reviewing 
I had check with above error it won't be able recognise innertext because of HTMLElement is not Declared 
Hence I had Added Those Change No Error will Occurs over here

Thanks

Shivam7-1:
Hi @gino-m 
Is there any Changes are Required?

gino-m (approved):
LGTM!

Shivam7-1:
Hi @gino-m Thanks For Reviewing
Above PR was approved recently, and I wanted to check in to see if there's anything else needed from my side to proceed with the merge.

Thanks

NeilFraser:
@gino-m This test code is odd.  It apparently takes an HTML `IMG` tag and checks that `.innerHTML` (or now `.innerText`) contains a job name.  But HTML image tags don't contain either text or HTML.  Does this code even run?

gino-m:
@NeilFraser Indeed it does. `job1Item` and `job12tem` are the `DIV`s surrounding the `IMG` elements, not the `IMG` elements themselves. Thanks for keeping an eye out, though. Grateful for your input elsewhere on the codebase if/when you have the cycles.

Issue #1478, opened 2024-02-07
[Create/edit survey] Allow organizers to add conditional tasks


gino-m:
Trigger:
![Image](https://github.com/google/ground-platform/assets/228050/b1dfea9e-5718-45d4-99ed-897ff96cb15d)

List selected values as comma-separated list in drop-down text:
![Image](https://github.com/google/ground-platform/assets/228050/223d32d0-e390-414e-918a-8b2597c418b7)

Truncate long names:
![Image](https://github.com/google/ground-platform/assets/228050/0752a700-b7b5-4794-89f9-d9664d99f803)

@rfontanarosa @vittorino FYI

jcqli:
@gino-m is this a duplicate of [#1955](https://github.com/google/ground-android/issues/1955)?

gino-m:
Yes! I was trying to call out top priority tasks, we can close that other one  as duplicate, thanks!

jcqli:
Since they're filed in separate repos let's use that that bug to track the android work to support this

jcqli:
@rfontanarosa From our sync earlier, assigning to you to tackle in next iteration

sufyanAbbasi:
I'm working on the problem from the Android side, and we should work towards agreeing on a JSON schema to encode "conditions" into the Task system. There are some interesting implementation questions that we should discuss:

1. Do we bind condition-related parameters to the Tasks that fulfill the condition or not? Should each Task not know about the conditions which are bound to it?
2. Should conditions have their own IDs which can be tracked across various Tasks?
3. Can a Task have multiple conditions to trigger it? If so, do we want to be able to configure "any" or "all" conditions for them? (Task A with option 1 or Task B with option 2 etc.)
4. Can conditional Tasks be reordered conditionally?
5. Can Tasks have conditional options?
6. Are conditions based on the `label` or `index` field of the options?
7. Can conditions be bound to the "other" option?
8. Can conditions be based on location or a bounding box?
9. Regular expression matching? May be useful for constructing easy string matching conditions

Here is a typical Tasks JSON:
```
"tasks": {
        "ePpIYkygrcm6cANlrhs3": {
          "required": false,
          "type": "multiple_choice",
          "label": "Pick an option, any option!",
          "index": 1,
          "cardinality": "select_one",
          "addLoiTask": false,
          "id": "ePpIYkygrcm6cANlrhs3",
          "hasOtherOption": true,
          "options": {
            "vknPVEALVkUvlOoyNiV9": {
[...]

PR #1738, opened 2024-04-16
Conditional Task model and Firebase converter

rfontanarosa:
Towards #1478


web/src/app/converters/firebase-data-converter.ts
```diff
@@ -248,31 +247,19 @@ export class FirebaseDataConverter {
...
-      !Array.isArray(values) ||
-      values.some(v => !(v instanceof String))
-    )
-      return undefined;
-    else
-      return new MultipleChoiceResponseCondition(taskId, List<String>(values));
+    if (!map) return undefined;
+    return new TaskCondition(
+      map?.matchType,
+      map?.expressions
```
gino-m, line 253: Can this be simplified as `map?.expressions?.map( ... ) || List()`?

gino-m (changes requested):
One nit, LGTM

gino-m (approved):
✅ !

PR #1747, opened 2024-04-18
Task Condition: Add/Remove Toggler, Form/Task converters

rfontanarosa:
Towards #1478

### DIFF
--- a/web/src/app/components/tasks-editor/task-form/task-form.component.html
+++ b/web/src/app/components/tasks-editor/task-form/task-form.component.html
@@ -59,6 +59,10 @@
           <span>Duplicate</span>
         </button>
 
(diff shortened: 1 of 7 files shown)


web/src/app/components/tasks-editor/task-form/task-form.component.html
```diff
@@ -59,6 +59,10 @@
           <span>Duplicate</span>
         </button>
 
+        <button mat-menu-item (click)="onTaskConditionToggle()">
+          <span>{{ !conditionControl ? 'Add condition' : 'Delete condition' }}</span>
```
gino-m: Maybe "Remove condition" would be more correct here.

gino-m (approved):

PR #1749, opened 2024-04-19
Show only lois in the selected job in loi editor.

### LINKED ISSUE #1713: [LOI Editor] Map shows LOIs for all jobs
@nwkotto @rfontanarosa @DaoyuT  FYI

Map should only show predefined LOIs in the select job

DaoyuT:
Fix #1713

### DIFF
--- a/web/src/app/pages/main-page-container/main-page/map/map.component.ts
+++ b/web/src/app/pages/main-page-container/main-page/map/map.component.ts
@@ -26,7 +26,7 @@ import {
 } from '@angular/core';
 import {GoogleMap} from '@angular/google-maps';
 import {Map as ImmutableMap, List} from 'immutable';
-import {Observable, Subscription, combineLatest} from 'rxjs';
+import {BehaviorSubject, Observable, Subscription, combineLatest} from 'rxjs';
 
 import {Coordinate} from 'app/models/geometry/coordinate';
 import {Geometry, GeometryType} from 'app/models/geometry/geometry';
@@ -66,6 +66,9 @@ const enlargedPolygonStrokeWeight = 6;
 })
 export class MapComponent implements AfterViewInit, OnChanges, OnDestroy {
   private subscription: Subscription = new Subscription();
+  private selectedJobId$: BehaviorSubject<String> = new BehaviorSubject<String>(
+    ''
+  );
   lois$: Observable<List<LocationOfInterest>>;
   loisMap: ImmutableMap<string, LocationOfInterest> = ImmutableMap();
   activeSurvey$: Observable<Survey>;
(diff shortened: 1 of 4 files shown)


web/src/app/pages/main-page-container/main-page/map/map.component.ts
```diff
@@ -124,17 +126,29 @@ export class MapComponent implements AfterViewInit, OnDestroy {
   }
 
   ngAfterViewInit() {
+    this.initMap();
+  }
+
+  ngOnChanges() {
+    this.initMap();
```
gino-m, line 133: If we call `initMap()` multiple times, we'll accumulate one subscription for each change. Instead, there are two approaches you could take:

1. Do what you're doing here, but also dispose of the previous subscription before creating a new one, or
2. Create an Observable, selectedJobId$, which you combine with the other streams in `initMap()`.

Wdyt?

web/src/app/pages/main-page-container/main-page/map/map.component.ts
```diff
@@ -124,17 +126,29 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
         this.navigationService.getLocationOfInterestId$(),
       ]).subscribe(([survey, lois, locationOfInterestId]) => {
         const loisMap = ImmutableMap(
-          (this.showPredefinedLoisOnly
-            ? lois.filter(loi => loi.predefined !== false)
-            : lois
-          ).map(loi => [loi.id, loi])
+          lois
+            .filter(loi =>
+              this.showPredefinedLoisOnly ? loi.predefined !== false : loi
```
gino-m, line 146: Can `loi` ever be falsey? If so, why not just return `loi => loi.predefined !== false` as before?

web/src/app/pages/main-page-container/main-page/map/map.component.ts
```diff
@@ -124,17 +126,29 @@ export class MapComponent implements AfterViewInit, OnDestroy {
...
+            )
+            .filter(loi =>
+              this.selectedJobId !== '' ? loi.jobId === this.selectedJobId : loi
```
gino-m, line 149: Same question here. Can this be rewritten without evaluating the truthiness of `loi`?

gino-m (changes requested):

web/src/app/pages/main-page-container/main-page/map/map.component.ts
DaoyuT: RIght, I don't know why I wrote it that way, changed.

web/src/app/pages/main-page-container/main-page/map/map.component.ts
DaoyuT: Done.

gino-m (dismissed):
LGTM! 🚀

web/src/app/pages/main-page-container/main-page/map/map.component.ts
DaoyuT: Agreed, took option 2.

gino-m (approved):

PR #1799, opened 2024-05-24
Add Capture Location Task By Default

### LINKED ISSUE #1735: [Task editor] Add "capture location" task by default when creating a new job
* Added just after "Add LOI" task if present
* Mark "Required" by default

@rfontanarosa FYI

hallsjj08:
Fixes #1735

Add a Capture Location task anytime a new job is created.

[Screencast from 05-24-2024 01:42:30 PM.webm](https://github.com/google/ground-platform/assets/4734964/48f167b7-2864-42dc-abcc-2bd4eb09e9dd)


web/src/app/components/tasks-editor/tasks-editor.component.ts
```diff
@@ -133,13 +139,13 @@ export class TasksEditorComponent {
     return this.formGroup.get('tasks') as FormArray;
   }
 
-  onTaskAdd(group: TaskGroup) {
+  onTaskAdd(group: TaskGroup, required = false) {
```
gino-m, line 142: Nit: This method was originally only meant to handle user "add task" events. Perhaps it would be clearer if you refactored the logic into an `addTask` method which is called both by `onTaskAdd` and `ngOnInit`.

web/src/app/components/tasks-editor/tasks-editor.component.ts
```diff
@@ -106,6 +106,12 @@ export class TasksEditorComponent {
 
   multipleChoiceTasks = List<Task>();
 
+  ngOnInit(): void {
+    if (!this.formArray.length) {
```
gino-m, line 110: Not sure this will work if an "addLoi" has been already been added. This happens when the user clicks "free form data collection" in the previous step. @rfontanarosa can you confirm?

gino-m (changes requested):
Thanks for the change! @rfontanarosa to follow up

gino-m:
@hallsjj08 Note it appears some tests are failing. Run `npm run test` in web/ dir to run locally.

web/src/app/components/tasks-editor/tasks-editor.component.ts
hallsjj08: You're right. The logic didn't work for already pre-existing tasks. The flow that happened in the debugger was as follows:

1. ngOnChange -> No tasks exists yet
2. ngOnInit  -> initialize the capture location task. This was a problem because firebase hasn't yet pulled existing tasks.
3. ngOnChange -> firebase finished pulling task data from firestore which caused any existing forms in the formArray to be overwritten; thus, losing the default capture location task.

So I updated the PR to add a capture location task anytime a new job has been created.

web/src/app/components/tasks-editor/tasks-editor.component.ts
hallsjj08: This comment no longer applies due to the new changes if approved.

web/package.json
```diff
@@ -50,6 +50,7 @@
     "zone.js": "^0.14.5"
   },
   "devDependencies": {
+    "@types/ground-schema": "file:../schema/dist/generated",
```
gino-m, line 53: Was this added back by accident on merge?

gino-m (changes requested):
Overall LGTM! Could you kindly verify the following:

- [ ] Capture Location task is added to the 1st position when selecting free-form data collection during the survey creation flow. The "Drop a pin" task should be in the 2nd position.
- [ ] That the "Drop a pin" task is added/removed to the correct position when toggling free-form data collection under "Sites" In the "manage survey" view.

Thanks again!

web/package.json
gino-m: I removed it, please disregard.

web/src/app/components/tasks-editor/tasks-editor.component.ts
gino-m: Thank you for fixing that, and sorry for the delay!

PR #1815, opened 2024-06-04
Fixed wrong compiler options

rfontanarosa:
@types/googlemaps has been changed into @types/google.maps ,but the compilerOptions wasn't updated.


gino-m (approved):
⭐

PR #1818, opened 2024-06-04
Refactor proto codegen into proto/ 

gino-m:
This also configures a custom spec reporter to help debugging failures.


gino-m:
@rfontanarosa PTAL?

rfontanarosa (approved):

PR #1819, opened 2024-06-04
Filter LOIs by role/user

### LINKED ISSUE #1800: Data collector can see other's collector's record
**Data collector can see other's collector's record**
And they shouldn't

image

rfontanarosa:
Towards #1800

https://github.com/google/ground-platform/assets/10047945/6311c363-7273-4a2e-8925-7f64d836bf93


web/src/app/services/data-store/data-store.service.ts
```diff
@@ -311,20 +311,35 @@ export class DataStoreService {
...
+        ref
+          .where('predefined', '==', false)
+          .where('created.user.email', '==', email)
+    );
+
+    return combineLatest([
+      predefinedLois.valueChanges({idField: 'id'}),
+      userLois.valueChanges({idField: 'id'}),
+    ]).pipe(
+      map(lois => {
```
gino-m, line 332: Does array arg destructuring work the way I think it does? If so you can do `map([predefinedLois, userLois] => ...)`

web/src/app/services/loi/loi.service.ts
```diff
@@ -39,24 +41,35 @@ export class LocationOfInterestService {
...
+            )
         )
       );
 
     this.selectedLocationOfInterest$ = this.selectedLocationOfInterestId$.pipe(
       switchMap(loiId =>
-        surveyService.getActiveSurvey$().pipe(
-          switchMap(survey => dataStore.lois$(survey)),
-          map(lois => lois.find(loi => loi.id === loiId)!)
+        this.authService.getUser$().pipe(
```
gino-m, line 66: Can we rename this `selectedLoi$` while we're at it for consistency with other field and var names?

web/src/app/services/loi/loi.service.ts
```diff
@@ -39,24 +41,35 @@ export class LocationOfInterestService {
...
+          switchMap(user =>
+            this.surveyService.getActiveSurvey$().pipe(
```
gino-m, line 68: Instead of using active Survey and lois$, can we just filter `this.lois$`?

gino-m (changes requested):
Great work! A few small improvements, otherwise LG!

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -312,34 +313,68 @@ export class DataStoreService {
...
+            );
+
+            return combineLatest([
+              predefinedLois.valueChanges({idField: 'id'}),
+              userLois.valueChanges({idField: 'id'}),
+            ]).pipe(
+              map(([predefinedLois, userLois]) => {
+                const combined = predefinedLois.concat(userLois);
+                return List(
+                  combined
```
gino-m, line 346: Sorry I have a short memory - to remember what is combined here. Let's just call it something like `accessibleLois`, and rename the method to match?

Also, can we update this method to just take `surveyId` and `userEmail`? (principle of least privelege)

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -312,34 +313,68 @@ export class DataStoreService {
...
+            );
+          } else {
+            return this.db
+              .collection(`${SURVEYS_COLLECTION_NAME}/${id}/lois`)
+              .valueChanges({idField: 'id'})
+              .pipe(
+                map(array =>
+                  List(
+                    array
+                      .map(obj =>
```
gino-m, line 365: This map filter map operation is repeated in both conditions; can we move it outside?

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -312,34 +313,68 @@ export class DataStoreService {
...
-        return List(
-          combined
-            .map(obj => LoiDataConverter.toLocationOfInterest(obj.id, obj))
-            .filter(DataStoreService.filterAndLogError<LocationOfInterest>)
-            .map(loi => loi as LocationOfInterest)
-        );
-      })
-    );
+          .where(documentId(), '==', id)
+          .where(new FieldPath('acl', email), '==', Role.SURVEY_ORGANIZER)
```
gino-m, line 320: We have the survey in the method args, can we get the acl from there instead of loading it again? Ideally we would pass the user's role into the method.

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -312,34 +313,68 @@ export class DataStoreService {
...
   lois$({id}: Survey, {email}: User): Observable<List<LocationOfInterest>> {
-    const predefinedLois = this.db.collection(
-      `${SURVEYS_COLLECTION_NAME}/${id}/lois`,
-      ref => ref.where('predefined', '==', true)
-    );
-
-    const userLois = this.db.collection(
-      `${SURVEYS_COLLECTION_NAME}/${id}/lois`,
-      ref =>
+    return this.db
```
gino-m, line 316: This method has become unwieldy. Please consider refactoring the creation of the different streams into their own methods.

gino-m (changes requested):
A few more, thanks!

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -313,7 +313,7 @@ export class DataStoreService {
   }
 
   getAccessibleLois$(
-    {id: surveyId, acl}: Survey,
+    {id: surveyId}: Survey,
```
gino-m, line 316: If we only need `surveyId`, let's pass that instead of requiring the whole `Survey` object to be passed.

gino-m (changes requested):

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -317,7 +317,7 @@ export class DataStoreService {
     userEmail: string,
     canManageSurvey: boolean
   ): Observable<List<LocationOfInterest>> {
-    const toLocationOfInterests = (
+    const toLocationsOfInterest = (
```
gino-m, line 320: `s/toLocationsOfInterest/toLois/g`

Also, you can define this as a `private function` instead of inline lambda.

gino-m (changes requested):

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -312,26 +312,34 @@ export class DataStoreService {
       .pipe(map(data => FirebaseDataConverter.toUser(data as DocumentData)));
   }
 
+  private toLocationsOfInterest = (
```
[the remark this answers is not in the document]
gino-m: Ie. = and => not needed

gino-m (approved):
LGTM!

Issue #1699, opened 2024-04-03
[Code health] Set up end to end testing (local web, mobile, and database emulation)


sufyanAbbasi:
Mirrors https://github.com/google/ground-android/issues/2420

Formulating an overall strategy to run end to end testing on the mobile and web app. Will need to configure something like Docker or Github Hosted Runners which will spin up a local web server + Selenium WebDriver, local Android emulator + app, and a set up local Firebase database/emulator then run through a set of core tests.

gino-m:
Next steps:

- [ ] Get full flow (web+Android) working on GitHub Actions to size memory and runtime
- [ ] Split into two (web + Android), testing Firestore state in between
- [ ] Add additional test coverage (CUJs and screen rotation)

sufyanAbbasi:
Update: With the submission of #1820, we now have the ability to run three commands in PR comments:

- `/createTest` will run through the survey creation flow
- `/verifyTest` will run through the survey submission verification flow against the `data/test-submit` local snapshot
- `/e2eTest` will run the create test, then pass through data to the ground-android's submit test flow, then will pass that data through to the verify test.
    - Unfortunately, there is still a bug with the emulator that we need to fix up
 
 Next up, we can add testing for other CUJs like offline imagery download, or other Android life cycle consistency triggers like screen rotation.

PR #1820, opened 2024-06-05
Add Github Actions to run end-to-end tests, and also hermetically

sufyanAbbasi:
Adds two new actions:

1. Commenting `/e2eTest` in a PR runs the local `create-test` action, then the `ground-android` `submit-test`, then the local `verify-test` action, all feeding data into each other.
    1. Note that the `submit-test` still has some emulator problems, will need to be addressed later.
1. Commenting `/createTest` or `/submitTest` (or both in one comment) in a PR will trigger just the `create-test` or `submit-test` actions respectively which will run on checked in repo data.
    1. Note that create test can be flaky sometimes, still working out the issues there. 

See test PR: https://github.com/sufyanAbbasi/ground-platform/pull/1 and actions:
  - https://github.com/sufyanAbbasi/ground-platform/actions/runs/9359329129
  - https://github.com/sufyanAbbasi/ground-platform/actions/runs/9359391638

Addresses: #1699


os-micmec (approved):

e2e-tests/spec/create-survey.spec.ts
```diff
@@ -45,15 +45,45 @@ describe('create survey flow', () => {
...
+        JOB_NAME,
+        AD_HOC
+      );
+    } catch (e) {
+      console.error(e);
+      console.log((e as Error).stack);
+      throw new Error(`Unable to setSurveyMetadata: ${(e as Error).message}`);
+    }
+    try {
+      await helper.addAllTasks(AD_HOC ? LOI_TASK_TYPE : null);
```
gino-m, line 68: This method might be more readable if try/catch blocks are inside each helper.

gino-m (changes requested):
@sufyanAbbasi Great work, I know this wasn't easy to get set up! Does the check appear as failing until the /e2etest comment is added? If not, it might be easy to forget to trigger it on each PR. Is there a way to do that?

e2e-tests/spec/create-survey.spec.ts
sufyanAbbasi: Good call, done.

sufyanAbbasi:
> Does the check appear as failing
I'm not sure how the checks are configured, but from what I can tell, it doesn't block the PR from being merged.

> If not, it might be easy to forget to trigger it on each PR. Is there a way to do that?
That's a good idea though, we can configure it once we've worked out all the issues, but right now due to the flakiness and potential cost, it's worth leaving it up to individuals who feel inclined to run it.

e2e-tests/package.json
```diff
@@ -6,10 +6,10 @@
   "scripts": {
     "clean": "rm -rf build node_modules",
     "build": "npm i",
-    "test:create": "tsx jasmine spec/create-survey.spec.ts",
```
gino-m, line 9: I'm a bit confused by `build` only `installing` and pretest` only compiling, and as to why we need `tsx` if we're calling `tsc` to compile. 

Here's how I'd expect things to work, please cmiiw;

```
"prebuild": "npm install",
"build": "npm run compile",
"pretest: "npm run build",
"test:create-survey": "jasmine dist/spec/create-survey.spec.ts"
"test:create-survey:debug": "npm run test:create-survey -- --inspect" <-- not sure this works
```

Wdyt?

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Root package for Ground hosted components",
   "scripts": {
     "clean": "rm -rf node_modules && for p in lib web functions e2e-tests; do npm run clean --prefix $p; done",
-    "build": "npm i && npm run build --prefix schema && npm run build --prefix functions && npm run build:$npm_config_variant --prefix web && npm run build --prefix e2e-tests",
+    "build": "npm i && npm run build --prefix functions && npm run build:$npm_config_variant --prefix web && npm run build --prefix e2e-tests",
```
gino-m, line 7: Thanks!

gino-m (changes requested):

e2e-tests/webdriver-helpers.ts
```diff
@@ -95,141 +101,185 @@ export class WebDriverHelper {
...
-      } catch (e) {
-        lastError = e as Error;
-      }
-    } while (tries-- > 0);
-    fail(`No survey submissions appeared: ${lastError}`);
+      } while (tries-- > 0);
+      fail(`No survey submissions appeared: ${lastError}`);
+    } catch (e) {
+      console.error(e);
+      console.log((e as Error).stack);
```
gino-m, line 280: I believe .error() shows the stack already - we shouldn't need both .log and .error?

e2e-tests/package.json
sufyanAbbasi: Ah, sorry for the random boilerplate...`tsx`  lets you run `.ts` files directly in node without any compile step in between, plus supports the `--inspect` flag natively (otherwise, you'd need to somehow invoke it with a `node` command or pass it in from the top level or something). The quirk is that you have to refer to the `node_modules` version of jasmine, but I was unable to get this to work with the normal compile --> js version (was receiving "jasmine is not a module" errors).

I did remove the unecessary compile step to simplify it down to its most necessary parts.

e2e-tests/webdriver-helpers.ts
sufyanAbbasi: You are totally right! Done.

.github/actions/create-test/action.yml
```diff
@@ -0,0 +1,100 @@
...
+      shell: bash
+      run: echo "CHROME_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> $GITHUB_ENV
+
+    - name: Install nx-linux-x64-gnu
+      shell: bash
+      run: npm install @nx/nx-linux-x64-gnu
+
+    - name: Install dependencies
+      shell: bash
+      run: npm run build:local
```
gino-m, line 82: Given the desc I'd expect this to be `npm install`. Can we update desc to something like "Install and build"?

.github/actions/create-test/action.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    - name: Run create e2e tests
```
gino-m, line 84: Nit: At first glance I think this would actually create the tests - perhaps we can just call these various steps "step 1, 2," etc, or refer to them by a more complete name like "'Create survey' test"?

.github/actions/create-test/action.yml
```diff
@@ -0,0 +1,100 @@
...
+        # npm cache files are stored in `~/.npm` on Linux/macOS
+        path: ~/.npm
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-build-${{ env.cache-name }}-
+          ${{ runner.os }}-build-
+          ${{ runner.os }}-
+
+    - if: ${{ steps.cache-npm.outputs.cache-hit != 'true' }}
+      name: List the state of node modules
```
gino-m, line 61: Nit: "Node" capitalized for consistency

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: actions/checkout@v4
+      - name: Run create-test
+        uses: ./.github/actions/create-test
+        with:
+          platform-repository: ${{ github.repository }}
+          upload-artifacts: true
+
+  submitTest:
+    needs: createTest
+    name: Submit to survey
```
gino-m, line 48: Nit: "Submit data to survey", or even better "Collect and submit data"

gino-m (changes requested):
Overall LGTM. Sorry for late nits!

.github/actions/create-test/action.yml
sufyanAbbasi: Good call, done.

.github/actions/create-test/action.yml
sufyanAbbasi: Done for both files.

.github/workflows/test-e2e.yml
sufyanAbbasi: Nice! Done.

.github/actions/create-test/action.yml
sufyanAbbasi: Done for both files.

sufyanAbbasi:
Added support for default tasks using the config and some slight readability refactors there

gino-m (dismissed):
🧪 !

sufyanAbbasi:
Happy to report that they work! 

Invoked in a test PR via commenting: `/createTest`
https://github.com/sufyanAbbasi/ground-platform/actions/runs/9471364160/job/26094424106
|
Invoked in a test PR via commenting: `/verifyTest`
https://github.com/sufyanAbbasi/ground-platform/actions/runs/9471364153/job/26094424029

.github/actions/create-test/action.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    - name: Copy Firebase emulator data
+      if: inputs.upload-artifacts == 'true'
+      uses: actions/upload-artifact@v4
+      with:
+        name: data-create
+        path: '**/test'
+        retention-days: 7
+        overwrite: true
+        if-no-files-found: error
```
os-micmec, line 100: nit: can we add new line at the end?

os-micmec (dismissed):

os-micmec (approved):

.github/actions/create-test/action.yml
sufyanAbbasi: Done.

Issue #1758, opened 2024-04-29
[PROJECT] New proto-backed Firestore structure


jcqli:
Parent issue tracking the work needed to introduce protos as our db and Firestore schema.

- [x] Implement proto converters for Android (https://github.com/google/ground-android/issues/2049)
- [x] #1933
- [x] #1934
- [x] https://github.com/google/ground-platform/issues/1859
- [x] https://github.com/google/ground-platform/issues/1775
- [x] https://github.com/google/ground-android/issues/2500
- [x] https://github.com/google/ground-android/issues/2501
- [x] #1778
- [x] https://github.com/google/ground-platform/issues/1857
- [x] https://github.com/google/ground-android/issues/2498
- [x] https://github.com/google/ground-android/issues/2499
- [x] https://github.com/google/ground-android/issues/2494
- [x] https://github.com/google/ground-android/issues/2548
- [x] #1779
- [x] #1780
- [ ] #1781 
- [x] https://github.com/google/ground-android/issues/2468
- [ ] https://github.com/google/ground-platform/issues/1858

Additional tasks related to this migration:

- [x] https://github.com/google/ground-platform/issues/1884
- [x] https://github.com/google/ground-platform/issues/1911
- [x] https://github.com/google/ground-platform/issues/1922
- [x] https://github.com/google/ground-platform/issues/1921
- [x] https://github.com/google/ground-platform/issues/1925
- [x] https://github.com/google/ground-android/issues/2556

gino-m:
Did some research on various protoc plugins for generating js+ts code from protodef. We need something which makes metadata accessible (esp. field numbers) in order to implement the representation described in #2455. Here are my findings:

### protobufjs ✅ 

The most popular lib I found (16M weekly downloads), it can generate js and ts, but to get what we want you need to call it three times - once to generate the js, once to generate to ts bindings, and once to generate JSON with pb metadata.

```
npm i -D protobufjs protobufjs-cli
npx pbjs -t static-module ../proto/survey.proto -o dist/survey.js
npx pbts dist/survey.js -o dist/survey.d.ts
npx pbjs -t json ../proto/survey.proto -o dist/survey.proto.json
```

We could load the JSON at runtime to pick out field numbers we need.

#### ts-proto

OTOO 300k weekly downloads, and doesn't encode the field numbers anywhere we can use at runtime, so we can exclude this one for now.

#### protoc-gen-ts

It seems the `protoc` packaged with the one didn't support my M1 Mac.

[...]

PR #1825, opened 2024-06-10
Write new surveys using proto-based Firestore representation

rfontanarosa:
Towards #1758

Screenshot 2024-06-10 alle 11 06 39


web/src/app/converters/firebase-data-converter.ts
```diff
@@ -113,15 +117,27 @@ export class FirebaseDataConverter {
...
+      new Pb.Survey({
+        name,
+        description,
+        acl: {
+          [ownerEmail]: Pb.Role.SURVEY_ORGANIZER,
+        },
+      })
+    );
+
+    //TODO return data
```
gino-m, line 134: Accidental TODO?

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -449,7 +449,7 @@ export class DataStoreService {
    */
   async createSurvey(
     ownerEmail: string,
-    title: string,
+    name: string,
```
gino-m, line 452: Maybe add a basic test for `createSurvey`?

gino-m (changes requested):

Issue #1135, opened 2023-08-25
Clean up styles and migrate to M3


gino-m:
In approximate order:

- [x] Upgrade to the latest Angular ASAP so we minimize deltas with Material 3 once released.
- [ ] For now, only make styling tweaks related to font, color, and spacing.
- [ ] Font and color could be defined in global styles.
- [ ] Upgrade Angular again once M3 is release
- [ ] Refactor existing styles
- [ ] Tweak styles as needed

gino-m:
This is a bigger project so moving to next milestone. @amysorto please cmiiw.

gino-m:
Depends on #965

amysorto:
Unblocked now since we are on v17

amysorto:
As a note to myself, break this issue down into subtasks once I look through the code

kenstershiro:
@amysorto could you provide any details on the latest state of where this was left off?

kenstershiro:
Closing as we seem to be on M3 now

PR #1832, opened 2024-06-12
Upgrade to M3 theming

amysorto:
Part of #1135
This PR does the following:
- Creates a custom M3 theme (result from [custom theme schematic](https://material.angular.io/guide/theming#custom-theme))
- Update `ground-theme.scss` to use the custom theme instead
- Update styles where it was accessing the theme object values (M2 has a totally different color and typography object which is why this PR has a lot of diffs).
- Add styles for text that that needed custom typography styles to match the spec.

The outstanding thing after this is that there are still a bit of hard coded values for color and typography scattered across the code. Although that isn't urgent and this PR has the majority of changes anyways . Changing those hardcoded values is useful to fix later for code health and readability.


web/src/m3-theme.scss
```diff
@@ -0,0 +1,148 @@
+// This file was generated by running 'ng generate @angular/material:m3-theme'.
```
gino-m, line 1: Please add copyright header.

gino-m (changes requested):
One nit, otherwise LGTM. Glad we can finally refer to colors and themes by name! @rfontanarosa FYI.

gino-m (approved):
🎨 💯 🥳

PR #1861, opened 2024-06-18
Refactor spec reporter and set up source maps

gino-m:
..so stack traces refer to .ts source instead of compiled .js.

@rfontanarosa PTAL?


rfontanarosa (approved):

gino-m:
@rfontanarosa Thanks!

Issue #1857, opened 2024-06-17
Use protos when reading submissions


jcqli:
Tracked in https://github.com/google/ground-platform/issues/1758

jcqli:
@rfontanarosa Can you take a look at this one?

PR #1881, opened 2024-06-28
Read LOIs from Firestore using new proto-based format

### LINKED ISSUE #1778: Use protos when reading LOIs
To make the transition from the old schema to the new smoother, we can use the new fields if they're provided, otherwise revert to the old method. 

To do this, we can look for `job_id` (data['2']); if it's present, we convert to proto, if not we use the legacy code. @rfontanarosa Wdyt?

gino-m:
Fixes #1778 
Towards #1758


gino-m:
@amysorto Feel free to use this as a template for #1857.

amysorto (dismissed):

PR #1882, opened 2024-06-28
Remove unused LOI subclasses

gino-m:
This PR also removes unused code related to adding points via drawing tools and related tests. We don't full excise drawing tools code since it's quite certain they will be needed in the future, but the backend impl will have changed so much by then it doesn't make sense to carry the old serialization code forward.

@DaoyuT or @nwkotto PTAL?

### DIFF
--- a/web/src/app/models/loi.model.ts
+++ b/web/src/app/models/loi.model.ts
@@ -14,70 +14,11 @@
  * limitations under the License.
  */
 
-import {GeoPoint} from 'firebase/firestore';
 import {List, Map} from 'immutable';
 
 import {Geometry} from './geometry/geometry';
 
-export abstract class LocationOfInterest {
-  abstract readonly id: string;
-  abstract readonly jobId: string;
-  abstract readonly name?: string;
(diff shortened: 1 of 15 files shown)


web/src/app/models/loi.model.ts
```diff
@@ -86,4 +27,12 @@ export class GenericLocationOfInterest implements LocationOfInterest {
     readonly customId: string = '',
     readonly predefined: boolean = true
   ) {}
+
+  static getSmallestByArea(lois: List<LocationOfInterest>): LocationOfInterest {
```
nwkotto, line 31: Is the diff just getting confused, and this is actually the same method in same class? Otherwise I would ask if this is getting tested.

nwkotto (approved):

web/src/app/models/loi.model.ts
gino-m: This is basically a merged version of LocationOfInterest and GenericLocationOfInterest. We're punting on tests on web for now due to lack of developer coverage, but we'll pick up adding coverage once the community project is ramped up.

PR #1896, opened 2024-07-08
Add shx and cross-env dependencies to make it compatible with Windows

rfontanarosa:
towards #1438


nwkotto (approved):

functions/package.json
```diff
@@ -4,14 +4,14 @@
...
-    "clean": "rm -rf dist *.log",
+    "clean": "shx rm -rf dist *.log",
     "ci-all": "npm run ci-local-deps && npm ci",
     "ci-local-deps": "npm ci --workspace=../proto --workspace=../lib",
     "lint": "eslint --ext .js,.ts src/",
     "lint:fix": "npm run lint -- --fix",
     "shell": "firebase functions:shell",
     "emulators": "firebase emulators:start --only functions,firestore --project local",
-    "deploy": "firebase login && firebase deploy --project $npm_config_project --only functions",
+    "deploy": "firebase login && cross-env-shell firebase deploy --project $npm_config_project --only functions",
```
gino-m, line 14: Why is `cross-env-shell` needed fir `firebase deploy` but not `firebase login` and other local commands?

gino-m (changes requested):
One Q, otherwise LGTM

functions/package.json
rfontanarosa: there cross-env-shell is used to make $npm_config_project compatible with command prompt or powershell (they want %npm_config_project%)

gino-m: I feel like if we do this for one we should do this for all, since it's not obvious which commands are compatible and which aren't, but I think this ok as a patch for now.

One question - how does `&&` behave in Windows? Shouldn't that also be wrapped by `cross-env-shell`?

gino-m (changes requested):
I believe we don't need these changes if using GitBash. Let's hold off on these to make sure they're really necessary.

gino-m:
Closing this in favor of #1946.

PR #1898, opened 2024-07-09
Fix Property Generator when property is an Array/Object

### LINKED ISSUE #1897: WHISP data missing on new plots
Recently recorded plots are not showing WHISP data, just the name. Seems to be an issue in the layer between the registry and Ground

rfontanarosa:
closes #1897

### DIFF
--- a/functions/src/on-create-loi.ts
+++ b/functions/src/on-create-loi.ts
@@ -52,12 +52,9 @@ export async function onCreateLoiHandler(
       wkt
     );
 
-    Object.keys(properties).forEach(key => {
-      properties[key] =
-        typeof properties[key] === 'object'
-          ? JSON.stringify(properties[key])
-          : properties[key];
-    });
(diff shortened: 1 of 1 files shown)


functions/src/on-create-loi.ts
```diff
@@ -51,6 +51,13 @@ export async function onCreateLoiHandler(
       properties,
       wkt
     );
+
+    Object.keys(properties).forEach(key => {
```
gino-m, line 55: You could also use `filter()` to only modify the object properties and leave the rest along.

gino-m: e.g. 
```ts
Object.keys(properties)
  .filter(key => typeof properties[key] === 'object')
  .forEach(key => { properties[key] = JSON.stringify(properties[key]); }
```

Does this work?

gino-m (approved):

PR #1900, opened 2024-07-10
Fix canImport permission

rfontanarosa:
@gino-m PTAL?

 (survey organizer was not able to import geojson)

### DIFF
--- a/functions/src/common/auth.ts
+++ b/functions/src/common/auth.ts
@@ -23,7 +23,7 @@ import {EmulatorIdToken} from '../handlers';
 // https://firebase.google.com/docs/hosting/manage-cache#using_cookies
 export const SESSION_COOKIE_NAME = '__session';
 export const OWNER_ROLE = 'OWNER';
-const DATA_COLLECTOR_ROLE = 'DATA_COLLECTOR';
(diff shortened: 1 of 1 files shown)


nwkotto (approved):
The change itself looks pretty straightforward, but where are these roles defined? Where did you see where to change this to? Just asking so I can reference that.

rfontanarosa:
> The change itself looks pretty straightforward, but where are these roles defined? Where did you see where to change this to? Just asking so I can reference that.

I hadn't noticed this before, but at survey creation the assigned role is "OWNER." However, if I understand correctly, the "OWNER" role will be deprecated in favor of "SURVEY_ORGANIZER."

Additionally, the "DATA_COLLECTOR" role cannot edit surveys. This means they won't have access to the "IMPORT SITES" button in the interface.

nwkotto:
That makes sense to me. But how did you know that 'SURVEY_ORGANIZER' was the correct string to use for that role? Is that defined in Firebase somwhere?

rfontanarosa:
nope, it's not defined in firebase, it was defined by the applications (web/android/functions), but soon the will be defined by proto model (and so centralised)

Issue #1884, opened 2024-06-28
[Edit survey] "Publish changes" doesn't write survey in new format


gino-m:
Changes to the survey are not reflected in the database using the new proto-based format.

@rfontanarosa PTAL?

PR #1902, opened 2024-07-12
Save task changes into proto format

### LINKED ISSUE #1899: MultipleChoiceSelection Condition Proto is missing task ID field
Here we define a list of option IDs for selection in the task condition:

https://github.com/google/ground-platform/blob/3581003c6d610b15f0af4dd1313449608d7bc9d7/proto/src/google/ground/v1beta1/job.proto#L245

However, we require a task ID to know whose options they belong to.

The fix is to simply add a `string task_id = 2` field.

rfontanarosa:
towards #1884
closes #1899

### DIFF
--- a/web/src/app/converters/proto-model-converter.ts
+++ b/web/src/app/converters/proto-model-converter.ts
@@ -17,7 +17,7 @@
 import {DocumentData} from '@angular/fire/firestore';
 import {toDocumentData} from '@ground/lib';
 import {GroundProtos} from '@ground/proto';
-import {Map} from 'immutable';
+import {List, Map} from 'immutable';
 
 import {Job} from 'app/models/job.model';
 import {Role} from 'app/models/role.model';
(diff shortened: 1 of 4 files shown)


web/src/app/converters/proto-model-converter.ts
```diff
@@ -225,3 +210,25 @@ function taskConditionToPartialMessage(task: Task): Pb.Task.ICondition[] {
       .toArray() || []
   );
 }
+
+/**
+ * Returns the proto representation of a list of Task model objects.
+ */
+export function tasksToDocument(tasks: List<Task>): Pb.ITask[] {
```
gino-m, line 217: These methods return proto Messages, not Firestore Documents. Please rename to `*ToMessage` where appropriate.

web/src/app/converters/proto-model-converter.ts
```diff
@@ -99,28 +99,13 @@ export function jobToDocument(job: Job): DocumentData {
...
-        .toList()
-        .toArray(),
+      tasks: tasksToDocument(job.tasks?.toList() || List([])),
     })
   );
 }
 
 /**
- * Creates a partial rapresentation of a Task.
+ * Returns the proto representation of a partial Task model object.
```
gino-m, line 108: Why do we have so many separate methods to handle "partial" model objects? Aren't partial objects just regular objects with some values just set to `null`, or in the case of protos, being unset?

gino-m (changes requested):
A few suggestions, thank you!

web/src/app/converters/proto-model-converter.ts
```diff
@@ -99,17 +103,22 @@ export function jobToDocument(job: Job): DocumentData {
       index: job.index,
       name: job.name,
       style: new Pb.Style({color: job.color}),
-      tasks: tasksToDocument(job.tasks?.toList() || List([])),
+      tasks: toTasksMessage(job.tasks?.toList() || List([])),
```
gino-m, line 106: Why not just handle the list conversion here and have the method just be `toTaskMessage()`?

gino-m (changes requested):

gino-m (changes requested):

gino-m (approved):
Nice work!

Issue #1779, opened 2024-05-13
Use protos when exporting to CSV


jcqli:
Download CSV functions will need be updated to use the new schema. To do this, we can use the shared protos and converter lib being added to lib/ in this project.

Tracked in https://github.com/google/ground-platform/issues/1758

jcqli:
@gino-m Could you add more detail to this?

gino-m:
> @gino-m Could you add more detail to this?

Added to desc.

PR #1917, opened 2024-07-17
[Export CSV] Export data stored in new proto-based format

gino-m:
Towards #1779.

One merged, I'll send a separate PR to read the `Survey` metadata in the new proto format when exporting.


functions/src/export-csv.ts
```diff
@@ -86,23 +87,46 @@ export async function exportCsvHandler(
...
   const submissionsByLoi = await getSubmissionsByLoi(survey.id, jobId);
 
   loiDocs.forEach(loiDoc => {
-    submissionsByLoi[loiDoc.id]?.forEach(submission =>
-      writeRow(csvStream, loiProperties, tasks, loiDoc, submission)
+    const loi = toMessage(loiDoc.data(), Pb.LocationOfInterest);
+    if (loi instanceof Error) {
+      throw loi;
+    }
+    const submissions = submissionsByLoi[loiDoc.id] || [{}];
```
sufyanAbbasi, line 103: Why not default to empty list and skip the `forEach`?

functions/src/export-csv.ts
```diff
@@ -145,130 +170,149 @@ function writeRow(
...
-  return geojsonToWKT(Datastore.fromFirestoreMap(geometryObject));
+function quote(value: any): string {
+  if (value === null || value === undefined) {
+    return '';
+  }
+  if (typeof value === 'number') {
+    return value.toString();
+  }
+  const escaped = value.toString().replaceAll('"', '""');
+  return '"' + escaped + '"';
```
sufyanAbbasi, line 206: \`"${escaped}"\`

functions/src/export-csv.ts
```diff
@@ -145,130 +170,149 @@ function writeRow(
...
+  taskId: string,
+  task: Task,
+  data: Pb.ITaskData[]
+): string | number | null {
+  const result = data.find(d => d.taskId === taskId);
+  if (!result) {
+    return null;
+  }
+  if (result.textResponse) {
+    return result.textResponse.text || null;
```
sufyanAbbasi, line 226: Better to use `??` to allow for empty string response.

functions/src/export-csv.ts
```diff
@@ -145,130 +170,149 @@ function writeRow(
...
+    return getPhotoUrlValue(result.takePhotoResult);
   } else {
-    return result;
+    return null;
   }
 }
 
+function getNumberValue(response: Pb.TaskData.INumberResponse): number | null {
+  const number = response.number;
+  if (number === undefined || number === null) {
```
sufyanAbbasi, line 254: Using non-strict equality, i.e., `number != null`, here is the one time that it's actually useful!

Or you can simplify this whole expression to:
```
return response.number ?? null 
```

functions/src/export-csv.ts
```diff
@@ -145,130 +170,149 @@ function writeRow(
...
+    return null;
+  }
+  return number;
+}
+
+function getDateTimeValue(
+  response: Pb.TaskData.IDateTimeResponse
+): string | null {
+  const seconds = response.dateTime?.seconds;
+  if (seconds === undefined || seconds === null) {
```
sufyanAbbasi, line 264: `seconds != null`

functions/src/export-csv.ts
```diff
@@ -145,130 +170,149 @@ function writeRow(
...
 function getPropertyNames(lois: QuerySnapshot<DocumentData>): Set<string> {
   return new Set(
-    lois.docs
-      .map(loi =>
-        Object.keys(loi.get('properties') || {})
-          // Don't retrieve ID because we already store it in a separate column
-          .filter(prop => prop !== 'id')
-      )
-      .flat()
+    lois.docs.map(loi => Object.keys(loi.get(l.properties) || {})).flat()
```
sufyanAbbasi, line 306: Can use `flatMap()` instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

proto/src/ground/v1beta1/audit_info.proto
```diff
@@ -40,4 +39,7 @@ message AuditInfo {
 
   // URL of the user's profile picture.
   string photo_url = 5;
+
+  // The user's email address.
+  string email_address = 6;
```
sufyanAbbasi, line 44: I should note that we will need to fix the app client to address this, since we were storing and querying by `user_id`, which may constitute a large/breaking change...

Can you file a bug on the Android side and I can quickly get this fixed?

proto/src/ground/v1beta1/submission.proto
```diff
@@ -102,7 +102,7 @@ message TaskData {
   // A manually entered number response.
   message NumberResponse {
     // The number provided by the user.
-    double number = 2;
+    double number = 1;
```
sufyanAbbasi, line 105: This will be a breaking change for any recently serialized surveys, which are all tests, but I think this is more correct.

sufyanAbbasi (approved):
Just nits and some interesting db/client sync questions with the proto changes. Otherwise, LGTM!

functions/src/export-csv.ts
gino-m: This effectively ensures we include one row for jobs without submissions (outer join on jobs). Definitely deserves a comment - adding one now!

functions/src/export-csv.ts
gino-m: Done.

functions/src/export-csv.ts
gino-m: Done.

functions/src/export-csv.ts
gino-m: Nice!

functions/src/export-csv.ts
gino-m: Good catch! Done.

proto/src/ground/v1beta1/audit_info.proto
gino-m: Querying for what, exactly? Ad hoc LOIs? If so, wouldn't that continue to work, since this is just an additive change?

proto/src/ground/v1beta1/submission.proto
gino-m: Ack. There's no real data in the db at the moment, and we haven't rolled out this change to prod, so I think this is safe.

gino-m (commented):
@sufyanAbbasi All comments resolved, modulo two follow-up questions. PTAL?

functions/src/export-csv.ts
sufyanAbbasi: Sorry, I should have been more clear! There's a funny ~quirk~ language feature that allows you to check for both `null` and `undefined` with non-strict equality like this: 
https://stackoverflow.com/a/16607837

I personally think that it's better to not explicitly check for both null and undefined unless the specifications of the object explicitly define null and undefined to represent different states (which is the 1% case), and opt for simpler, cleaner code to help JS hide it's mistake of putting in TWO nulls in its language ahah.

proto/src/ground/v1beta1/audit_info.proto
sufyanAbbasi: Ooh... I must have gotten the Survey's ACL field confused with AuditInfo. I think we should be ok! We'll just want to make sure we're clear which field is meant to be the primary key etc.

sufyanAbbasi (approved):

sufyanAbbasi (approved):
Thank you so much!!

sufyanAbbasi (approved):
Thank you++! Sorry we had to modify the linter to satisfy the odd hills I want to die on haha

Issue #1665, opened 2024-03-16
[Submission details] "Other" text breaks submission panel


gino-m:
Submission panel still fails on invalid data.

<img width="1728" alt="image" src="https://github.com/google/ground-platform/assets/228050/4284a5cc-5ae0-41d0-bf39-c45a33ec8862">

@rfontanarosa @amysorto @jcqli FYI

gino-m:
@amysorto Could you PTAL?

amysorto:
@gino-m Can you give me access to that survey? I'm trying to reproduce this issue and am unable to

gino-m:
@sufyanAbbasi Can you provide @amysorto with a repro case?

gino-m:
Fixing #1918 should fix this, assigning to @rfontanarosa to confirm.

rfontanarosa:
closed by #1918

PR #1918, opened 2024-07-17
Read Submissions from Firestore using new proto-based format

### LINKED ISSUE #1857: Use protos when reading submissions

rfontanarosa:
closes #1857 

Screenshot 2024-07-22 alle 17 16 19


web/src/app/converters/submission-data-converter.ts
```diff
@@ -0,0 +1,237 @@
+/**
+ * Copyright 2022 The Ground Authors.
```
gino-m, line 2: 2024 please

web/src/app/converters/submission-data-converter.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  private static toResult(
+    resultValue: number | string | List<string>,
+    task?: Task
+  ): Result | Error {
+    try {
+      if (typeof resultValue === 'string') {
+        return new Result(resultValue as string);
+      }
+      if (typeof resultValue === 'number') {
```
gino-m, line 203: Please chain with `else if` to make it clear there's no fall-through logic.

gino-m (changes requested):
I realize this is still in draft, providing some early feedback.

web/src/app/converters/submission-data-converter.ts
```diff
@@ -229,30 +229,42 @@ export class LegacySubmissionDataConverter {
...
           List(
-            resultValue
-              .filter(optionId => !optionId.startsWith('['))
-              .map(
-                optionId =>
+            resultValue.map(optionId => {
+              if (optionId.startsWith('['))
+                return new Option(
+                  'otherOption',
+                  'otherOption',
```
gino-m, line 252: Where do these special values `'otherOption'` come from? Can we refer to a const w/docs which explain their usage?

gino-m (changes requested):
LG! Just a nit or two.

sufyanAbbasi (dismissed):
Thank you! I'll approve so we can merge this and iterate on this further!

sufyanAbbasi:
Note that this solution isn't fully complete because we still need to change all the places whenever we query the firestore, e.g. `where('jobId', ...`, those need to change to the field number instead. I'm happy to fix that after merging this one in!

sufyanAbbasi:
Ok, I went ahead and filled in the field numbers for all the queries!

sufyanAbbasi (approved):

sufyanAbbasi:
So part of the issue was that the app was not sending up the correct timestamp, it multiplied milliseconds by 1000. That is now fixed, but it still looks like TimeStamps are not being properly serialized just yet.

sufyanAbbasi:
I think the fix is something like in `toFieldValue()` in `firestore-to-proto.ts`, the deserializer should looks something like this but I can't seem to get it to work right now:
```ts
  switch (fieldType) {
    case 'string':
    case 'int32':
    case 'int64':
    case 'bool':
    case 'double':
      return firestoreValue;
    case '.google.protobuf.Timestamp':
      return toMessageOrEnumValue(['google', 'protobuf'], 'Timestamp', firestoreValue);
    default:
      return toMessageOrEnumValue(messageTypePath, fieldType, firestoreValue);
```

sufyanAbbasi (dismissed):
Thanks so much, approved! I'll take a look at the Timestamp proto serialization problem.

sufyanAbbasi:
It worked! Pushed a fix to this branch:

image

sufyanAbbasi (approved):
Ok now we should be good to go!!

gino-m:
@rfontanarosa Does this fix #1665? If so please add 'Fixes #1665' to the description.

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -159,7 +137,11 @@ export class DataStoreService {
   loadAccessibleSurveys$(userEmail: string): Observable<List<Survey>> {
     return this.db
       .collection(SURVEYS_COLLECTION_NAME, ref =>
-        ref.where(new FieldPath(SURVEY_ACL_FIELD, userEmail), 'in', [1, 2, 3])
+        ref.where(
+          new FieldPath(FieldNumbers.Survey.acl, userEmail),
+          'in',
+          [1, 2, 3]
```
gino-m, line 143: Can we replace these with references to the enum values instead?

functions/src/common/datastore.ts
```diff
@@ -117,10 +119,13 @@ export class Datastore {
...
 
   fetchSubmissionsByJobId(surveyId: string, jobId: string) {
-    return this.db_
-      .collection(submissions(surveyId))
-      .where('jobId', '==', jobId)
-      .get();
+    return (
+      this.db_
+        .collection(submissions(surveyId))
+        // Field number for Submission.job_id.
```
gino-m, line 125: You can remove this and similar comments now.

functions/src/on-write-submission.spec.ts
```diff
@@ -62,7 +62,7 @@ describe('onWriteSubmission()', () => {
       .and.returnValue({
         where: jasmine
           .createSpy('where')
-          .withArgs('loiId', '==', loiId)
+          .withArgs('2', '==', loiId)
```
gino-m, line 65: Please use constant instead

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -130,7 +137,11 @@ export class DataStoreService {
   loadAccessibleSurveys$(userEmail: string): Observable<List<Survey>> {
     return this.db
       .collection(SURVEYS_COLLECTION_NAME, ref =>
-        ref.where(new FieldPath('acl', userEmail), 'in', Object.keys(Role))
+        ref.where(
+          new FieldPath(FieldNumbers.Survey.acl, userEmail),
+          'in',
+          [1, 2, 3]
```
gino-m, line 143: Also, doesn't excluding the string values from the query immediately break backwards compatibility?

gino-m (changes requested):
Allmosttt theere! :D

gino-m (changes requested):
Also, something is breaking the build of `functions/`:

```
TypeError: Cannot read properties of undefined (reading 'v1beta1')
    at Object.<anonymous> (/home/runner/work/ground-platform/ground-platform/functions/src/import-geojson.ts:30:40)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Function.Module._load (node:internal/modules/cjs/loader:101[9](https://github.com/google/ground-platform/actions/runs/10169772702/job/28127451209?pr=1918#step:6:10):12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (/home/runner/work/ground-platform/ground-platform/functions/src/import-geojson.spec.ts:41:1)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1422:[10](https://github.com/google/ground-platform/actions/runs/10169772702/job/28127451209?pr=1918#step:6:11))
npm error Lifecycle script `test` failed with error:
npm error Error: command failed
npm error   in workspace: @ground/functions@0.0.1
npm error   at location: /home/runner/work/ground-platform/ground-platform/functions
```

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -381,15 +397,20 @@ export class DataStoreService {
...
       ref =>
         ref
-          .where('predefined', '==', false)
-          .where('created.user.email', '==', userEmail)
+          .where(
+            FieldNumbers.LocationOfInterest.source,
+            '==',
+            Source.FIELD_DATA
+          )
+          .where(FieldNumbers.LocationOfInterest.owner_id, '==', userId)
```
gino-m, line 413: This and related changes will cause old submissions to stop showing once deployed. Is @lecrabe ok with us breaking existing surveys?

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -381,15 +397,20 @@ export class DataStoreService {
 
     const predefinedLois = this.db.collection(
       `${SURVEYS_COLLECTION_NAME}/${surveyId}/lois`,
-      ref => ref.where('predefined', 'in', [true, null])
+      ref =>
+        ref.where(FieldNumbers.LocationOfInterest.source, '==', Source.IMPORTED)
```
gino-m, line 401: Consider also updating the name of variable from `predefined` and `user` to `imported` and `fieldData` for consistency.

gino-m (changes requested):

web/src/app/services/data-store/data-store.service.ts
gino-m: We discussed. Let's merge but not deploy.

gino-m (approved):
Approved!

PR #1938, opened 2024-07-31
Add method to get field ids for a proto message class

gino-m:
@rfontanarosa FYI

Usage:

```ts
import {registry} from '@ground/lib';
import {GroundProtos} from '@ground/proto';

import Pb = GroundProtos.google.ground.v1beta1;
const s = registry.getFieldIds(Pb.Survey);
const survey = {
  [s.name]: 'Survey name',
  [s.description]: 'Description something something',
  ...
};
```


amysorto (approved):

amysorto (approved):

gino-m:
@rfontanarosa I still can't get this to work. For some reason when adding a new import of `GroundProtos` or `registry`, `GroundProtos` gets replaced with an invalid representation containing the following:

```js
{
  "google": {
    "protobuf": {
      "NullValue": {
        "NULL_VALUE": 0
      }
    },
    "events": {
      "cloud": {
        "firestore": {
          "v1": {}
        }
      }
    },
    "type": {}
  }
}
```

gino-m:
@rfontanarosa It seems there was a collision with `google.*` protos. Renaming proto packages to `ground.*` to fix. @shobhitagarwal1612 @sufyanAbbasi @scolsen A fix will also be needed on Android when we upgrade the proto dep, but the protos should be wire compatible even without.

rfontanarosa (approved):

Issue #1856, opened 2024-06-17
Per-survey data sharing agreement in create/edit survey


jcqli:
Adding agreement information in the create / edit survey flow

Have some multiple choice options or enabling custom text for survey creators to inform collectors of agreement

gino-m:
Detailed FR:

* When creating a survey, survey organizers must select one of the following data sharing agreement options which must be accepted by data collectors before they begin collecting data:
  * Private: Data will be shared with survey organizers only.
  * Public: Survey organizers may share and use data publicly with no constraints (perhaps under [CC0 license](https://creativecommons.org/public-domain/cc0/))
  * Custom agreement: The survey organizer will provide text which the data collector must accept before collecting data.

@vittorino @rawbzz @amegantz for your attention.

gino-m:
@vittorino Could you please paste the designs for changes to create/edit survey flow to allow the organizer to select or specify the data sharing agreement? See latest comment in Figma thread with @rawbzz for details.

rawbzz:
FYI, copy is still WIP

amegantz:
![image](https://github.com/user-attachments/assets/131e728d-a6d1-4920-87fe-9678bff2f07c)
![image](https://github.com/user-attachments/assets/80de73c3-1fdc-4213-b7ba-d4432c121a8d)
PTAL

gino-m:
@amegantz @vittorino This looks great!!! Perhaps it would be helpful to show the survey organizers the exact text the data collectors will see, along with an explanation that that's what they're looking at?

vittorino:
Do you think it's still unclear that the text is what they will see? If so, I would recommend explaining in the instructions rather than simulating a preview because it would be weird if we don't show the same for all 3 options. In my opinion we would be making the UI more complex than what it needs to be.

vittorino:
FYI, small change in the dialog (no horizontal margins) to match the design for mobile:
![image](https://github.com/user-attachments/assets/82cac570-33e2-4bfe-bef8-892541664c78)

amegantz:
> Do you think it's still unclear that the text is what they will see? If so, I would recommend explaining in the instructions rather than simulating a preview because it would be weird if we don't show the same for all 3 options. In my opinion we would be making the UI more complex than what it needs to be.

Like this?
![image](https://github.com/user-attachments/assets/d330be2c-635b-4e9d-b97e-b325c6fc7e42)

gino-m:
[...]

PR #1940, opened 2024-08-02
Update survey proto to support data sharing agreement

nwkotto:
Part 1 for https://github.com/google/ground-platform/issues/1856


proto/src/google/ground/v1beta1/survey.proto
```diff
@@ -37,6 +37,18 @@ message Survey {
 
   // Required. User id of the survey owner.
   string owner_id = 5;
+
+  // Properties related to sharing of survey data.
+  message DataSharing {
```
sufyanAbbasi, line 42: Maybe `DataSharingConsent dataSharingConsent` and `DataSharingConsentType`? Nouns for messages are preferable and making the "consent" part explicit differentiates it from ACLs.

sufyanAbbasi:
Awesome! And I believe we will also want an associated response field on the `Submission` too, since we'll want to save the user's response to the DB in case the survey organizers change the sharing consent after the fact. Maybe just an enum will suffice?

gino-m:
> Awesome! And I believe we will also want an associated response field on the `Submission` too, since we'll want to save the user's response to the DB in case the survey organizers change the sharing consent after the fact. Maybe just an enum will suffice?

Not sure about storing Submission - a Firestore query would scan all submissions to find out if one has the consent bit. That said, Survey isn't editable.. maybe a new collection specifically for this purpose? For a first plass we'll just ask once per session per survey, so nothing needs to be stored.

proto/src/google/ground/v1beta1/survey.proto
nwkotto: Sounds good! Updated the proto names per that suggestion.

nwkotto:
> > Awesome! And I believe we will also want an associated response field on the `Submission` too, since we'll want to save the user's response to the DB in case the survey organizers change the sharing consent after the fact. Maybe just an enum will suffice?
> 
> Not sure about storing Submission - a Firestore query would scan all submissions to find out if one has the consent bit. That said, Survey isn't editable.. maybe a new collection specifically for this purpose? For a first plass we'll just ask once per session per survey, so nothing needs to be stored.

So are we saying we're ok to not include any submission bit about consent being agreed to? Or how are we expecting that to be captured?

proto/src/google/ground/v1beta1/survey.proto
```diff
@@ -53,3 +65,15 @@ enum Role {
   // User may also manage the survey, update locations of interest, etc.
   SURVEY_ORGANIZER = 3;
 }
+
+// Defines the type of data sharing consent agreement the user can see.
+enum DataSharingConsentType {
```
gino-m, line 70: For consistency, we can move this inside `DataShareConsent` above the field where it's used and rename it to `Type`.

proto/src/google/ground/v1beta1/survey.proto
```diff
@@ -37,6 +37,18 @@ message Survey {
 
   // Required. User id of the survey owner.
   string owner_id = 5;
+
+  // Properties related to data sharing consent.
+  message DataSharingConsent {
```
gino-m, line 42: I think "data sharing terms" would be more correct than "consent" here. Consent can be reserved for once the data collectors actually agree to the terms.

proto/src/google/ground/v1beta1/survey.proto
```diff
@@ -37,6 +37,18 @@ message Survey {
...
+    // Type of data sharing consent agreement the user will see.
+    DataSharingConsentType type = 1;
+
+    // Text of custom data sharing consent terms.
+    string custom_text = 2;
+  }
+
+  // Defines how data sharing consent agreemnent is displayed to users.
+  DataSharing dataSharingConsent = 6;
```
gino-m, line 51: s/DataSharing/DataSharingTerms/

Also, dataSharingConsent -> lower_snake_case.

gino-m:
> > > Awesome! And I believe we will also want an associated response field on the `Submission` too, since we'll want to save the user's response to the DB in case the survey organizers change the sharing consent after the fact. Maybe just an enum will suffice?
> > 
> > 
> > Not sure about storing Submission - a Firestore query would scan all submissions to find out if one has the consent bit. That said, Survey isn't editable.. maybe a new collection specifically for this purpose? For a first plass we'll just ask once per session per survey, so nothing needs to be stored.
> 
> So are we saying we're ok to not include any submission bit about consent being agreed to? Or how are we expecting that to be captured?

Sorry, I think I misunderstood the original question. Adding a bit to submissions as to which consent was given is a good idea. For now the Android client will need to store that consent was given on the first data collection flow locally, in the future we may store that in the remote db as well.

gino-m (approved):
I made a few small adjustments, LGTM!

PR #1951, opened 2024-08-07
Use owner id when writing and authenticating LOIs and submissions

### LINKED ISSUE #1154: Add `owner` field to submissions and LOIs
This will be used by ACLs to grant data collectors access to create/edit/delete their own submissions and LOIs. 

This will allow us to correctly implement #870 (#1153). Currently we assume "creator == owner", which may or may not be a valid assumption.

@DaoyuT Wdyt?

gino-m:
Fixes #1154

### DIFF
--- a/firestore/firestore.rules
+++ b/firestore/firestore.rules
@@ -103,11 +103,17 @@
     }
 
     /**
-     * Returns true iff the user with the provided email created the specified resource.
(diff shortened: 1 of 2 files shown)


gino-m:
This is blocking other work so forcing merge to keep things moving forward.

PR #1957, opened 2024-08-08
Fix importing of LOI properties

### LINKED ISSUE #1953: [Import GeoJSON] LOI properties not imported correctly
Since characters values are appearing in place of imported properties on both LOIs with and without submissions, for example:

```
,"h","e",,"F","o",,,,,,,,,,,,,,,,
```

gino-m:
Fixes #1953


gino-m:
This is blocking other work so merging for now.

PR #1959, opened 2024-08-09
Added support for reading and writing data sharing terms from survey model

nwkotto:
- Added data sharing terms to survey model
- Added updateDataSharingTerms to DataStoreService
- Added dataSharingTermsToDocument to proto-model-converter
- Added updateDataSharingTerms to SurveyService
- Added conversion of data sharing terms to survey-data-converter
- Added data sharing agreement to tests using the survey model


gino-m (approved):
Nice progress! @rfontanarosa Are we planning to move away from making partial updates to the survey and jobs, since we write them all at once during Publish? Are these needed for the `create` flow?

PR #1968, opened 2024-08-12
Created Data Sharing Terms component and integrated into create flow

nwkotto:
Data Sharing Terms component...
- Takes in data sharing terms type and custom text to seed the form
- Only shows customText textarea when "Custom agreement" is selected
- Disables customText when custom agreement type is not selected
- Send validation status as an output

### DIFF
--- a/web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.html
+++ b/web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.html
@@ -24,9 +24,31 @@
     >
       <mat-card-content>
         <mat-radio-button class="option-radio-button" [value]="option.value">
-          <span class="data-sharing-type-label">{{ option.label }}</span>
+          <span class="data-sharing-label">{{ option.label }}</span>
           <p>{{ option.description }}</p>
         </mat-radio-button>
+        <div *ngIf="shouldShowCustomizeAgreementSection(option.value)">
+          <h3 id="customize-agreement-header"
+            class="field-header">
+            Customize agreement
+          </h3>
+          <p class="field-description">
+            Create a custom agreement which will be shown to data collectors before they can
+            collect data. Data collectors must agree to the terms of this agreement.
+          </p>
+          <mat-form-field class="custom-terms">
+            <textarea matInput
+              formControlName="customText"
+              placeholder="Enter the terms of your custom agreement..."
+              aria-labelledby="customize-agreement-header">
+            </textarea>
+            <mat-error *ngIf="customTextControl.touched && customTextControl.invalid">
+              <ng-container *ngIf="customTextControl.getError('required')">
+                Required
+              </ng-container>
+            </mat-error>
+          </mat-form-field>
+        </div>
       </mat-card-content>
     </mat-card>
   </mat-radio-group>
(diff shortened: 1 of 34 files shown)


nwkotto:
image
image

gino-m:
Thanks, Nick! @rfontanarosa PTAL?

web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.html
```diff
@@ -0,0 +1,55 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<form [formGroup]="formGroup">
+  <mat-radio-group class="data-sharing-terms"
+    [id]="typeControlKey"
+    [formControlName]="typeControlKey"
+  >
+    <mat-card *ngFor="let option of dataSharingTermsOptions"
```
amysorto, line 22: Totally optional, but we can now use the `@for` and `@if` syntax in Angular since we are on a more recent version now and it can be more readable: https://angular.dev/api/core/@for.

web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    fixture = TestBed.createComponent(DataSharingTermsComponent);
+    fixture.componentInstance.type = DataSharingType.PUBLIC;
+    fixture.componentInstance.customText = 'Hey there here is an agreement';
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('displays mat-cards for each of the three data sharing types', () => {
+    const matCards: HTMLElement[] =
```
amysorto, line 41: For the material components, it's highly recommended to use the component harnesses. Here is the documentation for card: https://material.angular.io/components/card/api#MatCardHarness as well as an example: https://material.angular.io/components/card/examples#card-harness.

I think it's reasonable to make this a separate issue to do in a follow up PR after the code freeze since this is a test.

web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.ts
```diff
@@ -0,0 +1,113 @@
...
+  @Input() type!: DataSharingType;
+  @Input() customText?: string;
+  @Output() onValidationChange: EventEmitter<boolean> =
+    new EventEmitter<boolean>();
+
+  readonly typeControlKey = 'type';
+  readonly customTextControlKey = 'customText';
+  formGroup!: FormGroup;
+
+  somethingThatIsTrue = true;
```
amysorto, line 43: seems like this is leftover from debugging

amysorto (dismissed):

web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.spec.ts
nwkotto: Yeah, good call, will do this in a follow-up to get this merged before the freeze

web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.html
nwkotto: Cool, yeah, will probably do this together with updating to use the harness, in a follow-up.

web/src/app/pages/create-survey/data-sharing-terms/data-sharing-terms.component.ts
nwkotto: Ah yeah, good catch, removed.

amysorto (approved):

Issue #1956, opened 2024-08-08
[LOI Panel] Properties not shown on LOIs with properties


gino-m:
Current:

![image](https://github.com/user-attachments/assets/830551c3-addd-4965-858e-38cdb0293f63)

Properties are also not available in the menu, potentially lost during the DB migration?

@rfontanarosa PTAL?

jcqli:
As discussed, filed https://github.com/google/ground-platform/issues/1967 to introduce the widget to the panel. Updated the description above for this bring back LOI properties in its current implementation

PR #1977, opened 2024-08-16
Convert properties proto to model

sufyanAbbasi:
Looks like there were a few JavaScript errors!

  - Need to use Object.keys()
  - Use for...of rather than for...in

Sorry for not catching these in review!
image


rfontanarosa (approved):

rfontanarosa:
this should fix #1956

PR #1994, opened 2024-08-26
Disables user role change when user is owner + User list sorted by email

### LINKED ISSUE #1982: [Create survey] Do not allow owners to change their role in ACLs
Changing role of owner to "data collector" or "viewer" makes it impossible for them to delete the survey and breaks the UI, since they can no longer add other users. 

- Show "owner" instead of survey organizer in creation flow
- Don't allow owner to change their ACL

rfontanarosa:
closes #1982 

Screenshot 2024-08-28 alle 17 01 01

### DIFF
--- a/firestore/firestore.rules
+++ b/firestore/firestore.rules
@@ -118,7 +118,8 @@
 
     // Allow users to read, create, and write their own profiles in the db.
     match /users/{userId} {
-      allow read, create, write: if userId == request.auth.uid;
+      allow read: if canAccess();
+      allow create, write: if userId == request.auth.uid;
     }
 
     // All users need to be able to read the passlist for rules to work.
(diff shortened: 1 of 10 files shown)


gino-m:
@rfontanarosa Please always place "Owner" at the top of the list.

gino-m:
Any idea what this is about in checks?

```
Error: src/message-registry.ts(139,3): error TS2352: Conversion of type '{ options: { syntax: string; }; nested: { ground: { nested: { v1beta1: { options: { java_multiple_files: boolean; java_package: string; }; nested: { AuditInfo: { fields: { userId: { type: string; id: number; }; ... 4 more ...; emailAddress: { ...; }; }; }; ... 13 more ...; Role: { ...; }; }; }; }; }; google: { ...; ...' to type 'MessageRegistryJson' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

gino-m (approved):

PR #2014, opened 2024-09-06
Update bug_report.md

jcqli:
Add app versions to bug template. Towards https://github.com/google/ground-platform/issues/2040

@gino-m PTAL

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -10,6 +10,11 @@ assignees: ''
 **Describe the bug**
 A clear and concise description of what the bug is.
 
+**App versions**
+Android:
+
+Web:
+
 **To Reproduce**
 Steps to reproduce the behavior:
 1. Go to '...'


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -10,6 +10,11 @@ assignees: ''
 **Describe the bug**
 A clear and concise description of what the bug is.
 
+**App versions**
+Android:
```
gino-m, line 14: As this for is mainly used for bugs in the web UI, perhaps we don't need Android in the template?

Also, I just realized we don't actually have build numbers / versions in the web UI, only in Android. Perhaps we should add an FR to show the version in the about screen before adding this to the template?

gino-m (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
jcqli: Filed https://github.com/google/ground-platform/issues/2015

I think it would be useful to include both when applicable, since a mismatch of versions could also be a root cause of bugs, or there could be a cross-app interaction that needs debugging. WDYT?

jcqli:
@gino-m should we close this PR or leave it in draft until https://github.com/google/ground-platform/issues/2015 is fixed?

gino-m:
> @gino-m should we close this PR or leave it in draft until #2015 is fixed?

We can leave it open, maybe add as a subtask of #2015.

gino-m:
Closing this since we don't yet have versioned releases of the web interface.

PR #2041, opened 2024-09-18
Redirects to main survey page when changes are published

### LINKED ISSUE #1975: [Create/edit survey] Clicking "Publish changes" doesn't bring user back to the main screen
...you currently need to click "Publish changes" followed by "Cancel".

rfontanarosa:
closes #1975

https://github.com/user-attachments/assets/ab6418ca-0f75-4ed1-9a6a-2efcc849e979


web/src/app/components/header/header.component.html
```diff
@@ -54,14 +54,16 @@
         <button
           mat-flat-button
           class="finish-edit-button"
-          [disabled]="!isDraftSurveyDirtyAndValid()"
+          [disabled]="publishingChanges || !isDraftSurveyDirtyAndValid()"
```
amysorto, line 57: nit: can we rename this to `isPublishingChanges` just to make it more clear that this variable represents a state

web/src/app/components/header/header.component.html
```diff
@@ -54,14 +54,16 @@
...
           (click)="onFinishEditSurveyClick()"
         >
           Publish changes
           <mat-icon *ngIf="publishingChanges">
             <mat-spinner diameter="18"></mat-spinner>
           </mat-icon>
         </button>
+
+        <div class="header-divider"></div>
```
amysorto, line 66: It looks like the header divider styles were removed in `_survey-header.component-theme.scss`. Is this supposed to still be here?

amysorto (changes requested):

web/src/app/components/header/header.component.html
rfontanarosa: good point. survey header is just a customisation of <ground-header>.

the best thing to do is move survey-header component out from main-page and merge it with ground-header. I'll file a code health issue.

amysorto (approved):

PR #2046, opened 2024-09-20
Updates data sharing agreement texts

### LINKED ISSUE #2027: [Data sharing consent] Text on data sharing consent page not unclear
The text should make it clear that the terms define what survey organizers can do with the data... that it's an agreement between the survey organizer and the data collector, _not_ what the system can do with the data. @amegantz PTAL?

rfontanarosa:
closes #2027


gino-m (approved):
👍 !

PR #2059, opened 2024-10-04
Fetches only the Lois and Submissions that the current user has access to

### LINKED ISSUE #2058: [Export CSV] Data collectors can export data collected by others

rfontanarosa:
closes #2058

### DIFF
--- a/functions/src/common/auth.ts
+++ b/functions/src/common/auth.ts
@@ -107,10 +107,3 @@ export function canImport(
   const role = getRole(user, survey);
   return !!role && [Pb.Role.SURVEY_ORGANIZER].includes(role);
 }
-
-export function canManage(
-  user: DecodedIdToken,
-  survey: DocumentSnapshot
-): boolean {
-  return canImport(user, survey);
-}
--- a/functions/src/common/auth.ts
+++ b/functions/src/common/auth.ts
@@ -107,3 +107,10 @@ export function canImport(
   const role = getRole(user, survey);
   return !!role && [Pb.Role.SURVEY_ORGANIZER].includes(role);
 }
+
+export function canManage(
+  user: DecodedIdToken,
+  survey: DocumentSnapshot
+): boolean {
+  return canImport(user, survey);
+}
--- a/functions/src/common/datastore.ts
+++ b/functions/src/common/datastore.ts
@@ -168,20 +168,23 @@ export class Datastore {
           .get()
       ).docs;
     } else {
-      const importedLois = await this.db_
+      const importedLois = this.db_
         .collection(lois(surveyId))
         .where(l.jobId, '==', jobId)
-        .where(l.source, '==', Pb.LocationOfInterest.Source.IMPORTED)
-        .get();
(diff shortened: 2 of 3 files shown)


functions/src/common/auth.ts
```diff
@@ -107,3 +107,10 @@ export function canImport(
   const role = getRole(user, survey);
   return !!role && [Pb.Role.SURVEY_ORGANIZER].includes(role);
 }
+
+export function canManage(
```
amysorto, line 111: It seems like this just calls `canImport`, maybe it makes sense to just call it directly instead?

functions/src/common/datastore.ts
```diff
@@ -132,25 +132,57 @@ export class Datastore {
...
+    jobId: string,
+    userId?: string
+  ): Promise<DocumentData[]> {
+    if (!userId) {
+      return (
+        await this.db_
+          .collection(lois(surveyId))
+          .where(l.jobId, '==', jobId)
+          .get()
+      ).docs;
```
amysorto, line 169: Is this function being called elsewhere in the application currently? I am assuming `.docs` being added at the end here might cause some issues.

amysorto (dismissed):

functions/src/common/datastore.ts
rfontanarosa: nope, it is called only there. something similar have been done by data-store.service.ts under web, but there I could use combineLatest to merge results. I didn't find a better way. what do you think @gino-m?

```
  getAccessibleLois$(
    {id: surveyId}: Survey,
    userId: string,
    canManageSurvey: boolean
  ): Observable<List<LocationOfInterest>> {
    if (canManageSurvey) {
      return this.db
        .collection(`${SURVEYS_COLLECTION_NAME}/${surveyId}/lois`)
        .valueChanges({idField: 'id'})
        .pipe(map(this.toLocationsOfInterest));
    }

    const importedLois = this.db.collection(
      `${SURVEYS_COLLECTION_NAME}/${surveyId}/lois`,
      ref => ref.where(l.source, '==', Source.IMPORTED)
    );

    const fieldDataLois = this.db.collection(
      `${SURVEYS_COLLECTION_NAME}/${surveyId}/lois`,
      ref =>
        ref
          .where(l.source, '==', Source.FIELD_DATA)
          .where(l.ownerId, '==', userId)
    );

    return combineLatest([
      importedLois.valueChanges({idField: 'id'}),
      fieldDataLois.valueChanges({idField: 'id'}),
    ]).pipe(
      map(([predefinedLois, fieldDataLois]) =>
        this.toLocationsOfInterest(predefinedLois.concat(fieldDataLois))
      )
    );
  }
```

gino-m: Lgtm

gino-m: Is the idea here that you're trying run both queries in parallel? Is the additional complexity of all the stream logic worth it? Perhaps this is an over-optimization?

gino-m (approved):

Issue #1744, opened 2024-04-17
[Create/edit survey] Allow data collectors to optionally choose between draw point or polygon tasks


gino-m:
When doing EUDR reporting, for smaller plots farmers are only required to report a point. Organizers should have the option of punting the decision of geometry type to the data collector. To make this work, we could “Drop a pin” and “Draw or walk a perimeter” into a single task, with two names depending on context:

* “Map a new site” for “Add LOI tasks” 
* “Annotate site on map” for other “advanced” geometry tasks

In both cases, the survey organizer must specify one or more “Allow data collectors to:”  options:
* “Drop a pin”
* “Draw or walk perimeter”

If the survey organizer chose >1 geometry of these options, collectors will be prompted to choose which they want to do (e.g., in modal with pin and polygon icons).

 @vittorino @amegantz for designs.

Blocked by https://github.com/google/ground-android/issues/2469

gino-m:
@vittorino I think we'd either want the dropdown or the chips (possible with checkboxes to indicate state. Can you update with the final version?

@amegantz For consistency with other field labels (eg 'Instructions") it should probably be in heading voice, eg "Allowed mapping actions" or simiar.

amegantz:
I think "Mapping actions" works for the heading. Updated the Figma

gino-m:
@amegantz I think adding "Allowed" makes it clearer that the user has a choice, not that they need to perform all the actions specified there. If it becomes too long, we can drop "Mapping" since it should be clear from context.

amegantz:
![image](https://github.com/google/ground-platform/assets/112978024/14b625a9-1387-4a43-8bf3-87661e6f079b)

gino-m:
@lecrabe Punting this to GA release since there's a workaround for smaller plots (draw a small polygon)

gino-m:
As per conversation today with @jo-spek, we should also allow users to select "Capture location" as a geometry task to use when adding new sites. Please cmiiw.

gino-m:
@rfontanarosa Please coordinate with @anandwana001 on https://github.com/google/ground-android/issues/2469.

The `DrawGeometry` proto already has `allowed_methods` which can be used for this purpose. For this FR I would leave out @jo-spek's request to allow adding a LOI using the device location for now. Users can work around this by adding a Capture Location task manually to the job.

I suggest @anandwana001 does https://github.com/google/ground-android/issues/2469 before implementing this on web.

gino-m:
[...]

PR #2061, opened 2024-10-09
Adds "Map a new site" task type

rfontanarosa:
towards #1744

Screenshot 2024-10-09 alle 16 04 03

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -250,8 +250,7 @@ GEM
     rb-fsevent (0.11.0)
     rb-inotify (0.10.1)
       ffi (~> 1.0)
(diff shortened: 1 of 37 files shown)


gino-m:
@amysorto PTAL?

gino-m:
> towards #1744
> 
> Screenshot 2024-10-09 alle 16 04 03

We spoke. Note that there can only be at most one "Map a new site" (Add LOI) task in a job. You'll also need to toggle the "free form" setting accordingly when adding/removing this task.

gino-m:
@rfontanarosa Let's hold off on this until we speak with @lecrabe about whether it's really needed. I've started a thread on the Maintainers chat about this.

PR #2064, opened 2024-10-10
Upload GeoJSON - Added CRS check

### LINKED ISSUE #1366: [Import GeoJSON] Fail with error if not in WGS84 projection
See also #718

rfontanarosa:
closes #1366

### DIFF
--- a/functions/src/import-geojson.ts
+++ b/functions/src/import-geojson.ts
@@ -247,6 +247,6 @@ function isGeometryValid(geometry: Geometry): boolean {
   return true;
 }
 
-function isPositionValid([lat, lng]: Position) {
-  return lat >= -90 && lat <= 90 && lng >= -180 && lng <= 180;
+function isPositionValid([lng, lat]: Position) {
+  return lng >= -180 && lng <= 180 && lat >= -90 && lat <= 90;
 }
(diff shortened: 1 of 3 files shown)


functions/src/import-geojson.ts
```diff
@@ -81,25 +81,64 @@ export function importGeoJsonCallback(
...
 
     file
-      .pipe(JSONStream.parse(['features', true], undefined))
-      .on('data', (geoJsonLoi: any) => {
-        try {
-          if (geoJsonType !== 'FeatureCollection') {
+      .pipe(JSONStream.parse('crs', undefined))
+      .on(
+        'data',
+        (
```
gino-m, line 105: @rfontanarosa Now might be a good time to move this handler into its own function (`onGeoJsonCrs()`?)

functions/src/import-geojson.ts
```diff
@@ -81,25 +81,64 @@ export function importGeoJsonCallback(
...
+                geoJsonCrs
+              )}'`
             );
           }
+        }
+      );
+
+    file
+      .pipe(JSONStream.parse(['features', true], undefined))
+      .on('data', (geoJsonLoi: any) => {
```
gino-m, line 130: Same here (`onGeoJsonFeature`?)

gino-m (changes requested):
LGTM! Added a small code health tax if you don't mind :)

gino-m (approved):
🍾 !

PR #2073, opened 2024-10-18
Fixes not showing task skipped bug

### LINKED ISSUE #2072: [Submission View] - Not showing skipped tasks

rfontanarosa:
closes #2072

### DIFF
--- a/web/src/app/converters/submission-data-converter.ts
+++ b/web/src/app/converters/submission-data-converter.ts
@@ -76,7 +76,8 @@ function taskDataPbToModel(pb: Pb.ITaskData[]): SubmissionData {
 
     let value = null;
 
-    if (textResponse) value = textResponse.text;
+    if (skipped) value = null;
+    else if (textResponse) value = textResponse.text;
     else if (numberResponse) value = numberResponse.number;
     else if (dateTimeResponse)
       value = new Date(timestampToInt(dateTimeResponse.dateTime));
@@ -91,10 +92,11 @@ function taskDataPbToModel(pb: Pb.ITaskData[]): SubmissionData {
         coordinatesPbToModel(captureLocationResult.coordinates!)
       );
     else if (takePhotoResult) value = takePhotoResult.photoPath;
-    else if (!skipped)
+
+    if (value === undefined)
       throw new Error('Error converting to Submission: invalid task data');
 
-    submissionData[taskId!] = new Result(value!);
+    submissionData[taskId!] = new Result(value);
   });
 
   return Map(submissionData);
--- a/web/src/app/converters/submission-data-converter.ts
+++ b/web/src/app/converters/submission-data-converter.ts
@@ -87,12 +87,9 @@ function taskDataPbToModel(pb: Pb.ITaskData[], job: Job): SubmissionData {
     else if (dateTimeResponse)
       value = new Date(timestampToInt(dateTimeResponse.dateTime));
     else if (multipleChoiceResponses) {
-      value = new MultipleSelection(
-        task.multipleChoice?.options.filter(({id: optionId}) =>
-          multipleChoiceResponses!.selectedOptionIds?.includes(optionId)
-        ) || List([]),
-        multipleChoiceResponses.otherText
-      );
+      const {selectedOptionIds, otherText} = multipleChoiceResponses;
+
+      value = new MultipleSelection(List(selectedOptionIds || []), otherText);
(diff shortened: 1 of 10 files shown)


web/src/app/pages/main-page-container/main-page/secondary-side-panel/submission-panel/submission-panel.component.html
```diff
@@ -34,59 +34,61 @@
...
+                <div *ngIf="task.type === taskType.MULTIPLE_CHOICE" class="submission-response-multiple-choice">
+                  <mat-checkbox disabled="true" checked="true" *ngFor="let option of getTaskMultipleChoiceSelections(task).values">
+                    {{option.label}}
+                  </mat-checkbox>
+    
+                  <mat-checkbox disabled="true" checked="true" *ngIf="getTaskMultipleChoiceSelections(task).otherValue as otherValue">
+                    Other: {{otherValue}}
+                  </mat-checkbox>
+                </div>
+                <!-- Photo task types -->
```
gino-m, line 58: The view visualization of various task types should be componentized, though not necessarily in the PR.

web/src/app/pages/main-page-container/main-page/secondary-side-panel/submission-panel/submission-panel.component.html
```diff
@@ -34,59 +34,61 @@
...
+                <div *ngIf="task.type === taskType.PHOTO">
+                  <img src="{{this.firebaseURLs.get(getTaskSubmissionResult(task)!.value)}}" class="submission-img"/>
```
gino-m, line 60: We're exposing a lot of logic in the view here - how the URLs are stored in the component, how we get the result for a task, how we extract the value.. can we move this into the component in a follow-up PR?

web/src/app/converters/submission-data-converter.ts
```diff
@@ -97,9 +91,10 @@ function taskDataPbToModel(pb: Pb.ITaskData[], job: Job): SubmissionData {
         coordinatesPbToModel(captureLocationResult.coordinates!)
       );
     else if (takePhotoResult) value = takePhotoResult.photoPath;
-    else throw new Error('Error converting to Submission: invalid task data');
+    else if (!skipped)
+      throw new Error('Error converting to Submission: invalid task data');
 
-    submissionData[task.id] = new Result(value!);
+    submissionData[taskId!] = new Result(value!);
```
gino-m, line 97: What is `value` when `skipped` is `true`? If it's `null` or `undefined`, won't `value!` throw an error? If that's not the case, is there a way to make this code more self-explanatory?

gino-m (changes requested):

web/src/app/pages/main-page-container/main-page/secondary-side-panel/submission-panel/submission-panel.component.html
rfontanarosa: ok

web/src/app/pages/main-page-container/main-page/secondary-side-panel/submission-panel/submission-panel.component.html
rfontanarosa: ok

gino-m (changes requested):
Since "Skipped" has special meaning you may want to consider passing around a special value rather than `null` in the future to make the code more self-documenting / explicit.

gino-m (approved):

web/src/app/converters/submission-data-converter.ts
```diff
@@ -65,27 +61,28 @@ function taskDataPbToModel(pb: Pb.ITaskData[], job: Job): SubmissionData {
       takePhotoResult,
     } = taskData;
 
+    let value = null;
```
gino-m, line 64: @rfontanarosa  Can you extra this logic to its own method in follow-up PR (eg `taskValueToResult`)? Rather than relying on a mutable variable you could just `return` the desired value or `null` depending on

gino-m, line 64: @rfontanarosa Also, can you please make the treatment of skipped explicit, and log the invalid state where `value` is non-null but `skipped` is `true`? Something like:

```ts
if (skipped) {
  if (value) {
    console.warn(`Ignoring unexpected value ${value} on skipped task`);
  }
  return new Result(null, true);
}
```

PR #2079, opened 2024-11-06
Fixed missing "proto options" syntax attribute

rfontanarosa:


gino-m (approved):
Thanks for fixing this

Issue #2108, opened 2024-11-27
Switch to custom HTML and CSS google maps markers


amysorto:
Having a custom HTML will improve text wrapping, allow for a more square shape, and custom styles for selected markers (like increasing the scale).

Requires some investigation since the example of custom styles here doesn't work: https://developers.google.com/maps/documentation/javascript/examples/advanced-markers-basic-style.

Issue #1477, opened 2024-02-07
[Submission details] Show geometry annotation reference on map


amysorto:
Currently after the submission of #1472, the number annotations don't appear on the map but only on the submissions panel.

Mock:
![image](https://github.com/google/ground-platform/assets/8575252/d82227bb-fd86-4dec-895d-f543465b3e3c)

gino-m:
@amysorto @jcqli Let's punt on this one for now?

gino-m:
@amysorto Please be sure numbers are sequential starting at "1". 🙏

kenstershiro:
@amysorto any update on this one?

amysorto:
Closing this issue since this was added to the app for points: #2109.

Another issue is filed for polygons since that is currently blocked by some other work to enable it: https://github.com/google/ground-platform/issues/2126

PR #2109, opened 2024-11-27
Use AdvancedMarkerElements and add index to markers on map

amysorto:
Part of #1477. Adds annotations for submission markers (but not polygons). @gino-m As far as I know we don't have polygons yet in our submissions. Should this PR close this issue instead and we file a separate issues for polygons?

I also filed #2108 to use custom HTML and CSS for the markers. When I initially tried the styles were not being correctly applied so I figured I can take a look into that separately so we can have better map annotiations in the meanwhile.
 
Screenshot:
Screenshot 2024-11-27 at 10 37 28 AM 1

Also I changed the index in the submission panel to start at 1 instead of 0. I thought there was a filed issue for this but I couldn't find it to link it here.

### DIFF
--- a/web/src/app/pages/main-page-container/main-page/map/map.component.ts
+++ b/web/src/app/pages/main-page-container/main-page/map/map.component.ts
@@ -234,7 +234,7 @@ export class MapComponent implements AfterViewInit, OnChanges, OnDestroy {
             task.id,
             taskResult!.value as Point,
             this.submission?.job?.color,
-            task.index.toString(),
+            task.index.toString()
           );
           this.markers.set(task.id, marker);
         } else if (geometryType === GeometryType.POLYGON) {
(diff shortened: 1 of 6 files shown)


web/src/app/pages/main-page-container/main-page/map/map.component.ts
```diff
@@ -137,52 +138,56 @@ export class MapComponent implements AfterViewInit, OnChanges, OnDestroy {
...
         this.lois$,
         this.navigationService.getLocationOfInterestId$(),
+        this.navigationService.getTaskId$(),
         this.selectedJob$,
-      ]).subscribe(([survey, lois, locationOfInterestId, selectedJob]) => {
-        const loisMap = ImmutableMap(
-          lois
-            .filter(loi =>
-              this.showPredefinedLoisOnly ? loi.predefined : true
+      ]).subscribe(
```
gino-m, line 143: This method is becoming unwieldy. While you're here any chance you could refactor it a bit, perhaps pulling out meaningful logic into helper functions?

gino-m (changes requested):
Thanks, @amysorto! Lgtm, one small code health tax if you have time. 

@rfontanarosa for  final approval.

gino-m:
@amysorto Please ping when RFAL!

amysorto:
@gino-m This PR is ready for review!

PR #2132, opened 2025-01-08
Refresh user profile on ID token change

### LINKED ISSUE #1109: Refresh db on changes to user profile
We currently only update on sign-in. We should also `/profile-refresh` when user's profile in db != latest from auth token refresh.

nwkotto:
Closes #1109 

Previously, we were only updating the user and user profile on sign-in / sign-out. This updates that logic to also make those updates on changes to the ID token.

### DIFF
--- a/web/src/app/services/auth/auth.service.spec.ts
+++ b/web/src/app/services/auth/auth.service.spec.ts
@@ -33,7 +33,13 @@ describe('AuthService', () => {
       providers: [
         {provide: AngularFirestore, useValue: {}},
         {provide: AngularFireFunctions, useValue: {}},
-        {provide: AngularFireAuth, useValue: {authState: NEVER}},
+        {
+          provide: AngularFireAuth,
+          useValue: {
+            authState: NEVER,
+            onIdTokenChanged: (callback: Function) => callback(null),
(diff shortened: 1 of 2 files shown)


nwkotto:
> LGTM. Were you able to verify this works manually?

I wasn't sure how, exactly. Do you know how to reproduce this?

gino-m:
Not sure, you'd probably need to modify the exporationbut I suppose you can try hard coding it to emit a refresh

nwkotto:
So what I was able to confirm is that an additional onAuthStateChange call is getting made on load with this code added, in localhost... which implies that the onIdTokenChanged callback is getting triggered on top of afAuth.authState emitting. But it seems like the initial user value on load for both afAuth.authState and onIdTokenChanged is null, so I haven't been able to confirm the profile refresh and how user is getting set. The function signature for onIdTokenChanged says that the callback will emit either a User or null, so no reason to believe that the user would get overwritten incorrectly.

gino-m:
Ok this sgtm, let's just get a stamp from @rfontanarosa since he's the current owner of this code.

rfontanarosa (approved):

PR #2138, opened 2025-02-04
Add "Download GeoJSON" menu option

### LINKED ISSUE #2013: [Export job data] Allow data to be exported as GeoJSON
Currently, the data export "Download job data" delivers a CSV. It would be great to also have the option to download the job data as GeoJSONs. 

Users that want to display their recorded geodata in a GIS need to do extra steps to draw the WKT notation from the CSV, whereas a GeoJSON can simply be dragged and dropped into any GIS. Also, regarding the EUDR aspect, the EU Information System is set to accept only GeoJSONs containing the geometries, the ISO2 country code and a few optional columns.

Thanks for considering.

rfontanarosa:
closes #2013

https://github.com/user-attachments/assets/d218d150-d163-4f89-8314-b7c3ce10e5b8

Screenshot 2025-02-04 alle 17 15 04


functions/src/export-geojson.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+    [l.submissionCount]: 0,
+    [l.source]: Pb.LocationOfInterest.Source.IMPORTED,
+    [l.properties]: {
+      name: {[pr.stringValue]: 'Dinagat Islands'},
+      area: {[pr.numericValue]: 3.08},
+    },
+  };
+  const testCases = [
+    {
+      desc: '[GEOJSON] - export point',
```
nwkotto, line 134: Out of curiosity, what is the "[GEOJSON]" supposed to indicate? That this is exporting to geojson specifically, and that there are going to be other formats? I would probably put these cases in a "describe" block instead, that better captures the category of things being tested.

functions/src/export-geojson.ts
```diff
@@ -0,0 +1,144 @@
...
+import * as HttpStatus from 'http-status-codes';
+import {DecodedIdToken} from 'firebase-admin/auth';
+import {toMessage} from '@ground/lib';
+import {GroundProtos} from '@ground/proto';
+import {toGeoJsonGeometry} from '@ground/lib';
+
+import Pb = GroundProtos.ground.v1beta1;
+
+/**
+ * Iterates over all LOIsin a job returning a valid GeoJSON file.
```
nwkotto, line 31: *LOIs in

functions/src/export-geojson.ts
```diff
@@ -0,0 +1,144 @@
...
+    res.status(HttpStatus.NOT_FOUND).send('Survey not found');
+    return;
+  }
+  if (!canExport(user, surveyDoc)) {
+    res.status(HttpStatus.FORBIDDEN).send('Permission denied');
+    return;
+  }
+  const ownerId = canImport(user, surveyDoc) ? undefined : userId;
+
+  console.log(
```
nwkotto, line 53: Did you mean to keep this in here? Remove?

functions/src/export-geojson.ts
```diff
@@ -0,0 +1,144 @@
...
+  const surveyDoc = await db.fetchSurvey(surveyId);
+  if (!surveyDoc.exists) {
+    res.status(HttpStatus.NOT_FOUND).send('Survey not found');
+    return;
+  }
+  if (!canExport(user, surveyDoc)) {
+    res.status(HttpStatus.FORBIDDEN).send('Permission denied');
+    return;
+  }
+  const ownerId = canImport(user, surveyDoc) ? undefined : userId;
```
nwkotto, line 51: Should this logic be reversed? If canImport is true, the ownerId is the userId? Or is this the correct logic? Maybe a comment would help clarify, if this is right.

functions/src/export-geojson.ts
```diff
@@ -0,0 +1,144 @@
...
+    return;
+  }
+  const {name: jobName} = job;
+
+  res.type('application/json');
+  res.setHeader(
+    'Content-Disposition',
+    'attachment; filename=' + getFileName(jobName)
+  );
+  const jsonStream = JSONStream.stringify(
```
nwkotto, line 78: Out of curiosity, why does this need to be defined as a string, with all the newlines? Make it a bit difficult to read. Maybe you could add a comment explaining this?

functions/src/export-geojson.ts
```diff
@@ -0,0 +1,144 @@
...
+  );
+  jsonStream.pipe(res);
+
+  const rows = await db.fetchLocationsOfInterest(surveyId, jobId);
+
+  rows.forEach(row => {
+    try {
+      const loi = toMessage(row.data(), Pb.LocationOfInterest);
+      if (loi instanceof Error) throw loi;
+      if (!isAccessibleLoi(loi, ownerId)) return;
```
nwkotto, line 92: Should these also be thrown, so they can be logged as skipped, for consistency?

functions/src/export-geojson.ts
```diff
@@ -0,0 +1,144 @@
...
+import {toMessage} from '@ground/lib';
+import {GroundProtos} from '@ground/proto';
+import {toGeoJsonGeometry} from '@ground/lib';
+
+import Pb = GroundProtos.ground.v1beta1;
+
+/**
+ * Iterates over all LOIsin a job returning a valid GeoJSON file.
+ */
+export async function exportGeojsonHandler(
```
nwkotto, line 33: This is quite a long function. This is a nit, but any way this could be split up into useful chunks of logic? e.g. error checking, writing rows, etc.

functions/src/export-geojson.spec.ts
nwkotto, line 1: There are a fair few things from the function that are not tested here, especially error cases. I understand not testing everything, but anything that's especially important that you think we should test?

functions/src/export-geojson.ts
rfontanarosa: /**
   * Creates a JSON stream representing a GeoJSON FeatureCollection.
   * The arguments follow a pattern of open, separator, close, and
   * indent, which helps to organize the JSON structure for clarity.
   */

functions/src/export-geojson.spec.ts
rfontanarosa: After investigating the export-csv test logic, I found that the jsonstream library uses a queue for writing to the stream and I didn't find a way to capture this.
Maybe we can change the mocking strategy, so instead of mocking the entire response, I'll likely get better results by focusing my tests on the transformation of he LOI data into GeoJSON features.

rfontanarosa: wdyt @nwkotto ?

nwkotto: Yeah, that makes sense, it would be nice to check that the transformation from LOI to GeoJSON is working correctly, instead of the focus on the overall request. But I don't want to hold this up, either, if you've tested this manually -- could split out that functionality and test it in a follow-up PR.

nwkotto (approved):

PR #2146, opened 2025-03-03
Instruction task proto

### LINKED ISSUE #1991: Add Task instructions feature
It would be great to have a feature where a text block can be shown before, in between or after a task when collecting data with the Ground Android app.

This was identified as a useful addition for SERVIRs Ghana campaign to provide instructions to data collectors or to show/ read out statements to farmers about what is being done on their fields.

To jump to the next task in GA, one has to press an 'OK' button. Issue will be raised in GA as well.

This relates to the Ground Android feature request https://github.com/google/ground-android/issues/2690

Textblock_desktop

anandwana001:
Fixes #1991

@gino-m  PTAL?


gino-m (approved):
@anandwana001 It looks like you just copied one of the other task types and didn't add a reference to the actual job proto. Added those here, PTAL before merging.

Issue #2149, opened 2025-03-06
[Sidepanel LOI list] Improve LOI list style


rfontanarosa:
- [x] Align tag icons on the left with the burger menu icon on the top bar
- [x] Overflow menu icon on each job should appear only on mouse hover
- [x] Right align submissions number with the overflow menu of the job, or hide it if zero
- [x] Fit the expand arrow in the margin between the icon and the page border (see Gmail as example), and hide it if there is no content to show in the job
- [x] Remove the lines dividing each job

![Image](https://github.com/user-attachments/assets/94b5e840-297a-4780-97bd-a6e252fed9f0)

@vittorino

PR #2150, opened 2025-03-06
Hide Job/Loi Menu when no Lois are available + Various fixes

### LINKED ISSUE #1613: [Submissions list] Don't show dropdown arrow if there are no LOIs
**Describe the bug**
If there are no LOIs for a job, we should not show the dropdown arrow
image

rfontanarosa:
closes #1613
part of #2149

https://github.com/user-attachments/assets/8ebbbcb8-f1e7-45ef-bdca-5c0c360542c0

### DIFF
--- a/web/src/app/components/job-list-item/job-list-item.component.ts
+++ b/web/src/app/components/job-list-item/job-list-item.component.ts
@@ -49,7 +49,8 @@ export class JobListItemComponent implements OnInit, OnDestroy {
 
   getLevel = (node: DynamicFlatNode) => node.level;
   isExpandable = (node: DynamicFlatNode) => node.expandable;
-  hasChild = (_: number, _nodeData: DynamicFlatNode) => _nodeData.expandable;
+  hasChild = (node: DynamicFlatNode) => node.childCount > 0;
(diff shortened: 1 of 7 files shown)


web/src/app/components/job-list-item/job-list-item.component.ts
```diff
@@ -82,18 +82,14 @@ export class JobListItemComponent implements OnInit, OnDestroy {
...
-        /* expandable= */ true,
-        /* iconName= */ 'label',
-        /* iconColo= */ this.job!.color!,
-        /* jobId= */ this.job!.id
-      ),
-    ]);
+    this.subscription.add(
+      this.loiService.getLocationsOfInterest$().subscribe(lois => {
+        console.log(this.job!.name, lois.size);
+        this.dataSource.data = this.dataSource.data.concat([
```
gino-m, line 88: Is there a way to do this without relying on side effects? Can you instead use `*ngIf=` and look at the LOI count, for example?

gino-m (changes requested):

gino-m (approved):

Issue #2121, opened 2024-12-10
[Auth & ACLs] Introduce "unlisted public" surveys


gino-m:
These surveys would be accessible to all pass-listed users provided they know the URL. They would not show up in survey list unless the user has previously access the survey. 

This would unblock #1793 and #2020.

Qs:

 * How do they get back into an unlisted survey if they no longer have the QR code or URL? Introduce @vittorino @lecrabe should we introduce a "recent surveys" section? 
* Where is this option set and how? @vittorino

gino-m:
In https://github.com/google/ground-platform/issues/2117#issuecomment-2556986607 @jo-spek said:

> TBH I don't know what exactly you mean with these distinctions (public listed / unlisted / survey templates). This is what I understand:

My understanding is that there at least three related CUJs we want to address here (@kenstershiro @lecrabe cmiiw):

1. [P0] When organizing trainings or campaigns, adding data collectors by email address is error prone and time consuming. Survey organizers want an easier way to provide access to an entire group of users.
2. [P1?] New users want to try out Ground for Android quickly, even if they haven't been invited to a survey yet.
3. [P1?] Survey organizers often surveys with similar data collection tasks, so they want to have a way to quickly add those common tasks.


Solutions:

* *Anyone with a link or QR Code*: Allowing organizers to share a survey by QR code or URL would solve CUJ 1 - survey organizers would share the URL or QR code with data collectors, who would discover the survey using that artifact. The survey is not listed in the list of publicly available surveys, so only those with the link or QR code would be able to access the survey.
* *Public*: Allowing organizer to set a survey as "public" would allow everyone to contribute data to the survey as data collectors. This addresses CUJ 2 above - new users will automatically see these in their survey list after they download the app. 
* *Task templates*: Allow survey organizers to save one or more tasks as a template. With the above solution the priority of this solution was bumped down, since it introduces additional complexity to the survey editor, and will require some careful thought from product and UX.

Does that make sense?

gino-m:
@lecrabe @kenstershiro isn't this blocking trainings? If so may want to bump to P0?

lecrabe:
Yes, let's bump to P0, not sure where to do that...

kenstershiro:
Upped to P0, was already top of the list for @rfontanarosa and @vittorino - please shout if there are any questions on the detailed requirements. @jo-spek FYI

gino-m:
Thanks @vittorino for the designs. @rfontanarosa FYI

[...]

PR #2151, opened 2025-03-07
Added Survey Visibilty (proto, model, converter)

rfontanarosa:
part of #2121

### DIFF
--- a/proto/src/ground/v1beta1/survey.proto
+++ b/proto/src/ground/v1beta1/survey.proto
@@ -85,6 +85,24 @@ message Survey {
 
   // Defines the survey's current state.
   State state = 7;
+
+  // Defines the survey's current visibility.
+  enum Visibility {
+    // Default value when visibility is missing.
+    VISIBILITY_UNSPECIFIED = 0;
+
+    // Anyone with a link or QR Code can contribute.
+    RESTRICTED = 1;
+
+
+    UNLISTED = 2;
+
+    // Anyone can contribute.
+    PUBLIC = 3;
+  }
+
+  // Defines the survey's current visibility.
+  Visibility visibility = 8;
 }
 
 // Defines a user's role in a survey.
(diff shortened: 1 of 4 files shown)


proto/src/ground/v1beta1/survey.proto
```diff
@@ -85,6 +85,24 @@ message Survey {
 
   // Defines the survey's current state.
   State state = 7;
+
+  // Defines the survey's current visibility.
+  enum Visibility {
```
gino-m, line 90: Not sure "visibility" is the right name here - it's really who can collect data beyond those in ACLs. Let's just mirror Doc & Drive: `GeneralAccess` and `general_access`?

proto/src/ground/v1beta1/survey.proto
```diff
@@ -85,6 +85,24 @@ message Survey {
...
+    // Default value when visibility is missing.
+    VISIBILITY_UNSPECIFIED = 0;
+
+    // Only users in shared list can contribute.
```
gino-m, line 94: `// Only users with the `DATA_COLLECTOR role in the survey ACLs can contribute.`

gino-m (changes requested):
Great! A few minor edits, otherwise LG.

proto/src/ground/v1beta1/survey.proto
```diff
@@ -85,6 +85,24 @@ message Survey {
 
   // Defines the survey's current state.
   State state = 7;
+
+  // Defines the survey's current general_access.
```
gino-m, line 89: The enum defines possible values, not the current survey's general access.

Try:
```proto
// Possible values for who can access the survey beyond users in the ACLs.
```

proto/src/ground/v1beta1/survey.proto
```diff
@@ -85,6 +85,24 @@ message Survey {
...
+    RESTRICTED = 1;
+
+    // Anyone with a link or QR Code can contribute.
+    UNLISTED = 2;
+
+    // Anyone can contribute.
+    PUBLIC = 3;
+  }
+
+  // Defines the survey's current general_access.
```
gino-m, line 104: This description doesn't provide any additional value.

```proto
// Defines who can access the survey beyond users in the ACLs.
```

gino-m (changes requested):

PR #2158, opened 2025-03-19
Adds "Invalid Survey" dialog instead of disabling "Publish changes" button

### LINKED ISSUE #1634: [Feature] Avoid disabling the "Publish Changes" button if there are errors
It's generally discouraged to disable a form save button, i.e. "Publish Change", when there are issues instead of letting the user click the button and get an error message. That way, the user doesn't have to look for the missing required fields, and furthermore, the button click can trigger any error validation as well.

rfontanarosa:
closes #1634

https://github.com/user-attachments/assets/d006d7fb-19d2-483a-a2dc-ab95f6b5b221

### DIFF
--- a/web/src/app/components/header/header.component.html
+++ b/web/src/app/components/header/header.component.html
@@ -54,7 +54,7 @@
         <button
           mat-flat-button
           class="finish-edit-button"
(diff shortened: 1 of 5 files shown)


gino-m (changes requested):
Please add/update tests accordingly

gino-m (approved):

PR #2165, opened 2025-03-25
Refactor survey creation phase labels and logic

rfontanarosa:
- SetupPhase.REVIEW into SetupPhase.SHARE_SURVEY
- Moved progress bar label, title and descriptions into one variable **setupPhaseMetadata**
- Reduced code length using <ground-step-card>

### DIFF
--- a/web/src/app/pages/create-survey/create-survey.component.ts
+++ b/web/src/app/pages/create-survey/create-survey.component.ts
@@ -91,7 +91,7 @@ const createSurveyPhaseMetadata = new Map<
   [
     CreateSurveyPhase.DEFINE_DATA_SHARING_TERMS,
     {
-      progressBarTitle: 'Review abd share survey',
+      progressBarTitle: 'Define data sharing terms',
       cardTitle: 'Data sharing consent',
       cardDescription:
         'Require data collectors to agree to data sharing terms before submitting data',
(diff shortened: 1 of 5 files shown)


web/src/app/pages/create-survey/create-survey.component.ts
```diff
@@ -35,6 +35,79 @@ import {TaskService} from 'app/services/task/task.service';
 
 import {SurveyLoiComponent} from './survey-loi/survey-loi.component';
 
+export enum SetupPhase {
```
gino-m, line 38: Nit: The component is called "Create" survey, but this enum is called "Setup", which might be taken to mean these are referring to two different things. Can we use one or the other consistently? (eg. `CreateSurveyPhase`)?

gino-m (changes requested):
Nice work! One nit.

web/src/app/pages/create-survey/create-survey.component.ts
rfontanarosa: done

rfontanarosa:
@gino-m  @vittorino  moved back to draft. waiting for a feedback: https://docs.google.com/spreadsheets/d/1wjG2Ce7ExpcIXNGdqIs_VtcSnDzmn1050YpAfN-qpVs/edit?usp=sharing

gino-m (approved):
Can we merge this and update the labels in a follow-up?

PR #2187, opened 2025-04-15
Fixes a bug in Export CSV with unset LOI IDs

### LINKED ISSUE #2186: [Export CSV] - Empty file and Limited results
A user found a bug with Export CSV functionality.

It returns an empty file even if the Location of interest contains submissions and this is visible in the web application.

Looking at the database, I've found that those LOIs don't have any loiId.

Moreover, in some cases the results are limited to the pagination size (50 by default).

rfontanarosa:
closes #2186

The optimal pagination solution using startAfter currently has issues. The startAfter function doesn't work as expected, even with the same orderBy and correct indexes. It appears the best approach is to avoid explicitly specifying fields in startAfter and instead use the last document from the previous page to determine the starting point for the next.

This change, however, breaks the existing tests because mock-cloud-firestore, the mocking library, doesn't support Firestore indexes, which are needed for the current pagination strategy.


functions/src/common/datastore.ts
```diff
@@ -179,32 +179,29 @@ export class Datastore {
...
     jobId: string,
     ownerId: string | undefined,
-    page: number
+    pageSize: number
   ) {
     const loisQuery = this.db_
       .collection(lois(surveyId))
       .where(l.jobId, '==', jobId)
-      .orderBy(l.id);
+      .orderBy(FieldPath.documentId());
```
gino-m, line 194: Nice find!

PR #2189, opened 2025-04-16
Fixes survey deletion

### LINKED ISSUE #2188: [Delete Survey] - Survey document is not completely deleted
I notice that in the database there are some empty surveys with jobs and submissions subcollections.

It seems that the problem is the async nature of the operations.

rfontanarosa:
closes #2188

Due to the asynchronous nature and different data handling mechanisms between Angular Firestore and pure Firestore, survey deletion didn't always work well.

Using a synchronous for loop and moving everything under a transaction, this fix avoids side effects.


web/src/app/services/data-store/data-store.service.ts
```diff
@@ -249,27 +252,29 @@ export class DataStoreService {
       .set(jobToDocument(job));
   }
 
-  async deleteSurvey(survey: Survey) {
+  async deleteSurvey(survey: Survey): Promise<void> {
```
gino-m, line 255: Can we do this in a single atomic transaction? Otherwise if the process gets interrupted the db will end up in an invalid state.

https://firebase.google.com/docs/firestore/manage-data/transactions

gino-m (changes requested):

web/src/app/services/data-store/data-store.service.ts
rfontanarosa: done

gino-m (changes requested):
A few more suggestions. Please also add what was broken and how it's been fixed to the PR description.

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -249,39 +254,71 @@ export class DataStoreService {
...
+
+        const surveyRef = this.db.firestore
+          .collection(SURVEYS_COLLECTION_NAME)
+          .doc(surveyId);
+
+        transaction.delete(surveyRef);
+      });
+    } catch (error) {
+      console.error(`Error deleting survey: ${surveyId}`, error);
+      throw error;
```
gino-m, line 314: If we log and rethrow we'll end up with duplicate log messages which will add noise and make debugging harder. Why not just let the error throw and let the exception will be logged upstream?

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -290,19 +327,10 @@ export class DataStoreService {
...
-      ref => ref.where(l.jobId, '==', jobId)
-    );
-    const querySnapshot = await firstValueFrom(lois.get());
-    return await Promise.all(querySnapshot.docs.map(doc => doc.ref.delete()));
-  }
-
-  async deleteLocationOfInterest(surveyId: string, loiId: string) {
+    loiId: string
+  ): Promise<void> {
     await this.deleteAllSubmissionsInLocationOfInterest(surveyId, loiId);
```
gino-m, line 334: This longer name doesn't help readability imo.. could just be `deleteSubmissionsByLoiId` or `deleteSubmissionsForLoi`

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -249,39 +254,71 @@ export class DataStoreService {
...
+      console.error(`Error deleting survey: ${surveyId}`, error);
+      throw error;
+    }
   }
 
   private async deleteAllSubmissionsInLocationOfInterest(
     surveyId: string,
     loiId: string
-  ) {
+  ): Promise<void[]> {
```
gino-m, line 321: What does `void[]` mean? Shouldn't it just be `Promise<void>`?

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -290,19 +327,10 @@ export class DataStoreService {
     return await Promise.all(querySnapshot.docs.map(doc => doc.ref.delete()));
```
gino-m, line 327: We don't need `return` here.

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -188,7 +187,13 @@ export class DataStoreService {
   ): Promise<void> {
     const {id: surveyId, jobs} = survey;
 
-    jobIdsToDelete?.forEach(jobId => this.deleteJob(surveyId, jobId));
+    if (jobIdsToDelete) {
+      await this.db.firestore.runTransaction(async transaction => {
+        for (const jobId of jobIdsToDelete) {
```
gino-m, line 192: The entire operation `updateSurvey` should be done in the same transaction to ensure atomicity and isolation, not just deleting jobs.

gino-m (changes requested):

web/src/app/services/survey/survey.service.ts
```diff
@@ -158,10 +158,15 @@ export class SurveyService {
...
-   * @param surveyId the id of the survey.
+   * @param survey the survey instance.
    */
-  deleteSurvey(survey: Survey): Promise<void> {
-    return this.dataStore.deleteSurvey(survey);
+  async deleteSurvey(survey: Survey): Promise<void> {
+    try {
+      await this.dataStore.deleteSurvey(survey);
+    } catch (error) {
+      console.error(`Error deleting survey: ${survey.id}`, error);
```
gino-m, line 167: Either log or throw, but not both. How do other methods in this service return errors?

gino-m (changes requested):
LG, one nit.. 

General observation: data-store.service.ts seems to be getting a big unwieldy, a refactor may be needed soon.

gino-m (approved):

PR #2191, opened 2025-04-23
Fixes not showing public surveys created by others

rfontanarosa:
part of #2121

### DIFF
--- a/web/src/app/services/data-store/data-store.service.ts
+++ b/web/src/app/services/data-store/data-store.service.ts
@@ -164,11 +164,7 @@ export class DataStoreService {
         map(surveyChangeAction),
         map(surveys =>
           surveys.filter(
-            survey =>
-              ![
-                SurveyGeneralAccess.PUBLIC,
-                SurveyGeneralAccess.UNLISTED,
-              ].includes(survey.generalAccess!)
+            survey => survey.generalAccess === SurveyGeneralAccess.RESTRICTED
           )
         )
       );
@@ -241,6 +237,10 @@ export class DataStoreService {
       .collection(SURVEYS_COLLECTION_NAME)
       .doc(surveyId)
       .set(surveyToDocument(surveyId, {title: name}), {merge: true});
+  }
+
+  getDb() {
+    return this.db;
   }
 
   /**
--- a/web/src/app/services/data-store/data-store.service.ts
+++ b/web/src/app/services/data-store/data-store.service.ts
@@ -133,24 +133,30 @@ export class DataStoreService {
   }
 
   /**
+   * Transforms an array of Firestore `DocumentChangeAction` into an array of `Survey` models.
+   *
+   * It extracts the document data and ID from each action, converts it to a `Survey` model using `surveyDocToModel`,
+   * filters out any invalid surveys using `DataStoreService.filterAndLogError`, and finally casts the remaining
+   * elements to the `Survey` type.
+   */
+  private documentChangeToSurvey = (surveys: DocumentChangeAction<unknown>[]) =>
+    surveys
+      .map((action: DocumentChangeAction<unknown>) => {
+        const docData = action.payload.doc.data() as DocumentData;
+        const id = action.payload.doc.id;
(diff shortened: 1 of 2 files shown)


web/src/app/services/data-store/data-store.service.ts
```diff
@@ -140,7 +141,17 @@ export class DataStoreService {
     userEmail: string,
     userId: string
   ): Observable<List<Survey>> {
-    return this.db
+    const surveyChangeAction = (surveys: DocumentChangeAction<unknown>[]) =>
```
gino-m, line 144: Let's extract this into its own function with clear TSDoc and name as to what this does (eg `documentChangeToSurvey()`)

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -150,26 +161,43 @@ export class DataStoreService {
...
-                    (survey.generalAccess === SurveyGeneralAccess.UNLISTED &&
-                      survey.ownerId === userId))
-              )
-              .map(survey => survey as Survey)
+          surveys.filter(
+            survey =>
+              ![
+                SurveyGeneralAccess.PUBLIC,
+                SurveyGeneralAccess.UNLISTED,
+              ].includes(survey.generalAccess!)
```
gino-m, line 171: Filtering must be done on the query itself, not on the result set, since Firestore rule do not allow you to query an item which you do not have access to.

To make this work, you'll need to have two streams, one with changes to public and unlisted, and one with restricted surveys to which the user has been granted access. You'll then need to merge the two.

gino-m (changes requested):
Some design guidance.

web/src/app/services/data-store/data-store.service.ts
rfontanarosa: firebase is limited. you can't perform a not-in after an in. you can't use two not-in either. in the restricted case, we have to check if the user is in the acl and if the survey is restricted and if the generalAccess field is not set (for compatibility). there is no way to do this. the only way is to do this in the result set, when the generalAccess field has a value.

regarding the other two... we need two different streams because the users can see the UNLISTED survey only if they are the owner.

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -150,26 +167,39 @@ export class DataStoreService {
...
-              .filter(DataStoreService.filterAndLogError<Survey>)
-              .filter(
-                survey =>
-                  survey instanceof Survey &&
-                  (survey.generalAccess !== SurveyGeneralAccess.UNLISTED ||
-                    (survey.generalAccess === SurveyGeneralAccess.UNLISTED &&
-                      survey.ownerId === userId))
-              )
-              .map(survey => survey as Survey)
+          surveys.filter(
```
gino-m, line 172: Why not filter this in the Firebase query instead of after the fact in the client?

gino-m (changes requested):

web/src/app/services/data-store/data-store.service.ts
```diff
@@ -243,6 +245,10 @@ export class DataStoreService {
       .set(surveyToDocument(surveyId, {title: name}), {merge: true});
   }
 
+  getDb() {
```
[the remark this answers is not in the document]
rfontanarosa: sorry, I used it for testing proposes...

web/src/app/services/data-store/data-store.service.ts
rfontanarosa: because at the moment we have surveys with generalAccess = RESTRICTED and surveys without the generalAccess field. the natural and most efficient way is to add .where('generalAccess', '==', SurveyGeneralAccess.RESTRICTED) but it breaks compatibility with the old surveys.

we can improve this query once we have aligned all the old surveys setting them as RESTRICTED. I can open a new issue about it, because we can do this after the PUBLIC SURVEY feature is in production. wdyt? @gino-m

gino-m (approved):

PR #2196, opened 2025-04-29
Adds Copy Link and Copy QR code buttons

### LINKED ISSUE #2121: [Auth & ACLs] Introduce "unlisted public" surveys

rfontanarosa:
closes #2121

### DIFF
--- a/web/src/app/components/share-buttons/share-buttons.component.ts
+++ b/web/src/app/components/share-buttons/share-buttons.component.ts
@@ -1,0 +1,28 @@
+/**
+ * Copyright 2025 The Ground Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Component, Input} from '@angular/core';
+
+@Component({
+  selector: 'ground-share-buttons',
+  templateUrl: './share-buttons.component.html',
+  styleUrls: ['./share-buttons.component.scss'],
+})
+export class ShareButtonsComponent {
+  @Input() surveyId = '';
+
+  constructor() {}
+}
--- a/web/src/app/pages/edit-survey/edit-survey.component.html
+++ b/web/src/app/pages/edit-survey/edit-survey.component.html
@@ -130,7 +130,10 @@
         <h1 class="edit-survey-title" *ngIf="sectionTitle !== ''">
           {{ sectionTitle }}
         </h1>
+
         <router-outlet></router-outlet>
+
+        <ground-share-buttons [surveyId]="surveyId"></ground-share-buttons>
       </div>
     </div>
   </div>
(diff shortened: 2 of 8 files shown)


web/src/app/components/share-buttons/share-buttons.component.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import {NavigationService} from 'app/services/navigation/navigation.service';
+import {NotificationService} from 'app/services/notification/notification.service';
+
+@Component({
+  selector: 'ground-share-buttons',
+  templateUrl: './share-buttons.component.html',
+  styleUrls: ['./share-buttons.component.scss'],
+})
+export class ShareButtonsComponent implements OnInit {
```
gino-m, line 27: Add TSDoc explaining what part of the UI this component implements.

gino-m, line 27: `CopySurveyControlsComponent` might be more appropriate here.

web/src/app/components/share-buttons/share-buttons.component.ts
```diff
@@ -0,0 +1,88 @@
...
+      if (!blob) {
+        console.error('Failed to create blob from canvas');
+        return;
+      }
+
+      const data = [new ClipboardItem({[blob.type]: blob})];
+      navigator.clipboard.write(data).then(
+        () => {
+          this.notificationService.success(
+            'Survey QR code copied into the clipboard'
```
gino-m, line 77: Nit: s/into the clipboard/to clipboard/

web/src/app/components/share-buttons/share-buttons.component.ts
```diff
@@ -0,0 +1,88 @@
...
+          );
+        },
+        _ => {
+          this.notificationService.error(
+            'Impossible to copy Survey QR code into the clipboard'
```
gino-m, line 82: "Survey QR code could not be copied to clipboard"

Why would this happen? Is there anything actionable we can tell the user? (Enable permissions, etc?)

web/src/app/pages/edit-survey/edit-survey.component.html
```diff
@@ -130,7 +130,12 @@ <h3 matSubheader class="job-label">Jobs</h3>
         <h1 class="edit-survey-title" *ngIf="sectionTitle !== ''">
           {{ sectionTitle }}
         </h1>
+
         <router-outlet></router-outlet>
+
+        <div class="button-section" *ngIf="router.url.endsWith('/share')">
```
gino-m, line 136: The individual components shouldn't need to know about the format of the URL, that's the job of the NavigationService. Let's ask that service instead?

web/src/app/services/navigation/navigation.service.ts
```diff
@@ -290,6 +290,10 @@ export class NavigationService {
     );
   }
 
+  getSurveyAppLink(surveyId: string): string {
+    return `${window.location.origin}/android/${SURVEY_SEGMENT}/${surveyId}`;
```
gino-m, line 294: Should we be using `APP_BASE_HREF` instead of directly reading `window`?

gino-m (changes requested):

web/src/app/components/share-buttons/share-buttons.component.ts
rfontanarosa: it is hard to reproduce because the error is raised only if the write is executed in an unsecure context: [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).

gino-m (approved):

Issue #2200, opened 2025-05-02
[Feature] AppLink adoption for android links


rfontanarosa:
https://developer.android.com/training/app-links

- [x] Add a strategy to deploy the **.well-known** folder and **assertLinks.json**
- [x] After the strategy is approved, update the documentation

kenstershiro:
@rfontanarosa could you add a couple of lines to the description of this so that I can assign a priority? Is this needed for QR links to open on the survey?

kenstershiro:
Done, just needs testing by @rfontanarosa and @anandwana001 : using the survey link should open the Android app on the specific survey.

PR #2201, opened 2025-05-02
Deployment of .well-known/assertLinks.json using the proper environment

rfontanarosa:
part of #2200

Checks if assertLinks.json is available in the keys/[ENV] folder and moves it to the .well-known folder within the dist directory.

An example of assertLinks.json:

```
[{
  "relation": ["delegate_permission/common.handle_all_urls"],
  "target": {
    "namespace": "android_app",
    "package_name": "",
    "sha256_cert_fingerprints":
    [""]
  }
}]

```


web/scripts/copy-well-known.js
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
+
+const project = process.env['npm_config_project'];
```
gino-m, line 19: Please add a JSDoc comment explaining what this file does and why it's needed.

gino-m: "well-known" should also be qualified (well-known what?), since without context it has several interpretations.

gino-m (changes requested):
The PR description says "deployment", but the PR only updates the build step. Is this a mistake?

web/scripts/copy-well-known.js
rfontanarosa: ok, I renamed it into asset-links. more readable and correct while .well-known is the folder where firebase automatically looks for it by default.

gino-m (approved):

gino-m:
Looks like this breaks CD, please check.

gino-m (approved):

PR #2203, opened 2025-05-07
Fixes missing 0 values in properties panel

### LINKED ISSUE #2202: [LOI Properties Dialog] Missing property when value is 0
There are some missing values into LOI properties. It seems that it filters all the 0 values.

Image

rfontanarosa:
closes #2202

Screenshot 2025-05-07 alle 11 32 58

### DIFF
--- a/lib/src/firestore-to-proto.ts
+++ b/lib/src/firestore-to-proto.ts
@@ -55,7 +55,7 @@ function toMessageInternal<T>(
       console.debug(messageValue);
       continue;
     }
-    if (messageValue !== false) properties[fieldName] = messageValue;
+    if (messageValue !== null) properties[fieldName] = messageValue;
   }
   return createInstance<T>(constructor, properties);
 }
--- a/lib/src/firestore-to-proto.ts
+++ b/lib/src/firestore-to-proto.ts
@@ -55,7 +55,7 @@ function toMessageInternal<T>(
       console.debug(messageValue);
       continue;
     }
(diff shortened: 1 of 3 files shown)


lib/src/proto-to-firestore.spec.ts
```diff
@@ -41,7 +41,9 @@ describe('toDocumentData()', () => {
       expected: {
         [s.name]: 'Survey name',
         [s.description]: 'Survey desc',
-        [s.state]: Survey.State.STATE_UNSPECIFIED
+        [s.state]: Survey.State.STATE_UNSPECIFIED,
```
gino-m, line 44: Did you mean for these changes to be in this PR?

lib/src/firestore-to-proto.ts
```diff
@@ -55,7 +55,7 @@ function toMessageInternal<T>(
       console.debug(messageValue);
       continue;
     }
-    if (messageValue) properties[fieldName] = messageValue;
+    if (messageValue !== false) properties[fieldName] = messageValue;
```
gino-m, line 58: Please add a test to prevent regresssions.

gino-m: What about boolean `false` values? Shouldn't we instead be checking for `messageValue !== null`?

gino-m (changes requested):

gino-m (approved):

Issue #1418, opened 2024-01-10
[ACLs] Allow survey organizers to control visibility of job data


gino-m:
My interpretation of feature requested by @lecrabe:

* Survey organizers should be able to specify whether data collectors can see each others' data (LOIs, submissions).
* Survey organizers will always see all data.

We should be sure to enforce this through Firestore rules.

@lecrabe Please cmiiw. @vittorino for UXD, @amegantz for UXW.

jcqli:
@gino-m Is this part of the ACL work for 3/5 field testing or is this different? If not part of field testing we should punt from this milestone

gino-m:
> @gino-m Is this part of the ACL work for 3/5 field testing or is this different? If not part of field testing we should punt from this milestone

This is different. Not needed for field testing, but def. needed for IDF announcement.

jcqli:
Tasks:
- add toggle in web interface on sharing screen
- add Firebase rules to enforce visibility
- Apply toggle on web & android
May need additional issues to track the multiple tasks (I'll look into this)

vittorino:
Creation flow toggle
![image](https://github.com/google/ground-platform/assets/679199/9b62372a-c45e-40ae-86e3-d9fb06dec61b)

Survey details toggle
![image](https://github.com/google/ground-platform/assets/679199/838cd64f-df3a-42a8-a7a0-278fcc9d4d10)

jcqli:
@rfontanarosa Let's deprioritize this for after March 13th, unassigning

gino-m:
Android issue: https://github.com/google/ground-android/issues/3068

gino-m:
Thanks @vittorino for the design:

![Image](https://github.com/user-attachments/assets/cce21947-24ac-4ed5-a6a2-d4ba112133e5)

PR #2204, opened 2025-05-07
Add Survey Data Visibility to model, converter, and test

rfontanarosa:
part of #1418

### DIFF
--- a/web/src/app/converters/survey-data-converter.ts
+++ b/web/src/app/converters/survey-data-converter.ts
@@ -54,7 +54,7 @@ export const MODEL_ROLES = Map([
   [Pb.Role.VIEWER, Role.VIEWER],
 ]);
 
-const MODEL_DATA_SHARING_TYPES = Map([
+const MODEL_DATA_SHARING_TERMS_TYPES = Map([
   [Pb.Survey.DataSharingTerms.Type.PRIVATE, DataSharingType.PRIVATE],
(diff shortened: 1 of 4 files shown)


web/src/app/converters/survey-data-converter.ts
```diff
@@ -64,12 +65,23 @@ const MODEL_STATES = Map([
   [Pb.Survey.State.READY, SurveyState.READY],
 ]);
 
-const MODEL_VISIBILITIES = Map([
+const MODEL_GENERAL_ACCESSES = Map([
```
gino-m, line 68: `MODEL_GENERAL_ACCESS_VALUES` is more correct here.

gino-m (changes requested):
One nit. Note that the PR description says this updates the proto, but there are no proto changes here.

web/src/app/converters/survey-data-converter.ts
```diff
@@ -53,31 +54,42 @@ export const MODEL_ROLES = Map([
   [Pb.Role.VIEWER, Role.VIEWER],
 ]);
 
-const MODEL_SHARING_MODEL_TYPES = Map([
+const MODEL_DATA_SHARING_TYPES = Map([
```
gino-m, line 57: This refers specifically to the terms, so should be included int he name - `MODEL_DATA_SHARING_TERMS_TYPES`

gino-m (changes requested):

gino-m (approved):

PR #2215, opened 2025-05-13
Added site/submission filter based on survey visibility setting

rfontanarosa:
part of #1418
depends on https://github.com/google/ground-platform/pull/2204


gino-m (changes requested):
Please use present active voice in PR titles, and describe the change from the user's perspective if possible.

So instead of:

> Added site/submission filter based on survey visibility setting

Something like:

> Show/hide sites and submissions based on survey visibility setting

Issue #1996, opened 2024-08-27
Translation to French and other languages


jo-spek:
Hi, I've created the French translations for the Android app and would like to proceed creating translations for the platform. Angular comes with the i18n module, making the task comparatively easy.

@rfontanarosa suggested I get you into the loop @vittorino. What do you think is the best way forward?

I could collect and translate the strings pretty easily employing i18n. 
But what would trigger the translation? A language switch button on the main page? Automatic recognition of the users system locale?

Hope you can help.

Best, Jonas

kenstershiro:
Waiting on @gino-m for some questions on the PR from @rfontanarosa

gino-m:
> Waiting on [@gino-m[<img alt="" width="15" height="15" src="chrome-extension://dlebflppeeemcdpidccbiblndppbmjmh/ospo-chrome-ext-logo.png">](https://teams.googleplex.com/gmiceli@google.com)](https://github.com/gino-m) for some questions on the PR from [@rfontanarosa](https://github.com/rfontanarosa)

Can you please link to the question or the PR? I only see #2128 linked here, which is by @jo-spek, and it doesn't have any unresolved comments.

kenstershiro:
@jo-spek needs the I18n file (JSON output) in order to proceed with translations (suggestion to use CrowdIn)

kenstershiro:
@rfontanarosa to provide details to @jo-spek for labelling the strings

kenstershiro:
Update: 
- Translations of French and Vietnamese labels done through CrowdIn (incl. Android app)
- Switched to new translation approach (PR open above, needs approval).

PR #2218, opened 2025-05-27
Switch from ngx-translate to @angular/localize for i18n

rfontanarosa:
part of #1996

Moved the internationalization system from the ngx-translate library to the official @angular/localize package.

This change necessitates a separate build for each language but offers a significant performance boost.

### DIFF
--- a/web/src/main.ts
+++ b/web/src/main.ts
@@ -1,3 +1,5 @@
+/// <reference types="@angular/localize" />
+
 /**
  * Copyright 2019 The Ground Authors.
  *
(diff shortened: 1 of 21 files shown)


web/src/main.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="@angular/localize" />
```
gino-m, line 1: Delete?

rfontanarosa: it is something added automatically by @angular/localize: https://v17.angular.io/guide/i18n-common-add-package

```
It adds types: ["@angular/localize"] in the TypeScript configuration files as well as the reference to the type definition of @angular/localize at the top of the main.ts file.
```

Issue #2225, opened 2025-07-08
[Create/edit survey] Only allow admins to create "public" surveys


gino-m:
@rfontanarosa Perhaps the easiest way is to add a flag to the user profile in the db? Something like a new field with "roles: ['admin']", perhaps?

@kenstershiro This is blocking public surveys so we need to slot it into work for the upcoming training milestone.

rfontanarosa:
for now we can just hide the public option

gino-m:
To be discussed in the broader context of what is an "admin", how they're defined, what they can do.

PR #2229, opened 2025-07-11
Hides Public General Access option

rfontanarosa:
related to #2225

### DIFF
--- a/web/src/app/components/general-access-control/general-access-control.component.ts
+++ b/web/src/app/components/general-access-control/general-access-control.component.ts
@@ -41,14 +41,14 @@ const generalAccessLabels = Map<SurveyGeneralAccess, any>([
       label: $localize`:@@app.labels.unlisted:Unlisted`,
     },
   ],
-  [
-    SurveyGeneralAccess.PUBLIC,
-    {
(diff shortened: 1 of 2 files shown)


web/src/app/components/general-access-control/general-access-control.component.ts
```diff
@@ -41,14 +41,14 @@ const generalAccessLabels = Map<SurveyGeneralAccess, any>([
...
-    },
-  ],
+  // [
+  //   SurveyGeneralAccess.PUBLIC,
+  //   {
+  //     description: $localize`:@@app.texts.generalAccess.public:Everyone can collect data for this survey`,
+  //     icon: 'public',
+  //     label: $localize`:@@app.labels.public:Public`,
+  //   },
+  // ],
```
gino-m: Please delete commented out code. We can add it back from history when needed.

gino-m: Alternatively, you filter these out based on a placeholder function which always returns false.

gino-m:
How does the user know what Unlisted means? Does this match the UX designs, or is it a temporary solution?

PR #2230, opened 2025-07-16
Fixes QR Code and default AppLink url

### LINKED ISSUE #2200: [Feature] AppLink adoption for android links

rfontanarosa:
closes #2200

https://github.com/user-attachments/assets/75466dc7-8e2c-4caf-9aeb-15901401e9fa

### DIFF
--- a/web/src/app/services/navigation/navigation.service.ts
+++ b/web/src/app/services/navigation/navigation.service.ts
@@ -300,7 +300,7 @@ export class NavigationService {
   }
 
   getSurveyAppLink(surveyId: string): string {
-    return `${this.document.location.origin}/android/${SURVEY_SEGMENT}/${surveyId}`;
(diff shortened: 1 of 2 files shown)


web/src/app/services/navigation/navigation.service.ts
```diff
@@ -300,7 +300,7 @@ export class NavigationService {
   }
 
   getSurveyAppLink(surveyId: string): string {
-    return `${this.document.location.origin}/android/${SURVEY_SEGMENT}/${surveyId}`;
+    return `${this.document.location.origin}/${SURVEY_SEGMENT}/${surveyId}`;
```
gino-m: If we do this, how do we differentiate links to the Android app from links to the web app?

gino-m (changes requested):

PR #2235, opened 2025-07-18
Fix: Improve retrieval of properties from LOI documents

rfontanarosa:
- query pagination
- query data instead of proto conversion
- promise.all/map instead of foreach (order is not important)
- select only necessary fields

### DIFF
--- a/functions/src/export-csv.ts
+++ b/functions/src/export-csv.ts
@@ -22,7 +22,7 @@ import {getDatastore} from './common/context';
 import * as HttpStatus from 'http-status-codes';
 import {DecodedIdToken} from 'firebase-admin/auth';
 import {List} from 'immutable';
-import {QuerySnapshot} from 'firebase-admin/firestore';
+// import {QuerySnapshot} from 'firebase-admin/firestore';
 import {timestampToInt, toMessage} from '@ground/lib';
 import {GroundProtos} from '@ground/proto';
 import {toGeoJsonGeometry} from '@ground/lib';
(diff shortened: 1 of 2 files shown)


gino-m:
Is this draft PR still valid / working?

functions/src/export-csv.ts
```diff
@@ -77,15 +93,15 @@ export async function exportCsvHandler(
     await Promise.all(
       snapshot.docs.map(async doc => {
         const loi = doc.data();
-        if (!(ownerId ? loi[9] === 2 && loi[5] === ownerId : true)) return;
+        if (!(ownerIdFilter ? loi[9] === 2 && loi[5] === ownerIdFilter : true))
```
gino-m: This is a bit hard to grok at first glance.. Could you maybe simplify this into two if statements without using the `?` operator?

functions/src/export-csv.ts
```diff
@@ -93,8 +93,11 @@ export async function exportCsvHandler(
     await Promise.all(
       snapshot.docs.map(async doc => {
         const loi = doc.data();
-        if (!(ownerIdFilter ? loi[9] === 2 && loi[5] === ownerIdFilter : true))
-          return;
+        if (ownerIdFilter) {
+          if (!(loi[9] === 2 && loi[5] === ownerIdFilter)) {
```
gino-m: Do we have named constants we can refer to field numbers instead of using literals?

functions/src/export-csv.ts
```diff
@@ -93,8 +93,11 @@ export async function exportCsvHandler(
     await Promise.all(
       snapshot.docs.map(async doc => {
         const loi = doc.data();
-        if (!(ownerIdFilter ? loi[9] === 2 && loi[5] === ownerIdFilter : true))
-          return;
+        if (ownerIdFilter) {
```
gino-m: Also, I think this is still a bit confusing.. maybe we want something like this:

```ts
if (canViewAll || loiOwnerId === userId) {
  return extractLoiProperties(loi);
} else {
  return undefined;
}
```

Then we don't need `ownerIdFilter` at all

functions/src/export-csv.ts
```diff
@@ -85,8 +86,31 @@ export async function exportCsvHandler(
...
+
+  const loiProperties = new Set<string>();
+  let query = db.fetchPartialLocationsOfInterest(surveyId, jobId, 1000);
+  let lastVisible = null;
+  do {
+    const snapshot = await query.get();
+    if (snapshot.empty) break;
+    await Promise.all(
+      snapshot.docs.map(async doc => {
+        const loi = doc.data();
```
gino-m: Can we extract this lambda into its own function for clarity?

PR #2238, opened 2025-07-28
Allows survey organizers to see Unlisted surveys

### LINKED ISSUE #2237: [Feature] Unlisted surveys must be visible to data collectors in the survey list
At the moment only Survey owners can see unlisted surveys in the survey list.

rfontanarosa:
closes #2237


web/src/app/services/data-store/data-store.service.ts
```diff
@@ -179,7 +179,11 @@ export class DataStoreService {
       .collection(SURVEYS_COLLECTION_NAME, ref =>
         ref
           .where(s.generalAccess, '==', GeneralAccess.UNLISTED)
-          .where(s.ownerId, '==', userId)
+          .where(
```
gino-m, line 179: The conditions in Firestore are applied incrementally, meaning this query will include results where the survey is UNLISTED _AND_ the user is the survey organizer.

gino-m (changes requested):

web/src/app/services/data-store/data-store.service.ts
rfontanarosa: yes, there are three different queries. one for public, one for restricted and one for unlisted. this is is for the unlisted one and only to show them in the survey list.

gino-m: Ah ok, can you maybe rename the variable and/or add a comment to make it more obvious that we're looking for "unlisted surveys which the current user manages?"

gino-m: Unresolving conversation in case you missed my last suggestion 👍

rfontanarosa: done. i've chosen to rename the variables. :)

PR #2249, opened 2025-09-12
Allow viewers to export into GeoJSON and CSV formats

### LINKED ISSUE #2244: Viewer can't see collected data
Came across this case where as a Viewer I cannot see the geolocation entries but export them as CSV (including Whisp columns).

As Survey Organizer I can see the entries and export them, but the exported CSV doesn't hold the Whisp columns.

@rfontanarosa

rfontanarosa:
closes #2244

**Survey's General Access**
- Public or Unlisted: Every user can export
- Restricted: Only users with a role can export

**Survey's Data Visibility** + **User's Role**
- Role:Organizer: Doesn't check the ownership
- Role:Collector and Role:Viewer: Check the data visibility
- DataVisibility:AllSurveyParticipants: Users can see each other data
- DataVisibility:ContributorsAndOrganizers: Check the ownership ONLY if user is not Survey Organizer

**LOI's Source**
- Imported: Doesn't check the ownership
- Field Data: Check the ownership

### DIFF
--- a/functions/src/common/utils.ts
+++ b/functions/src/common/utils.ts
@@ -34,9 +34,11 @@ export function isAccessibleLoi(
   loi: Pb.ILocationOfInterest,
   ownerId: string | null
 ) {
-  return loi.source === Pb.LocationOfInterest.Source.IMPORTED
-  || ownerId === null
-  || loi.ownerId === ownerId;
+  return (
+    loi.source === Pb.LocationOfInterest.Source.IMPORTED ||
+    ownerId === null ||
+    loi.ownerId === ownerId
+  );
 }
 
 export function stringFormat(s: string, ...args: any[]): string {


functions/src/export-csv.ts
```diff
@@ -47,11 +47,11 @@ export async function exportCsvHandler(
     res.status(HttpStatus.NOT_FOUND).send('Survey not found');
     return;
   }
-  if (!canExport(user, surveyDoc)) {
+  if (!hasRole(user, surveyDoc)) {
     res.status(HttpStatus.FORBIDDEN).send('Permission denied');
     return;
   }
-  const ownerId = canImport(user, surveyDoc) ? undefined : userId;
+  const ownerId = isDataCollector(user, surveyDoc) ? userId : undefined;
```
gino-m: Something's not right here; with this change, `ownerId` doesn't always represent the owner's ID. It gets used later when checking which LOIs should be included in the output; I assume that logic will no longer be correct if this doesn't actually contain the owner ID.

Instead, can we make these behaviors more explicit? What are the different states and conditions which can be applied here? I can thing of a few:

* Is this survey visible? (public, unlisted, or restricted). If restricted, do they have a `role` in this survey?
* Is a submission visible? If survey owner, yes. If anyone else, is "share with others" enabled? If so, yes. Otherwise, are they the owner/creator of the submission?
* Same question for LOIs.

functions/src/export-geojson.ts
```diff
@@ -43,11 +43,11 @@ export async function exportGeojsonHandler(
     res.status(HttpStatus.NOT_FOUND).send('Survey not found');
     return;
   }
-  if (!canExport(user, surveyDoc)) {
+  if (!hasRole(user, surveyDoc)) {
     res.status(HttpStatus.FORBIDDEN).send('Permission denied');
     return;
   }
-  const ownerId = canImport(user, surveyDoc) ? undefined : userId;
+  const ownerId = isDataCollector(user, surveyDoc) ? userId : undefined;
```
gino-m: Same comments here.

functions/src/export-csv.ts
rfontanarosa: i've changed some variable names to make it more clear, in any case we should discuss the general behaviour of the export functions and change it in a separate PR.

functions/src/export-csv.ts
```diff
@@ -75,7 +75,10 @@ export async function exportCsvHandler(
   }
   const {name: jobName} = job;
 
+  const isOrganizer = hasOrganizerRole(user, surveyDoc);
+
   const filterByOwnerId =
+    !isOrganizer &&
```
gino-m: This condition and how it's applied is getting a little confusing. Can we instead invert it, renaming the condition something like `canViewAll` or similar?

functions/src/common/auth.ts
```diff
@@ -94,17 +94,26 @@ function getRole(
...
+  const role = getRole(user, survey);
+  return !!role && [Pb.Role.SURVEY_ORGANIZER].includes(role);
+}
+
 export function canExport(
   user: DecodedIdToken,
   survey: DocumentSnapshot
 ): boolean {
+  const generalAccess = survey.get(s.generalAccess);
+  if (generalAccess !== Pb.Survey.GeneralAccess.RESTRICTED) return true;
```
gino-m: I'd check each value in the enum here instead of using `!==`; that will be more resilient to changes if we add more GeneralAccess types

functions/src/common/utils.ts
```diff
@@ -14,6 +14,22 @@
...
  */
 
+import {GroundProtos} from '@ground/proto';
+
+import Pb = GroundProtos.ground.v1beta1;
+
+/**
+ * Checks if a Location of Interest (LOI) is accessible to a given user.
+ */
+export function isAccessibleLoi(
```
gino-m, line 33: Please document the behavior when ownerId is null.

functions/src/common/utils.ts
```diff
@@ -14,6 +14,22 @@
...
+  loi: Pb.ILocationOfInterest,
+  ownerId: string | null
+) {
+  if (loi.source === Pb.LocationOfInterest.Source.IMPORTED) return true;
+  if (!ownerId) return true;
```
gino-m: `ownerId === null` is safer, since as written this will also return `true` when `ownerId` is empty.

functions/src/common/utils.ts
```diff
@@ -14,6 +14,31 @@
  * limitations under the License.
  */
 
+import {GroundProtos} from '@ground/proto';
```
gino-m, line 17: At some point we should rename this file to make it more clearly scoped. Ok for now.

functions/src/common/utils.ts
```diff
@@ -14,6 +14,31 @@
...
+ * @param loi The Location of Interest object to check.
+ * @param ownerId The ID of the user requesting access. Pass `null` to bypass the ownership check.
+ * @returns True if the LOI is accessible to the user, false otherwise.
+ */
+export function isAccessibleLoi(
+  loi: Pb.ILocationOfInterest,
+  ownerId: string | null
+) {
+  if (loi.source === Pb.LocationOfInterest.Source.IMPORTED) return true;
+  if (ownerId === null) return true;
```
gino-m: Let's combine these three lines into a single conditions?

```ts
return loi.source === Pb.LocationOfInterest.Source.IMPORTED
  || ownerId === null
  || loi.ownerId === ownerId
```

gino-m (changes requested):
Some small suggestions

functions/src/common/utils.ts
rfontanarosa: ok. noted.

PR #2251, opened 2025-09-12
Display LOI conversion error instead of returning empty value

### LINKED ISSUE #2250: Unable to display LOIs due to "Unsupported or missing geometry type" error
Sometimes, LOIs are saved into the database with an empty geometry field.

This causes an error during data conversion. When this happens, instead of logging the problem and filtering out the problematic entries, the function crashes without returning a value.

Image

rfontanarosa:
closes #2250 

Screenshot 2025-09-12 alle 12 25 49

### DIFF
--- a/web/src/app/converters/loi-data-converter.ts
+++ b/web/src/app/converters/loi-data-converter.ts
@@ -58,6 +58,8 @@ export function loiDocToModel(
       pb.source === Pb.LocationOfInterest.Source.IMPORTED
     );
   } catch (e: any) {
-    return Error(`${e.message} in loi ${id}`);
+    const wrappedError = new Error(`Error converting LOI with ID ${id}`);
+    (wrappedError as any).cause = e;
+    return wrappedError;
   }
 }


web/src/app/converters/loi-data-converter.ts
```diff
@@ -45,15 +45,19 @@ export function loiDocToModel(
...
+    return new LocationOfInterest(
+      id,
+      pb.jobId,
+      geometry,
+      properties,
+      pb.customTag,
+      pb.source === Pb.LocationOfInterest.Source.IMPORTED
+    );
+  } catch (e: any) {
+    return Error(`${e.message} in loi ${id}`);
```
gino-m: This will swallow the original stack trace, making issues harder to debug. Try this instead:

```ts
    return Error('Error converting LOI with ID ' + id, { cause: e });
```

rfontanarosa: it seems that Error constructor is available from ES2022, but we are using ES2018.

```
  } catch (e: any) {
    const wrappedError = new Error(`Error converting LOI with ID ${id}`);
    (wrappedError as any).cause = e;
    return wrappedError;
  }
```

PR #2253, opened 2025-09-16
Fix GeoJSON properties order

### LINKED ISSUE #2252: Export GeoJSON properties order issue
Export GeoJSON have a different properties order than Export CSV

rfontanarosa:
closes #2252

### DIFF
--- a/functions/src/export-geojson.spec.ts
+++ b/functions/src/export-geojson.spec.ts
@@ -127,8 +127,8 @@ describe('export()', () => {
     [l.submissionCount]: 0,
     [l.source]: Pb.LocationOfInterest.Source.FIELD_DATA,
     [l.properties]: {
-      name: {[pr.stringValue]: 'Dinagat Islands'},
       area: {[pr.numericValue]: 3.08},
+      name: {[pr.stringValue]: 'Dinagat Islands'},
     },
     [l.ownerId]: userId,
   };
@@ -159,7 +159,7 @@ describe('export()', () => {
         features: [
           {
             type: 'Feature',
-            properties: {name: 'Dinagat Islands', area: 3.08},
+            properties: {area: 3.08, name: 'Dinagat Islands'},
             geometry: {type: 'Point', coordinates: [125.6, 10.1]},
           },
         ],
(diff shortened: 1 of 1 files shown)


functions/src/export-geojson.ts
```diff
@@ -146,13 +147,13 @@ function getFileName(jobName: string | null) {
 
 function propertiesPbToModel(pb: {
   [k: string]: Pb.LocationOfInterest.IProperty;
-}): Map<string, string | number> {
-  const properties: {[k: string]: string | number} = {};
-  for (const k of Object.keys(pb)) {
-    const v = pb[k].stringValue || pb[k].numericValue;
+}): [string, string | number][] {
+  const entries: [string, string | number][] = [];
+  for (const k of Reflect.ownKeys(pb)) {
```
gino-m: Is this attempting to preserve the order of properties as they appear in the map in the db? Does that map preserve order, all the way from import to how they're persisted in Firestore?

rfontanarosa: the problem was not the order itself but the costincency with the CSV order.

In any case, at some point the order change.

This is what we have in the DocumentData:

```
name: { '1': 'ttt' },
whisp_plotId: { '1': '1' },
whisp_external_id: { '1': 'null' },
whisp_Area: { '2': 0.03500000014901161 },
whisp_Geometry_type: { '1': 'Polygon' },
```

this is what the firebase console shows (it probably applies a sort):

Screenshot 2025-09-26 alle 17 16 53

this is what we have after conversion (toMessage(row.data(), Pb.LocationOfInterest)):

```
    "whisp_nBR_PRODES_deforestation_Brazil_before_2020": 0,
    "whisp_nBR_MapBiomas_col9_coffee_2020": 0,
    "whisp_ESRI_2023_crop": 0,
    "whisp_European_Primary_Forest": 0,
```

so probably the solution is sort the keys at convertion time (the default one is fine for now) so we can have the control of it:

```
function propertiesPbToObject(pb: {
  [k: string]: Pb.LocationOfInterest.IProperty;
}): {[k: string]: string | number} {
  const properties: {[k: string]: string | number} = {};
  for (const k of Object.keys(pb).sort()) {
    const v = pb[k].stringValue || pb[k].numericValue;
    if (v !== null && v !== undefined) {
      properties[k] = v;
    }
  }
  return properties;
}

```

gino-m (changes requested):
LGTM, please add test to verify ordering.

PR #2271, opened 2025-10-17
Prevent unauthorized users (non-registered) from creating surveys

### LINKED ISSUE #2262: Unpasslisted users can login, but not proceed in survey creation
Users who have not registered through the [Passlist Google Form](https://forms.gle/RgnFcCz9ma1gLr6v7) can login at ground.openforis.org. They get to the main screen and can click "Create Survey" to get as far as this:

Image

When they click 'Continue' though, nothing happens anymore. 

Getting this far without having been passlisted creates a lot of confusion and support requests from users. Can we please let them fail already at login again?

(Probably related to unlisted surveys and passlisting-free use of the Android app.)

rfontanarosa:
closes #2262 

Screenshot 2025-10-17 alle 19 00 03

### DIFF
--- a/web/src/app/services/auth/passlist.guard.ts
+++ b/web/src/app/services/auth/passlist.guard.ts
@@ -34,9 +34,9 @@ export const passlistGuard: CanActivateFn = async (): Promise<
 
   if (isPasslisted) {
     return true;
-  } else {
-    const url = await navigationService.getAccessDeniedLink();
-
-    return router.parseUrl(url || '/');
   }
+
+  const url = await navigationService.getAccessDeniedLink();
+
+  return router.parseUrl(url || '/');
 };


web/src/app/pages/edit-survey/job-dialog/job-dialog.component.ts
```diff
@@ -90,6 +91,12 @@ export const dialogConfigs: Record<DialogType, DialogConfig> = {
     content: $localize`:@@app.dialogs.invalidSurvey.content:To publish changes, fix any outstanding issues with your survey.`,
     backButtonLabel: $localize`:@@app.labels.goBack:Go back`,
   },
+  [DialogType.SurveyCreationDenied]: {
+    title: $localize`:@@app.dialogs.surveyCreationDenied.title:Registration Required`,
```
gino-m: Please use "sentence case" (""Registration required") instead for consistency and so that Longer Heading Don't End Up Looking Like This :)

web/src/app/pages/edit-survey/job-dialog/job-dialog.component.ts
```diff
@@ -90,6 +91,12 @@ export const dialogConfigs: Record<DialogType, DialogConfig> = {
...
+    content: $localize`:@@app.dialogs.surveyCreationDenied.content:You must register for an account to create a new survey. Click "Continue" to be redirected to the registration form.`,
```
gino-m, line 103: > You do not have permission to create surveys. Click "Continue" to request access.

web/src/app/components/survey-list/survey-list.component.spec.ts
```diff
@@ -195,13 +195,13 @@ describe('SurveyListComponent', () => {
     expect(surveyCards.length).toBe(4, 'Should display 2 survey cards');
   });
 
-  it('should go to create survey page when add card is clicked', () => {
-    clickCard(fixture, 'add-card');
+  // it('should go to create survey page when add card is clicked', () => {
```
gino-m: Did you mean to uncomment these lines?

web/src/app/components/survey-list/survey-list.component.ts
```diff
@@ -110,8 +119,24 @@ export class SurveyListComponent implements OnInit, OnDestroy {
...
+          data: {dialogType: DialogType.SurveyCreationDenied},
+          panelClass: 'small-width-dialog',
+        })
+        .afterClosed()
+        .subscribe(async (result: DialogData) => {
+          if (!result) return;
+
+          this.navigationService.navigateToSubscriptionForm();
+        });
+    } else {
```
gino-m: Return early (short circuit) so that the base case doesn't need to be nested in the else statement

web/src/app/services/auth/passlist.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class PasslistGuard {
+  constructor(
+    private authService: AuthService,
+    private navigationService: NavigationService
+  ) {}
+
+  async canActivate(): Promise<void> {
```
gino-m: "activate" is more of an internal term that I think we'd want here... Also, this does more than just check whether the user can activate the survey, it actually redirects to the form. Perhaps something like "checkSurveyAcls" might be more appropriate?

web/src/app/services/navigation/navigation.service.ts
```diff
@@ -181,6 +183,18 @@ export class NavigationService implements OnDestroy {
     this.router.navigate([SURVEY_SEGMENT, SURVEY_ID_NEW]);
   }
 
+  async getAccessDeniedLink(): Promise<string | undefined> {
+    const accessDeniedMessage = await this.dataStore.getAccessDeniedMessage();
```
gino-m, line 187: Where do we document the contents of the DB config? Can we create a wiki page or readme for this? Ok to do after merging as well.

web/src/app/components/survey-list/survey-list.component.spec.ts
rfontanarosa: restored

web/src/app/services/navigation/navigation.service.ts
rfontanarosa: the whole config collection should be documented

web/src/app/services/auth/passlist.guard.ts
rfontanarosa: the property canActivate is part of the angular router's configuration used to list which route guards should run before navigating to a route.

it seems there is a modern way to write guards (using functions instead of classes). i've implemented it following this guide: https://stackoverflow.com/questions/75564717/angulars-canactivate-interface-is-deprecated-how-to-replace-it

PR #2274, opened 2025-10-28
Fix survey reload when path changes

### LINKED ISSUE #2273: Job add/duplication doesn't work (regression)
https://github.com/user-attachments/assets/af3d2020-04ef-4d12-b27a-06468c58c067

rfontanarosa:
closes #2273


web/src/app/pages/edit-survey/edit-survey.component.ts
```diff
@@ -61,16 +61,19 @@ export class EditSurveyComponent implements OnInit {
...
-        if (surveyId) {
-          this.surveyId = surveyId;
-          this.surveyService.activateSurvey(surveyId);
-          await this.draftSurveyService.init(surveyId);
-          this.draftSurveyService
-            .getSurvey$()
-            .subscribe(survey => (this.survey = survey));
-        }
-      })
+      this.navigationService
```
gino-m, line 64: Does edit survey manage its own loading state (e.g., show a spinner when loading?). If not, can we pass the survey in as an `@Input` once the draft survey is created/loaded?

PR #2327, opened 2025-12-19
Update firebase emulator data and add github action for Android e2e tests

andreia-ferreira:
Towards: https://github.com/google/ground-android/issues/3365

To make Android end to end tests reliable and self-contained, we need to pre-populate the Firebase emulator with a dedicated test survey. This PR adds a new folder with this test data `test-android` and a new command `nx start-android-test-data` to launch an emulator based on this data.

A separate github action was also added in order to run this from the ground-android repo. This action will be used by a workflow hosted in the `ground-android` repo

### DIFF
--- a/functions/src/session-login.ts
+++ b/functions/src/session-login.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { https, Response, logger } from 'firebase-functions/v1';
+import { Response, https, logger } from 'firebase-functions/v1';
 import { setSessionCookie } from './common/auth';
 import { UNAUTHORIZED } from 'http-status-codes';
 
--- a/functions/src/testing/http-test-helpers.ts
+++ b/functions/src/testing/http-test-helpers.ts
@@ -17,7 +17,7 @@
 import { buffer } from 'stream/consumers';
 import { FormDataEncoder } from 'form-data-encoder';
 import { FormData } from 'formdata-node';
-import { https, Response } from 'firebase-functions/v1';
+import { Response, https } from 'firebase-functions/v1';
 
 export async function createPostRequestSpy(
   args: object,
--- a/web/src/app/components/create-survey/create-survey.component.ts
+++ b/web/src/app/components/create-survey/create-survey.component.ts
@@ -18,7 +18,8 @@ import '@angular/localize/init';
 
 import { ChangeDetectorRef, Component, OnInit, ViewChild } from '@angular/core';
 import { List } from 'immutable';
-import { Subscription, combineLatest, filter } from 'rxjs';
+import { Subscription, filter } from 'rxjs';
+import { map, switchMap } from 'rxjs/operators';
 
 import { DataSharingTermsComponent } from 'app/components/create-survey/data-sharing-terms/data-sharing-terms.component';
 import { JobDetailsComponent } from 'app/components/create-survey/job-details/job-details.component';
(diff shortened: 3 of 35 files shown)


firebase.local.json
```diff
@@ -13,6 +13,9 @@
     ]
   },
   "emulators": {
+    "auth": {
+      "port": 9099
+    },
```
andreia-ferreira, line 18: I see that this has been removed in https://github.com/google/ground-platform/pull/2338 but it's needed for running the android project with the local firebase emulator so I'm re-adding it here. @gino-m could you please confirm if this is ok?

gino-m: Are you sure it's needed? Isn't auth disabled in the local version? I can't see how it would work today, since we never set up a test account in the auth emulator.

andreia-ferreira: without this the Android app crashes with : `com.google.firebase.FirebaseException: An internal error has occurred. [ Failed to connect to /10.0.2.2:9099 ]`

The Android app needs FirebaseAuth to initialize successfully when running against the local emulator. For local builds we rely on anonymous sign-in, which still requires the auth emulator to be configured, which currently is done like this:

```
    val auth = FirebaseAuth.getInstance()
    if (USE_EMULATORS) {
      // Use the auth emulator so we can sign-in anonymously during dev.
      auth.useEmulator(EMULATOR_HOST, AUTH_EMULATOR_PORT)
    }
```

gino-m: Interesting.. I'm surprised it's needed for anonymous sign in. Sure, please re-enable it!

andreia-ferreira:
> Would it be easier to use https://firebase.google.com/docs/test-lab instead of standing up an emulator?

with our current setup, working with a local Firebase emulator, it wouldn't be possible I believe

> Can we set it up so that the same GitHub Action is run from both repos? The action would check out most codebases at a particular commit, build, and run. Wdyt?

That would be ideal, but this PR adds static test data for Android, it's not generated dynamically yet so triggering the workflow on this repo wouldn't do much right now. Btw, I added some documentation in the Draft PR I'm working on the Android side: https://github.com/google/ground-android/pull/3429/files#diff-bfcda924045749dc64968acb3b1a38887b094a6a793d11cd7fe2bc4a42bb8d23

gino-m:
> with our current setup, working with a local Firebase emulator, it wouldn't be possible I believe

That sounds right. In a follow-up, perhaps at the end of integration tests (I wouldn't call them e2e at this point), at the end of tests we write the data back to the repo so that the other repo can use it as input? I.e., changes to `ground-platform` will trigger integration tests which write its output (eventually at multiple points in the process) to GitHub or elsewhere. `ground-android` would use that output as inputs to its tests. We could then do the opposite, without `ground-android` writing its db contents someplace which `ground-platform` could access. Wdyt?

> 
> > Can we set it up so that the same GitHub Action is run from both repos? The action would check out most codebases at a particular commit, build, and run. Wdyt?
> 
> That would be ideal, but this PR adds static test data for Android, it's not generated dynamically yet so triggering the workflow on this repo wouldn't do much right now. Btw, I added some documentation in the Draft PR I'm working on the Android side: https://github.com/google/ground-android/pull/3429/files#diff-bfcda924045749dc64968acb3b1a38887b094a6a793d11cd7fe2bc4a42bb8d23

Ok please disregard for now.

Issue #2304, opened 2025-12-11
[Code health] Simplify npm scripts


gino-m:
Align with expectations of developers familiar with npm. For ex, the base case should be

```$
$ npm i
$ npm run build
$ npm run start
```

And maybe `npm run dev`?

gino-m:
Closing in favor of #2307

PR #2330, opened 2025-12-22
Remove copy-asset-links command

rfontanarosa:
Towards https://github.com/google/ground-platform/issues/2304

it works: https://ground-dev-sig.web.app/.well-known/assetlinks.json

Screenshot 2026-01-02 alle 23 40 42


web/src/assets/.well-known/assetlinks.json
```diff
@@ -0,0 +1,9 @@
+[{
+  "relation": ["delegate_permission/common.handle_all_urls"],
+  "target": {
+    "namespace": "android_app",
+    "package_name": "org.groundplatform.android",
+    "sha256_cert_fingerprints":
+    ["9f:f1:da:1e:97:93:c7:93:9a:8d:52:24:af:8b:a6:07:da:0a:10:db:1c:d2:40:47:5c:07:eb:be:c4:e8:57:80"]
```
gino-m, line 7: Please replace values placeholder text here and move into openforis/ground-platform, since this refers to the openforis deployment specifically.

rfontanarosa: it's a bit complicated because firebase hosting only allows one public directory per deployment, which is not set as "web/dist/web".

if we want to keep .well-known physically outside of the web folder, we have two choices: multiple hosting or deploy-time merge using the following command:

```
    "predeploy": "cp -r .well-known web/dist/web/",
```

to make it compatible with all OS we should use shx.

wdyt? @gino-m

rfontanarosa: argh, i totally misunderstood your comment. you ment to move this under openforis repo. anyway, we can keep this assetLinks file because it is related to dev-sig environment.

gino-m: Ahh ok I didn't realize, that make sense. Please disregard!

web/src/assets/.well-known/assetlinks.json
```diff
@@ -0,0 +1,9 @@
+[{
```
gino-m, line 1: Nit: This filename is usually `assetlinks.json` (all lowercase).

gino-m (approved):
Approved with one nit.

PR #2353, opened 2026-01-02
Consolidate build deps

### LINKED ISSUE #2342: [Code health] Move common dev dependencies to root package
This is best practice and ensures a single source of truth for dev deps versions.

We should also move any dev deps which have been accidentally added as runtime deps.

gino-m:
Fixes #2342


gino-m:
@rfontanarosa will redo this after #2334.

PR #2355, opened 2026-01-02
Consolidate common dev deps into root package

rfontanarosa:
fixes https://github.com/google/ground-platform/issues/2342

### DIFF
--- a/.github/workflows/check-web.yml
+++ b/.github/workflows/check-web.yml
@@ -39,13 +39,7 @@ jobs:
       - name: Install deps
         run: npm install
 
-      - name: Build proto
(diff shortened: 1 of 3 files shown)


.github/workflows/check-web.yml
```diff
@@ -39,7 +39,13 @@ jobs:
       - name: Install deps
         run: npm install
 
-      - name: Build web and deps
+      - name: Build proto
+        run: npx nx build proto
+
+      - name: Build lib
```
gino-m: If this is necessary something is going wrong with the nx config, since should build dep transitively. Does nx build web not work locally?

Issue #2339, opened 2025-12-27
[DevX] Save and restore Nx cache for fast CI


gino-m:
We probably can't use Nx Cloud in our repo, so we'll need to save and restore the cache ourselves

PR #2364, opened 2026-01-06
Test Nx #2339

gino-m:
Verifies #2339


gino-m:
I see a cache hit in logs, so the fix caching appears to cut the runtime of checks in half.

Before:

Screenshot 2026-01-06 at 1 53 57 PM

After:

Screenshot 2026-01-06 at 2 38 45 PM

We could get additional gains by only build each monorepo package once rather than once per job.

PR #2365, opened 2026-01-06
Test check triggers

gino-m:
Towards #2339


gino-m:
Verified that unchanged packages were skipped as expected:

image

PR #2385, opened 2026-01-14
Migrate workspace from npm to pnpm

### LINKED ISSUE #2343: [Code health] Introduce pnpm for single-truth dep versioning
This way versions can be defined globally but specific deps can be defined per-package.

rfontanarosa:
closes #2343


.github/actions/setup/action.yml
```diff
@@ -18,12 +18,19 @@ runs:
...
           nx-${{ runner.os }}-${{ inputs.cache-key-suffix }}-
 
+    - name: Install pnpm
+      uses: pnpm/action-setup@v3
+      with:
+        version: '9'
+
     - name: Set up Node.js
-      uses: actions/setup-node@v6
+      uses: actions/setup-node@v4
```
gino-m: Do we have to downgrade from v6 to v4 here and elsewhere?

.github/actions/verify-test/action.yml
```diff
@@ -38,49 +39,36 @@ runs:
         restore-keys: |
           ${{ runner.os }}-firebase-emulators-
 
-    - name: Set up Node.js
-      uses: actions/setup-node@v2
+    - name: Install pnpm
```
gino-m: Can we refactor common setup here and in other actions into setup/action.yml while we're making changes here?

.github/workflows/check.yml
```diff
@@ -133,14 +133,26 @@ jobs:
         with:
           cache-key-suffix: web
 
+      - name: Install Chrome
```
gino-m, line 136: Can we move this into the setup action?

functions/package.json
```diff
@@ -30,24 +30,23 @@
   },
   "dependencies": {
     "@fast-csv/format": "4.3.5",
-    "@ground/lib": "*",
-    "@ground/proto": "*",
+    "@ground/lib": "workspace:*",
```
gino-m, line 33: Why are only some version numbers defined in `catalog:` and others defined here?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,47 @@
+packages:
+  - 'proto'
+  - 'lib'
+  - 'functions'
+  - 'web'
+  - 'e2e-tests'
+
+catalog:
+  # Shared general dependencies
```
gino-m: General -> runtime?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,47 @@
...
+  prettier: ^3.7.4
+  gts: ^7.0.0
+  '@eslint/eslintrc': ^3.3.3
+  '@typescript-eslint/eslint-plugin': ^8.50.1
+  '@typescript-eslint/parser': ^8.50.1
+
+  # Node & Tooling
+  nx: 22.2.1
+  ts-node: ^10.9.2
+  '@types/node': ^20.19.27
```
gino-m: Please split dev dependencies into a separate section.

functions/package.json
rfontanarosa: because they are used only in that specific workspace and to have evidence that they are local dependencies and not shared dependencies. do i have to move everything into the catalog? i understand the point: that way all the packages are visible in a single file...

pnpm-workspace.yaml
rfontanarosa: done!

gino-m (dismissed):

PR #2392, opened 2026-01-19
Dependency cleanup and library upgrades

rfontanarosa:
towards #2343

### DIFF
--- a/.github/actions/start-emulator/action.yml
+++ b/.github/actions/start-emulator/action.yml
@@ -25,7 +25,6 @@ runs:
       uses: actions/checkout@v4
       with:
         repository: ${{ inputs.platform-repository }}
-        path: ground-platform
 
     - name: Load cached Firebase emulator
       uses: actions/cache@v4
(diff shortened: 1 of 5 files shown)


functions/src/export-csv.spec.ts.ignore
```diff
@@ -23,7 +23,7 @@ import {
   createResponseSpy,
 } from './testing/http-test-helpers';
 import {DecodedIdToken} from 'firebase-admin/auth';
-import HttpStatus from 'http-status-codes';
+import {StatusCodes} from 'http-status-codes';
```
gino-m, line 26: Was this file accidentally submitted?

pnpm-workspace.yaml
```diff
@@ -34,13 +34,11 @@ catalogs:
     firebase: ^10.12.2
     firebase-admin: 12.1.0
     firebase-functions: ^6.6.0
-    geojson: 0.5.0
     google-auth-library: 6.1.3
     googleapis: 64.0.0
-    http-status-codes: 1.4.0
+    http-status-codes: 2.3.0
```
gino-m: Let' use "latest major" SemVer versioning here and throughout for consistency. Using "Exact" (1.2.3) is generally discouraged in pnpm workspaces because it breaks pnpm's ability to deduplicate packages, leading to bloated install sizes and longer install times.

functions/src/export-csv.spec.ts.ignore
rfontanarosa: I updated it now to prevent any confusion or conflicts when we eventually re-enable and modify in the future.

pnpm-workspace.yaml
```diff
@@ -36,7 +36,7 @@ catalogs:
     firebase-functions: ^6.6.0
     google-auth-library: 6.1.3
     googleapis: 64.0.0
-    http-status-codes: 2.3.0
+    http-status-codes: ^2.3.0
```
gino-m, line 39: While we're here, please use latest major (`^`) for all versions in this file, not just `http-status-codes`

gino-m (dismissed):

PR #2417, opened 2026-01-27
Resolve build warnings

### LINKED ISSUE #1461: [DevX] Noisy logs at startup due to unused components
Several errors of the form:

> Warning: /Users/gmiceli/Git/google/ground-platform/web/src/app/pages/main-page-container/main-page/title-dialog/title-dialog.module.ts is part of the TypeScript compilation but it's unused.
Add only entry points to the 'files' or 'include' properties in your tsconfig.

We should remove unused component which appear there.

rfontanarosa:
closes #1461

there are still two warning, one related to  "src/app/converters/geometry-converter.ts" which is not used.

the other one related to job-dialog-component.

should we have to keep boyh? @gino-m

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -221,8 +221,8 @@ catalogs:
       specifier: ^8.50.1
       version: 8.53.0
     diff:
-      specifier: ^4.0.4
-      version: 4.0.4
+      specifier: ^3.5.1
+      version: 3.5.1
     eslint:
       specifier: ^9.39.2
       version: 9.39.2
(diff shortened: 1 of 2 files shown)


web/src/app/components/main-page-container/main-page/secondary-side-panel/loi-panel/loi-panel.module.ts
```diff
@@ -24,10 +24,12 @@ import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { BrowserModule } from '@angular/platform-browser';
 
 import { LocationOfInterestPanelComponent } from './loi-panel.component';
+import { LoiPropertiesDialogModule } from 'app/components/shared/loi-properties-dialog/loi-properties-dialog.module';
```
gino-m, line 27: These can't be relative paths?

PR #2451, opened 2026-03-09
 Emit OK HTTP status in import callback instead of handler

### LINKED ISSUE #1920: [Code health] Emit OK HTTP status in handlers/callbacks instead of wrappers
Since other status codes are also set by the handlers, it makes more sense to set OK (200) there as well

rfontanarosa:
closes #1920

All status and error logic (except uncaught exceptions) is handled by the handlers. The wrapper does:                                                                                                - CORS
- Cookie parsing                                                                                                                                                                                    
- Auth                                                  
- Uncaught exceptions

This simplifies testing and avoids asymmetry between the wrapper and handlers.


functions/src/handlers.ts
```diff
@@ -123,7 +123,6 @@ export async function invokeCallbackAsync(
       res,
       user,
       () => {
-        res.status(StatusCodes.OK).end();
```
gino-m, line 129: This introduces an asymmetry which I think it would be better to avoid: `OK` is sent by handlers, but error status is set by this wrapper.

Why do we need to set OK in the handlers anyway?

functions/src/import-geojson.ts
```diff
@@ -37,18 +36,16 @@ class BadRequestError extends Error {
...
-  done: () => void,
-  error: ErrorHandler
+  user: DecodedIdToken
 ) {
   if (req.method !== 'POST') {
-    return error(
-      StatusCodes.METHOD_NOT_ALLOWED,
-      `Expected method POST, got ${req.method}`
-    );
+    res
```
gino-m, line 45: This is the opposite of what we discussed. Ideally the handlers wouldn't need to know anything about HTTP, they just define the algorithm independent of the transport layer.

gino-m (changes requested):

Issue #1537, opened 2024-02-29
[WHISP integration] Allow users to enable/disable integration per job


gino-m:
- [ ] Finish refactor https://github.com/google/ground-platform/issues/1536
- [ ] Add a UI somewhere (@vittorino Where?) to enable/disable one or more preconfigured "property generator" integrations, each with its own custom label (currently only "WHISP + AgStack Registry"
- [ ] Introduce `integration` collection under each job document

PR #2453, opened 2026-03-13
Add integrations map to Job proto and model

rfontanarosa:
towards #1537


proto/src/ground/v1beta1/job.proto
```diff
@@ -40,6 +40,15 @@ message Job {
   // One or more data collection tasks to be carried out by the user in order
   // to collect and submit data.
   repeated Task tasks = 5;
+
+  // Zero or more integrations enabled for this job.
+  repeated Integration integrations = 6;
```
gino-m: This should be called `enabled_integrations`

PR #2465, opened 2026-03-23
Fix crash when Firestore document contains null fields

### LINKED ISSUE #2048: [Feature] Handle null values in proto

rfontanarosa:
closes #2048

### DIFF
--- a/lib/src/firestore-to-proto.spec.ts
+++ b/lib/src/firestore-to-proto.spec.ts
@@ -125,6 +125,20 @@ describe('toMessage()', () => {
       expected: new Survey(),
     },
     {
+      desc: 'skips non-object map field',
+      input: {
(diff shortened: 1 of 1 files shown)


lib/src/firestore-to-proto.ts
```diff
@@ -77,7 +77,7 @@ function toMessageValue(
   if (!descriptor.fields) return null;
   const field = descriptor.fields[fieldName];
   const fieldType = field?.type;
-  if (field.keyType) {
+  if (field?.keyType) {
     if (field.keyType !== 'string')
       return Error(`${field.keyType} map keys not supported`);
     // TODO: Check that firestoreValue is an object.
```
gino-m: Can we also resolve this TODO while we're here?

Issue #2438, opened 2026-02-20
Possibility to enable or disable integrated services


kenstershiro:
First use case is WISP integration. Survey admins should be able to enable or disable the WHISP integration for a given Job
There will be other integrations to add in future, such as GeoID.

<img width="1210" height="832" alt="Image" src="https://github.com/user-attachments/assets/799c7a2c-a68d-4dcd-95be-bc3ff7737f18" />

kenstershiro:
@vittorino would you be able to look at this one next time you get a chance please?

rfontanarosa:
related to #1537

gino-m:
Done in #2476

PR #2468, opened 2026-03-24
Add the JobIntegrationControlComponent

rfontanarosa:
towards #2438

### DIFF
--- a/lib/src/firestore-to-proto.spec.ts
+++ b/lib/src/firestore-to-proto.spec.ts
@@ -111,6 +111,34 @@ describe('toMessage()', () => {
       }),
     },
     {
+      desc: 'skips null nested message',
+      input: {
+        '4': null,
+      },
+      expected: new Job(),
+    },
+    {
+      desc: 'skips null map field',
+      input: {
+        '4': null,
+      },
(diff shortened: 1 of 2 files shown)


web/src/app/components/shared/data-visibility-control/data-visibility-control.component.spec.ts
```diff
@@ -0,0 +1,129 @@
+/**
+ * Copyright 2025 The Ground Authors.
```
gino-m: 2026!

web/src/app/models/job.model.ts
```diff
@@ -25,7 +25,7 @@ export enum DataCollectionStrategy {
 }
 
 // Configuration for a single integration associated with a job.
-export interface Integration {
+export interface EnabledIntegration {
```
gino-m: I don't this needs to be `Enabled` - it's technically only enabled when it's in the `enabledIntegrations` list.

rfontanarosa: i thought that Integration was too generic. I wanted to call it JobIntegration but probably Integration is fine.

PR #2471, opened 2026-03-26
Fix session cookie being recreated on every export

### LINKED ISSUE #1160: Only generate new session cookie when needed
Export currently generates a new cookie on each export.

rfontanarosa:
closes #1160


web/src/app/services/auth/auth.service.ts
```diff
@@ -81,7 +88,20 @@ export class AuthService {
     private injector: Injector,
     private ngZone: NgZone
   ) {
-    this.auth.onIdTokenChanged(user => this.tokenChanged$.next(user));
+    // onIdTokenChanged via RxJS 'idToken' or 'user' specific helper?
```
gino-m: Please remove/cleanup agenti notes

web/src/app/services/auth/auth.service.ts
```diff
@@ -98,14 +118,22 @@ export class AuthService {
...
    * download arbitrarily large files (namely, "/exportCsv"). These functions can only be invoked via
    * HTTP GET, since browser do not allow streaming directly to disk via POST or other methods.
+   *
+   * Skips the server call if a valid session cookie was already created in this browser. The expiry
+   * timestamp is persisted in localStorage (sourced from the server response) so the check survives
+   * page refreshes. The entry is cleared when the user signs out (tracked via onIdTokenChanged).
    */
   async createSessionCookie() {
+    const stored = localStorage.getItem(SESSION_COOKIE_EXPIRES_AT_KEY);
+    if (stored !== null && Date.now() < parseInt(stored, 10)) {
```
gino-m: What if the cookie expires soon (e.g., 10ms) from now? In that case we should still refresh, since by the time it's needed it will have expired.

gino-m (changes requested):

web/src/app/services/auth/auth.service.ts
gino-m: We spoke, this was done.

functions/src/common/auth.ts
```diff
@@ -65,19 +65,22 @@ export async function getDecodedIdToken(
...
  * Generates and sets a session cookie for the current user into the provided response.
+ * Returns the absolute expiry timestamp (ms since epoch) of the created cookie.
  */
 export async function setSessionCookie(
   req: Request,
   res: Response
-): Promise<void> {
+): Promise<number> {
   const token = getAuthBearer(req);
   const expiresIn = 60 * 60 * 24 * 5 * 1000; // 5 days
```
gino-m: Please move into a constant in common lib.

functions/src/session-login.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+import * as adminAuth from 'firebase-admin/auth';
+import type { Auth } from 'firebase-admin/auth';
+import { Request } from 'firebase-functions/v2/https';
+import type { Response } from 'express';
+import { StatusCodes } from 'http-status-codes';
+import { sessionLoginHandler } from './session-login';
+
+const FIVE_DAYS_MS = 60 * 60 * 24 * 5 * 1000;
```
gino-m: Please refactor into lib.

web/src/app/services/auth/auth.service.spec.ts
```diff
@@ -16,17 +16,60 @@
...
 
 import { AuthService } from 'app/services/auth/auth.service';
 import { DataStoreService } from 'app/services/data-store/data-store.service';
+import { User } from 'app/models/user.model';
+import { environment } from 'environments/environment';
 
 import { HttpClientService } from '../http-client/http-client.service';
 
+const SESSION_COOKIE_EXPIRES_AT_KEY = 'sessionCookieExpiresAt';
+const FIVE_DAYS_MS = 60 * 60 * 24 * 5 * 1000;
```
gino-m: Please use lib.

gino-m (changes requested):

gino-m (approved):

PR #2484, opened 2026-04-03
Implement LOI property generator for GeoIDs

### LINKED ISSUE #1537: [WHISP integration] Allow users to enable/disable integration per job

### LINKED ISSUE #2113: [Integration] Send site+submission data to external API on upload
To enhance Ground's interoperability and enable seamless data flow with other applications, we propose implementing a mechanism to send collected data to external systems via an API at collection time.

This integration should be flexible and configurable, similar to the existing integration with WHISP for propertyGenerators. A generic configuration for each integration should be stored in Firestore.

Image

This probably should be linked to this issue https://github.com/google/ground-platform/issues/1537.

A detailed proposal from @vittorino is required to further define the technical implementation and design of this feature.

Image

**The main problem is the visibility of these integrations. maybe we list the integration using a sort of integration ACL?**

### LINKED ISSUE #2438: Possibility to enable or disable integrated services

rfontanarosa:
closes #2438
closes #1537
closes #2113

Screenshot 2026-09-03 alle 15 57 13


functions/src/property-generators/geoid.ts
```diff
@@ -14,8 +14,55 @@
...
-import type { Properties, PropertyGeneratorHandler } from './types';
+import type { Geometry } from 'geojson';
+import type { Headers, Properties, PropertyGeneratorConfig } from './types';
 
-// TODO: Implement geoid property generation.
-export const geoidHandler: PropertyGeneratorHandler =
-  async (): Promise<Properties> => ({});
+const defaultHeaders = { 'Content-Type': 'application/json' };
+
+export async function geoidHandler(
```
gino-m: Can we calle this geoId / GeoId throughout, to avoid confusion with the geometric shape "geoid"?

functions/src/property-generators/geoid.ts
```diff
@@ -14,8 +14,55 @@
...
+
+  const responseJson = await response.json();
+  const id = responseJson?.id;
+  if (!id) {
+    console.error(
+      `geoid: response missing id field, body=${JSON.stringify(responseJson)}`
+    );
+    return {};
+  }
+  console.log(`geoid: received id=${id}`);
```
gino-m: Since this will be emitted for every run, this looks more like a debug log than a regular info log message

functions/src/on-create-loi.ts
```diff
@@ -64,15 +64,27 @@ export async function onCreateLoiHandler(
...
+    if (!enabledIntegrationIds.has(generatorId)) {
+      continue;
+    }
 
-    properties = updateProperties(properties, newProperties, config.prefix);
+    try {
+      const newProperties = await handler(config, geometry);
+      properties = updateProperties(properties, newProperties, config.prefix);
+    } catch (e) {
+      console.error(
```
gino-m: Why would we continue to update the LOI properties if this fails?

gino-m (changes requested):
Please also update the PR description to something a bit more self-descriptive, e.g., "Implement property generator for GeoIDs"

functions/src/on-create-loi.ts
rfontanarosa: it is intentional. if an integration raises an error, the process continue with the others.

gino-m:
Doesn't WHISP already issue geoIDs? Why do we need a separate generator for GeoIDs?

jo-spek:
> Doesn't WHISP already issue geoIDs? Why do we need a separate generator for GeoIDs?

GeoIDs are supposed to be the ultimate pilar of interoperability. So that a geometry created in Ground can easily be loaded elsewhere without having to export import. If you would generate GeoIDs in Whisp, you would have to manually export-import from Ground to Whisp. We want any newly created geometry to have a GeoID immediately.

gino-m (dismissed):

gino-m (approved):

PR #2496, opened 2026-04-17
 Add test coverage for protobuf conversion edge cases

### LINKED ISSUE #1808: [Migration: protobuf for Firestore] Handle protobuf conversion edge cases
We currently don't have tests for when Errors are returned or for most reasons values might be skipped. We should add these.

@rfontanarosa FYI

rfontanarosa:
closes #1808

  - Add tests for error returns and value-skip branches in `firestore-to-proto.ts` and `proto-to-firestore.ts`                                                                                            
  - Covers: non-numeric keys, unrecognized field numbers, non-numeric enum values, `google.protobuf.Timestamp` conversion, repeated-with-non-array, Error-returning constructors, unknown message types,
  empty-string drop, unsupported value types, array input                                                                                                                                                 
  - No production code changes — test-only

### DIFF
--- a/lib/src/firestore-to-proto.spec.ts
+++ b/lib/src/firestore-to-proto.spec.ts
@@ -1,14 +1,14 @@
 /**
  * Copyright 2024 The Ground Authors.
  *
- * Licensed under the Apache License, Version 2.0 (the 'License');
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
  *     https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an 'AS IS' BASIS,
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


lib/src/firestore-to-proto.spec.ts
```diff
@@ -1,14 +1,14 @@
 /**
  * Copyright 2024 The Ground Authors.
  *
- * Licensed under the Apache License, Version 2.0 (the 'License');
```
gino-m, line 4: Please leave the header text alone.

lib/src/firestore-to-proto.spec.ts
```diff
@@ -1,14 +1,14 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
  *     https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an 'AS IS' BASIS,
```
gino-m, line 11: Same here.

gino-m (dismissed):
Nice work!

gino-m (approved):

PR #2497, opened 2026-04-23
Generate and show build version on About screen

### LINKED ISSUE #2015: Introduce build versions to web app
We currently don't have build numbers / versions in the web UI, only in Android. We should show a version in the about screen, which can help with formalizing release cuts as well as debugging issues

rfontanarosa:
closes #2015

Screenshot 2026-04-23 alle 14 58 19


gino-m (dismissed):

web/package.json
```diff
@@ -42,13 +42,13 @@
         },
         "configurations": {
           "development": {
-            "command": "ng build -c development"
+            "command": "node ../scripts/generate-build-info.mjs && ng build -c development"
```
gino-m: Instead of repeating this, can you create a separate nx script which this one depends on? e.g. `generateBuildInfo`

web/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@ground/web",
   "version": "0.0.1",
   "scripts": {
-    "postinstall": "node scripts/patch-karma.cjs"
+    "postinstall": "node scripts/patch-karma.cjs && node ../scripts/generate-build-info.mjs"
```
gino-m: Why do we do this both after install and on build? Can we just do it at one point in the workflow?

gino-m (changes requested):

gino-m (approved):

PR #2503, opened 2026-04-29
Sweep user-media files whose survey no longer exists

### LINKED ISSUE #2502: [Feature] Clean up orphan media files for deleted surveys
user-media files persist still even if the survey is deleted

rfontanarosa:
closes #2502


gino-m (approved):
Will you also remove media whose submission or loi were removed? This can be done in a follow-up PR.

PR #2512, opened 2026-05-12
Fix duplicateJob to regenerate task ids and remap conditions

### LINKED ISSUE #2517: [Feature] Duplicating a job reuses task IDs, breaking conditionals on Android
When a job or survey is duplicated, the cloned tasks keep their original IDs and conditional references still point to those old IDs. Android can't handle the resulting ID collisions, so duplicate-on-the-web produces invalid surveys. The fix needs to regenerate task IDs and remap any conditions to the new IDs.

rfontanarosa:
closes #2517

It seems that duplicating jobs and surveys while keeping the same ID for tasks causes issues on Android. Andreia mentioned this would require a major change on the Android side, so I handled the task IDs differently to ensure they are regenerated. The logic for the conditionals was the most complicated part, but it’s now working.

### DIFF
--- a/web/src/app/services/job/job.service.spec.ts
+++ b/web/src/app/services/job/job.service.spec.ts
@@ -16,10 +16,18 @@
 
 import { TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
-import { Map } from 'immutable';
+import { List, Map } from 'immutable';
 
 import { DataCollectionStrategy, Job } from 'app/models/job.model';
 import { DataSharingType, Survey } from 'app/models/survey.model';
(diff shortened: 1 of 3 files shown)


gino-m:
> It seems that duplicating jobs and surveys while keeping the same ID for tasks causes issues on Android. Andreia mentioned this would require a major change on the Android side, so I handled the task IDs differently to ensure they are regenerated. The logic for the conditionals was the most complicated part, but it’s now working.

This is also consistent with those IDs being unique (UUIDs). Is the fact that those IDs need to be globally unique documented in the protos and TS code?

gino-m (approved):

PR #2559, opened 2026-07-09
Rename "Unlisted" survey access to "Anyone with the link"

### LINKED ISSUE #2507: Rename "Unlisted" surveys to "Anyone with the link"
Calling survey access "Restricted" and "Unlisted" has just been creating confusion during user trainings. The term "Unlisted" makes sense from the developer's POV, but for users it is just weird and I have found it impractical and sometimes near to impossible to explain, resorting to "just use Unlisted if you want to share it by QR code".

Let us please change to something more user-friendly. I would simply suggest "Restricted" and "Unrestricted".

Please also let us move up the "Copy link" and "Copy QR code" buttons into the same pane and let them only show up, when "Unrestricted" is selected.

Image

@kenstershiro @rfontanarosa

rfontanarosa:
closes #2507

Screenshot 2026-07-09 alle 09 36 55

Screenshot 2026-07-09 alle 09 41 12

Are the three access levels really peers? "Restricted" and "Anyone with the link" both describe unlisted surveys — they differ only in who can open the link/QR (named people vs. anyone).

"Public" is a different dimension entirely: discoverability (the survey can be found without a link at all).

Given that, should the filter keep treating them as three flat siblings, or model it as listed vs. unlisted with link/QR access as a sub-option of the unlisted case?

This mainly affects how we label and group the survey-list filter chips.

wdyt? @gino-m @jo-spek

### DIFF
--- a/web/src/app/components/create-survey/create-survey.component.html
+++ b/web/src/app/components/create-survey/create-survey.component.html
@@ -114,11 +114,6 @@
       </div>
 
       <div class="right-button-section">
-        <ground-copy-survey-controls
-          *ngIf="createSurveyPhase === CreateSurveyPhase.SHARE_SURVEY"
-          [surveyId]="surveyId"
-        ></ground-copy-survey-controls>
-
         <button
           id="continue-button"
           class="continue-button"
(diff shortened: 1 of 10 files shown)


gino-m:
What do users use these filters for? I think organizers would normally be interested in surveys they own vs surveys shared with them vs public survey, so I don't think they're really fit for purpose and I would just remove them for now.

@jo-spek Wdyt?

gino-m (approved):

PR #2569, opened 2026-07-22
 Minor fixes: survey media access, Android intent fallback, LOI property refactor

rfontanarosa:
- Drop per-user scoping of survey media in Storage rules — any user with data-collection permission on a survey can read/write its media, not only the uploading user.
- Add an explicit Play Store fallback URL (S.browser_fallback_url) to the Android intent so the store opens when the app isn't installed (+ tests).
- Extract regenerateLoiProperties (and export propertiesPbToObject) from onCreateLoiHandler for reuse/testability.
- Add a fetchSubmission helper to Datastore.

### DIFF
--- a/functions/src/common/datastore.ts
+++ b/functions/src/common/datastore.ts
@@ -163,6 +163,10 @@ export class Datastore {
     return this.fetchDoc_(loi(surveyId, loiId));
   }
 
+  fetchSubmission(surveyId: string, submissionId: string) {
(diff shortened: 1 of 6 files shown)


gino-m:
There are a few related things in this one PR. Can you please have an agent split them into multiple PRS?

rfontanarosa:
closed because it has been split there:
https://github.com/google/ground-platform/pull/2575
https://github.com/google/ground-platform/pull/2576
https://github.com/google/ground-platform/pull/2577

PR #2576, opened 2026-07-27
Authorize user-media writes at the survey level

### LINKED ISSUE #2573: Survey media uploads rejected by Storage rules
Since #2501, the Storage rule only matches /user-media/surveys/{surveyId}/users/{userId}/**, but clients write to user-media/surveys/{surveyId}/submissions/... with no users/{userId} segment. The rule matches nothing on write, so photo upload fails for every data collector. Rules should authorize survey media by canCollectData(surveyId) / canViewSurvey(surveyId) at the survey level.

rfontanarosa:
closes #2573

### DIFF
--- a/storage/storage.rules
+++ b/storage/storage.rules
@@ -84,19 +84,12 @@ service firebase.storage {
       allow read: if isSignedIn();
     }
     
-    /**
-     * Returns true iff the requesting user owns the media at the given path.
-     */
-    function isResourceOwner(userId) {
-      return request.auth.uid == userId;
-    }
-
-    match /user-media/surveys/{surveyId}/users/{userId}/{allPaths=**} {
+    match /user-media/surveys/{surveyId}/{allPaths=**} {
       // Only users with permission to access the survey can read media.
       allow read: if canViewSurvey(surveyId);
 
-      // Only the owning user with permission to contribute data to the survey can create/update media.
-      allow create, write: if isResourceOwner(userId) && canCollectData(surveyId);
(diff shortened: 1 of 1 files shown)


storage/storage.rules
```diff
@@ -84,19 +84,12 @@ service firebase.storage {
...
-
-    match /user-media/surveys/{surveyId}/users/{userId}/{allPaths=**} {
+    match /user-media/surveys/{surveyId}/{allPaths=**} {
       // Only users with permission to access the survey can read media.
       allow read: if canViewSurvey(surveyId);
 
-      // Only the owning user with permission to contribute data to the survey can create/update media.
-      allow create, write: if isResourceOwner(userId) && canCollectData(surveyId);
+      // Only users with permission to contribute data to the survey can create/update media.
+      allow create, write: if canCollectData(surveyId);
```
gino-m, line 92: Shouldn't "write" only be for resources owners and survey organizers? Otherwise anyone can overwrite anyone elses' images?

gino-m (changes requested):

storage/storage.rules
rfontanarosa: we have a problem.

the resource owner is not saved anywhere. the path is /user-media/surveys/XXX/submissions/{taskId}-{uuid} and there is no metadata on it.

this change just reverts to the previous state, i've open a ticket related to this: https://github.com/google/ground-platform/issues/2578

gino-m (approved):

PR #2577, opened 2026-07-27
Preserve the survey id in the Play Store install fallback

### LINKED ISSUE #2572: Android app link does not open the Play Store when the app is not installed
The intent URL built by AndroidIntentLandingPageComponent has no S.browser_fallback_url, so Chrome has no fallback target and the user lands on a blank page instead of the Play Store. The existing JS timeout fallback races with the intent handoff and doesn't fire reliably.

rfontanarosa:
closes #2572

### DIFF
--- a/web/src/app/components/android-intent-landing-page/android-intent-landing-page.component.spec.ts
+++ b/web/src/app/components/android-intent-landing-page/android-intent-landing-page.component.spec.ts
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2025 The Ground Authors.
+ *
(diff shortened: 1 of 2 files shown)


web/src/app/components/android-intent-landing-page/android-intent-landing-page.component.spec.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * Copyright 2025 The Ground Authors.
```
gino-m: 2026

gino-m (dismissed):

gino-m (approved):

PR #2579, opened 2026-07-30
Add "delete site" option 

### LINKED ISSUE #2565: [Edit] Delete sites
Like https://github.com/google/ground-platform/issues/2560, but allowing deletion of the whole site.

rfontanarosa:
closes #2565

https://github.com/user-attachments/assets/763cab01-641f-4f9d-aa3d-86231f4dcc2b


gino-m (changes requested):
We should also call out that the data will be removed in the menu option, eg:

"Delete site and associated data"

rfontanarosa:
> We should also call out that the data will be removed in the menu option, eg:
> 
> "Delete site and associated data"

isn't the menu label too long? everything is described in the confirmation dialog and an additional option "Delete submissions" will be added in the near future (see image).

another possibility is to have "Delete site and data" or "Delete site and submissions", but they are super long as well.

wdyt? @gino-m @jo-spek @vittorino 

Screenshot 2026-08-01 alle 11 01 35

gino-m:
Then I would go with "Delete site"

gino-m (approved):

PR #2588, opened 2026-09-02
Send typed broadcast updates for survey, job, and LOI writes

### LINKED ISSUE #2580: Reduce unnecessary Firestore reads

rfontanarosa:
closes #2580

## Summary
- Replaces the generic, empty-payload `broadcastSurveyUpdate()` with a typed `broadcastUpdate()` sent from the survey, job, and LOI write triggers.
- Each FCM message now carries `type` (`survey` | `job` | `loi`), the affected entity's id, and the triggering event's commit time (`eventTime`); LOI updates also carry `deleted`.
- Messages keep sharing one collapse key per survey, so a burst of writes (e.g. importing LOIs) still collapses into a single wake-up.
- Newly created LOIs now have their created/lastModified serverTimestamp corrected to the actual Firestore trigger event time, replacing the client-guessed value written at import time.
- Splits the `lois/{loiId}` trigger: `onWriteLoi` is replaced by `onUpdateLoi` and `onDeleteLoi`, and `onCreateLoi` now announces too. Both were previously registered on the same path, so every created LOI was broadcast twice, the first time before its generated properties had been written.
- A created LOI is now announced once, after it settles: `onCreateLoi` stays silent when it writes the fix-up and leaves the announcement to `onUpdateLoi`, and announces the LOI itself when there is nothing to fix up.
- Property enrichment moves to `common/loi-properties.ts`, so `export-geojson` no longer imports it from a trigger module.

## Compatibility
Verified against `ground-android`: `FirebaseMessagingService.onMessageReceived()` only reads `remoteMessage.from` (the topic) to decide which survey to resync, and never inspects `remoteMessage.data`. The added fields are additive and ignored by the current app, so existing clients keep doing a full resync exactly as before. Newer clients can use the type/id hints to sync just what changed.


gino-m:
I see we opted to stick with hard-delete via FCM (push) rather than soft delete + pull? What happens if many entites are deleted (e.g. an entire survey)? Wouldn't clients get flooded with messages? Same question on batch create and write - is there a way to debounce the create and write update messages?

